Issue #13, opened 2026-03-08
Fix: Cooldown Manager "Cropped" icon shape shows grey lines on left/right of icons


nulltyto:
This issue was made on behalf of discord user Eko.

Description

When the "Cropped" icon shape is selected in Cooldown Manager, grey pixel lines appear on the left and right sides of each icon. This does not occur with Action Bars using the same shape.

Root Cause

In CDM, the icon texture was inset from the frame edges by `borderSize`, leaving WoW's built-in grey icon edge pixels visible - the default zoom of 0.02 only cropped 2% per side. In Action Bars, the icon fills the entire button (`SetAllPoints`) with borders rendered on top in the OVERLAY layer, hiding those edges.

Additionally, `RefreshCDMIconAppearance` did not account for the cropped shape, overwriting the asymmetric tex coords with uniform ones whenever settings changed.

Fix

- Fix implemented on fork at https://github.com/derekmikesell/EllesmereUI/tree/fix/cropped-cooldown-icons, can create pull-request if aligns w/ vision.
- CDM cropped icons now use `SetAllPoints` (matching Action Bars behavior)
- `RefreshCDMIconAppearance` now respects the `iconShape` setting
- Extracted shared helpers (`PP.CroppedHeight`, `PP.SetIconTexCoords`, `PP.AnchorIconTexture`) to eliminate duplicated magic numbers across 6 files
- Increased default cropped zoom from 2.0 to 4.0 (this eliminates default grey bars)
- Fixed several related preview rendering bugs (double-cropping on tab switch, border thin/normal collapsing at 1440p, CDM zoom slider inconsistency)

PR #62, opened 2026-03-12
feat: spell ID aura filtering (fixed oUF API)

danvernon:
## Summary
- Reapplies per-spell-ID whitelist/blacklist filtering for auras on nameplates and unit frames
- Fixes the oUF integration: uses `FilterAura(element, unit, data)` instead of the deprecated `CustomFilter` with positional args
- Shared filter core lives in `EllesmereUI.lua` (main addon) — no oUF dependency for nameplates

## Changes from original PR #55
- `CustomFilter` → `FilterAura` (current oUF API)
- Function signature: `(element, unit, data)` with `data.spellId` instead of positional arg #13

## Test plan
- [ ] Enable filter in Global Settings, set to Blacklist with a known spell ID
- [ ] Verify aura disappears from **nameplates**
- [ ] Verify aura disappears from **unit frame** target debuffs/buffs
- [ ] Switch to Whitelist → only listed auras show
- [ ] Disable per-system toggles → auras return to normal
- [ ] Test with unit frames addon disabled → nameplate filtering still works
- [ ] `/reload` → settings persist

### DIFF
--- a/EUI__General_Options.lua
+++ b/EUI__General_Options.lua
@@ -1698,6 +1698,68 @@ initFrame:SetScript("OnEvent", function(self)
 
         _, h = W:Spacer(parent, y, 20);  y = y - h
 
+        -----------------------------------------------------------------------
+        --  AURA SPELL ID FILTER
+        -----------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


EllesmereGaming:

PR #98, opened 2026-03-15
fix: taint fixes for action bar secret value errors   

natty:
## Bug                                                                                            
  `ActionButton.lua:609: attempt to compare a secret number value (tainted by 'EllesmereUIActionBars')` during combat. Lower frequency than the other one                                          
                                                                                                    
  ## Root Cause
  `GetButtonActionSlot()` reads `btn.action` directly — a protected attribute that became a secret value in Midnight. The `hooksecurefunc` on `UpdateUsable` calls this during combat, tainting the value before Blizzard's own `assertsafe()` comparison.

  ## Fix
  - Replace `btn.action` reads with a `buttonToBar` lookup table populated during setup
  - `GetButtonActionSlot()` resolves slots from `BAR_SLOT_OFFSETS` + cached `_mainBarOffset` instead
   of the protected attribute
  - `CaptureBlizzardDefaults()` uses `BAR_SLOT_OFFSETS` instead of
  `btn.action`/`HasAction(btn.action)`

  ## Test
  - Enter combat with range coloring enabled — no taint errors on action buttons
  - Verify out-of-range tinting still works
  - Stance bar and pet bar should remain functional (no ADDON_ACTION_BLOCKED)
  - Override action bar buttons (vehicles, etc.) should not throw errors

### DIFF
--- a/EllesmereUIActionBars/EllesmereUIActionBars.lua
+++ b/EllesmereUIActionBars/EllesmereUIActionBars.lua
@@ -750,13 +750,8 @@ local function HideBlizzardBars()
         local bar = _G[name]
         if bar then
             bar:UnregisterAllEvents()
-            if name ~= "StanceBar" and name ~= "PetActionBar" then
-                if bar.numButtonsShowable ~= nil then
-                    bar.numButtonsShowable = 12
-                end
-                bar:SetParent(hiddenParent)
-                bar:Hide()
-            end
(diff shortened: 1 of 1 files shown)


natty:
This is not intended to be merged, but can be I guess. It's more "i'm going to sleep, if you still have problems you can try this one" which was one of the taint sites.

PR #158, opened 2026-03-18
Add 4-way grow direction for CDM bars

danvernon:
## Summary
- Expands the **Grow Direction** dropdown from 2 options (Horizontal/Vertical) to all 4 directions: **Right, Left, Down, Up**
- Users can now control exactly which way CDM bar icons grow
- Backwards compatible — existing profiles with `verticalOrientation` or `growDirection` saved will read correctly

## Test plan
- [ ] Open CDM options for any bar (cooldowns, utility, buffs)
- [ ] Verify "Grow Direction" dropdown shows Right, Left, Down, Up
- [ ] Select each direction — icons should grow in the chosen direction
- [ ] Load an existing profile that had Vertical Orientation enabled — should show "Down" selected


EllesmereGaming:
Round 2 of cleanup ty king

PR #163, opened 2026-03-18
Add duration text to the cooldown manager bar glows

hjaltethor:
Add optional countdown duration text for bar glows in both the preview UI and active overlays with sensible default values. Enable users to toggle it on and off as well as change the looks in the options.
image
image


EllesmereGaming:

PR #191, opened 2026-03-21
Add Lag Tolerance Helper and recommendations

Kneeull:
Fixed Lag Tolerance resetting on every reload and added in a recommended Tolerance setting based on the users ping (scraped at game run time (on reload and game launch), however sometimes fails on first game launch, but works on a reload. This is due to how blizzard loads it after the user has entered the world)

### DIFF
--- a/EUI__General_Options.lua
+++ b/EUI__General_Options.lua
@@ -72,6 +72,45 @@ initFrame:SetScript("OnEvent", function(self)
         return cur == def
     end
 
+---------------------------------------------------------------------------
+    --  Lag Tolerance Helper/Adviser
+---------------------------------------------------------------------------
+    
+    local function Clamp(val, lo, hi)
+        if val < lo then return lo end
+        if val > hi then return hi end
+        return val
+    end
+
+    local function EnsureEUIDB()
+        if not EllesmereUIDB then EllesmereUIDB = {} end
+        return EllesmereUIDB
+    end
+
+    local function GetRecommendedSpellQueueWindow()
+        local _, _, homeMS, worldMS = GetNetStats()
+
+        homeMS = tonumber(homeMS) or 0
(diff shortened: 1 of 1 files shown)


Ignitetheskywow:
Just to comment on this that players should almost always play with 400 (Default), exposing this option to (uninformed) players in the base EUI options was a mistake already.

EllesmereGaming:
From my discussions and research on it, I haven't found that 400 being correct for most players is accurate at all, but I can agree this should be gated behind a toggle by default. Adding one today

EllesmereGaming:
> Just to comment on this that players should almost always play with 400 (Default), exposing this option to (uninformed) players in the base EUI options was a mistake already.

actually, i realized that this has been fixed for several months ever since my initial comment. I do not ever automatically set this setting for any users. the cvar is only ever updated if the user physically chooses to adjust the slider. and the sliders position is based on the users cvar when the addon is initially installed so I do not touch that setting unless the user chooses to change it. 

are you saying that I shouldn't even have the setting user facing at all? even though the user has to actively change it for it to actually change anything for them?

Ignitetheskywow:
https://maxroll.gg/wow/resources/spell-queue-window For a succinct explanation. You could argue that sometimes you dont want to cast a spell you queued 400ms before and instead want to do *nothing* in wich case the SQW can come to bite you, but that is such a niche situation... . (Also the Tooltip when you hover over the setting does not accurately resemble what the setting does)

EllesmereGaming:
I originally learned from jeathe many years ago about the SQW but perhaps information has changed. Xerwo knows his shit so i'll update the tooltip to 200+ping minimum, 400 if they are unsure of exactly what it is. personally i've tried many spellqueue window settings and have always preferred 150 as a melee but i'll give 200+ping another go. ty for the info!

PR #276, opened 2026-04-10
Feat/multiple visiblity options

Phylo:
This is allowing the choosing of multiple Visibility options for Action bars, Resource Bars, and Unit Frames.

it's an attempt at adding a common component the EllesmereUI_VisibilityModes  so that any of the modules can reference it to use as their visibility options and keep the logic the same between all the modules.


EllesmereGaming:

PR #277, opened 2026-04-10
Feature/ Mythic+Timer Module

Kneeull:
Creates and implements a Mythic+ Timer.

Apologies about the messy PR, my github decided it was time to be difficult.

Also - the shared media was removed in my branch when I pushed and not added yet, so added that back in - functionally no change to the OG file other than adding a hook for it.


Kneeull:
Currently in testing in Disc - Will update with any issues/pass or fail

Kneeull:
Closing - recreated in another branch.

PR #279, opened 2026-04-10
Add Mythic+ Timer module and UI hooks

Kneeull:
Introduce a new EllesmereUIMythicTimer module (standalone timer + options) and wire it into the main UI.

- Add EllesmereUIMythicTimer/EllesmereUIMythicTimer.lua: standalone Mythic+ timer overlay, run/objective tracking, HUD rendering, and Blizzard M+ suppression.
- Add EllesmereUIMythicTimer/EUI_MythicTimer_Options.lua: options page registration and controls for the new module.
- Update EllesmereUI.lua to register the new module so it appears in the sidebar.


Kneeull:
Moving back to draft - adding further customisation behind gated settings, adding presets after the most popular addons, and bringing feature parity up whilst doing performance optimisation passes to avoid high CPU usage.

PR #296, opened 2026-04-26
Introduce testing workflow and expand bug-finding coverage across core modules

Anysu94:
This PR focuses on establishing testing as a repeatable development workflow for the addon, not just adding a few isolated specs.

The main goal is to push the repository toward a behavior-first test workflow that exercises real module scenarios in the project’s existing containerized environment. The added tests are intentionally written close to user-facing behavior and module decision logic, so they can both improve coverage and expose real product defects early. Where bugs were confirmed, they were documented through intentionally red tests instead of being silently worked around or fixed in this PR.

Concretely, this PR adds and expands focused module tests for:

Quest Tracker
Action Bars
Friends
Resource Bars
It also consolidates currently confirmed failures in a central bug document:

Testing/BUGS.md
Why this PR matters

This change is primarily about workflow adoption.

Instead of treating testing as a one-off verification step, this PR moves toward a practical workflow where:

module logic is exercised through realistic scenarios
tests run in the repo’s containerized test environment
failing tests are allowed to document confirmed product bugs
coverage improvements come from meaningful behavioral checks, not synthetic assertions
bugs found during test authoring are captured centrally for follow-up by product developers
That makes testing useful not only for regression prevention, but also for active defect discovery.

What was added

New and expanded tests cover areas such as:

Quest Tracker loader, DB initialization, slash command wiring, suppression handling, QoL flows, and visibility behavior
Action Bars legacy visibility compatibility and migration-related behavior
Friends realm-group resolution and region fallback logic
Resource Bars exported helper behavior, class/spec resource routing, and resource color resolution
The tests follow the repo’s established container workflow via [run-tests.ps1](vscode-file://vscode-app/c:/Users/z002tymy/AppData/Local/Programs/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html), so they are aligned with the intended execution path for local and CI-style runs.

Confirmed bugs now documented by tests

This PR intentionally surfaces confirmed product issues instead of fixing them.

Examples now covered by red tests or documented failing specs include:

Quest Tracker mouseover visibility registration and stale auto-accept suppression state
Action Bars legacy mouseover alpha normalization/copy issues
Cooldown Manager default-shape and preset/variant handling issues
Resource Bars using FURY instead of PAIN for Vengeance Demon Hunters
Resource Bars generating inverted pip geometry on very narrow bars
These are documented in Testing/BUGS.md so the failing behavior is explicit and reviewable.

Testing notes

This PR validates the workflow through focused containerized runs rather than relying on ad hoc local execution.

Current characteristics of the test approach:

behavior-first specs rather than purely implementation-driven assertions
minimal environment stubs to execute real addon logic outside WoW
intentionally red tests where a bug is confirmed
LuaCov coverage improvements driven by meaningful module paths
As one example, EllesmereUIResourceBars/EllesmereUIResourceBars.lua now has focused behavioral coverage and confirmed bug coverage in the same spec, with local coverage reaching 16.34% in the latest targeted run.

Reviewer guidance

Please review this PR primarily as a testing-workflow PR:

it introduces and extends executable module-level test coverage
it documents confirmed product bugs through tests
it does not attempt to fix the underlying product issues
some red tests are intentional and expected


.vscode/settings.json
loicngr, line 1: I'd suggest removing this, it looks like a personal editor config. Maybe add .vscode/ to .gitignore instead ?

.github/workflows/ci.yml
loicngr, line 1: I notice there's no branch filter on triggers, so this would run CI on every push to any branch. Is that on intentional?

Testing/BUGS.md
loicngr, line 1: Nice research, the bug reports reference real internal code paths. Once failing tests become pending(), it might be worth opening GitHub issues for each bug so they're easier to track.

.vscode/settings.json
Anysu94: removed file and added to gitignore. was a local workspace config for my vs code that i added by accident.

.github/workflows/ci.yml
Anysu94: Yes, so far that is intentional. This way you can see if new changes would cause any tests to fail as soon as possible.

Other options would be to only execute tests on PR or to only run them manual, but I wouldn't recommend that. The tests only take a few seconds anyway.

Right now the Pipeline will fail if any of the tests will fail. This behavior could be changed to always succeed even if some of the tests fail. But I would also advice against that, because then you have to manual check the test run each time.

Testing/BUGS.md
Anysu94: Probably a good idea. i just wanted to provide some initial Data to make it more easy to fix the productive code.

Later on it would make sense to delete this file and keep track with issues.

PR #298, opened 2026-04-26
New tooltip options: mouse anchor and hide during combat

TobiWan54:
I have found that addons such as TipTac Reborn and Leatrix Plus conflict with the BlizzardSkin module and are not able to change the anchor or hide the skinned tooltips, so this PR adds a couple of options to control tooltips: anchoring them to the mouse and hiding them during combat.

Please let me know if it needs any improvements :)


TobiWan54:
I have also added an option to always hide tooltips, with a submenu to select which types, and therefore only show them while holding a modifier key.

EllesmereGaming:
going through old PRs, this was implemented sorry I missed this

Issue #302, opened 2026-05-05
[Unit Frames] Missing UpdateToolitp logic


Stein-N:
Every UnitFrame is missing a the UpdateTooltip logic.
This prevents any additional modification of the Tooltip by other AddOns.

This can be fixed with the Vanilla `UnitFrame_UpdateTooltip` function, this inherits the right click line:

<img width="271" height="205" alt="Image" src="https://github.com/user-attachments/assets/fa9320d0-453e-4f0b-8279-31cba1bf470f" />

Or with this method without the right click line:
```lua
local function Custom_UpdateTooltip(frame)
   GameTooltip_SetDefaultAnchor(GameTooltip, frame)
   if GameTooltip:SetUnit(frame.unit) then
      GameTooltip:Show()
      
      frame.UpdateTooltip = Custom_UpdateTooltip
   else
      frame.UpdateTooltip = nil
   end
end
```
<img width="229" height="164" alt="Image" src="https://github.com/user-attachments/assets/ff9aaa20-3d7c-4bdc-a989-df606d24f0d2" />

Stein-N:
This issue persists for Party and Raid frames, Unit frames updating fine

PR #304, opened 2026-05-08
Fix missing Tooltip Update Logic (#302)

Stein-N:
Simple fix for the missing Update Logic.

The Result:
image


Stein-N:
@EllesmereGaming would say this fix should be implemented, since there are enough addons that add additional data/lines to the tooltip.
Without this this is a compatibility issue

EllesmereGaming:
merged here but will be throttling it to .5 second update so it doesnt impact performance

EllesmereGaming:
thank you! it'll be live with next patch

PR #311, opened 2026-05-10
Minimap: grouped addon flyout expand left/right

liamcooper:
## Summary
- Adds `addonFlyoutExpand` (default `right`) so the grouped minimap addon flyout can anchor left or right of the stack toggle.
- Minimap options: **Expand direction** lives in **Icons and Buttons** on the same row as **Interactable Button Size**; following rows shift so the grid stays two-column.
- Refreshes options page layout cache (`pageLayoutVersion`, `PageCacheKey`) and fixes `DualRow` when the right column is omitted (full-width left).

## Testing
- [ ] /reload; open EllesmereUI → Minimap; set Expand direction Left/Right and open the grouped addon stack flyout.

image

### DIFF
--- a/EllesmereUI.lua
+++ b/EllesmereUI.lua
@@ -3818,7 +3818,7 @@ local function CreateMainFrame()
         -- Release cached pages for non-active tabs to free memory.
         -- Keep the active page so reopening shows the same content.
         if _pageCache then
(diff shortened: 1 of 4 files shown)


EllesmereGaming:
going through old PRs, this was implemented sorry I missed this

PR #317, opened 2026-05-13
EllesmereUIDamageMeters - Roster Panel feature

imsamdez:
Add an icon in the damage meter header.
When a player hover the icon, it shows a popup with all the member of the roster (group/raid) and their ilvl.

This is inspired from an existing feature of Details. In the Details one, the durability % is also shown. Unfortunately, this is done by the fact that Details broadcast to other player the durability since this data cannot be retrieve from the ingame API. That said, we can think of adding it later when EUI becomes a standard (hopefully and we're in a great pace).
image


EllesmereGaming:

PR #339, opened 2026-05-22
QoL: Current Expansion Only for AH and crafting orders

liamcooper:
## Summary
- Adds a single **Current Expansion Only** QoL toggle (`ahCurrentExpansion`) for both Auction House browse and Place Crafting Order.
- Moves logic into `EllesmereUIQoL_ExpansionFilters.lua` with secure `hooksecurefunc` hooks; setup events unregister after Blizzard addons are hooked.
- AH applies the filter after Blizzard resets search bars on show; crafting orders apply on open and respect manual uncheck until the window is reopened.

## Test plan
- [ ] Enable **Current Expansion Only** in QoL, `/reload`
- [ ] Open Auction House browse — filter is on; switching tabs still applies after Blizzard reset
- [ ] Open Place Crafting Order — filter is on at open (favorites list unchanged; expansion applies on Search)
- [ ] Uncheck expansion in Filter menu — stays off until window is closed and reopened
- [ ] Disable toggle — no filter applied on AH or crafting orders

### DIFF
--- a/EllesmereUIQoL/EUI_QoL_Options.lua
+++ b/EllesmereUIQoL/EUI_QoL_Options.lua
@@ -219,8 +219,8 @@ initFrame:SetScript("OnEvent", function(self)
         end
 
         _, h = W:DualRow(parent, y,
-            { type="toggle", text="AH Current Expansion Only",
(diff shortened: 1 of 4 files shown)


EllesmereGaming:
going through old PRs, sorry this got left behind without comments, it was implemented a while ago

PR #343, opened 2026-05-23
Improve chat session history restore with clickable player links

liamcooper:
## Summary

Single-file change: `EllesmereUIChat/EllesmereUIChat_SessionHistory.lua`.

### Clickable player names (realm chat)
- Unified player-link helpers (`FormatPlayerChatLine`, `RebuildPlayerMessageOnRestore`, `PlayerHyperlink`) for all captured player-linked types: say, yell, party, guild, officer, raid, channel, whisper (in/out).
- Capture stores `playerName`, `rawBody`, `lineID`, and `guid` so lines can be rebuilt with `|Hplayer:` links after `/reload`.
- Uses Blizzard `CHAT_*_GET` templates via `ApplyChatTemplate`, with correct handling for both single-`%s` (say/yell/whisper) and double-`%s` (party/guild/raid/channel) formats.
- Falls back to plain formatted lines when links cannot be built (e.g. secret values).

### BN whispers
- Persists `bnSenderID`, `bnPlayerToken`, `rawBody`, and `guid`; rebuilds `|HBNplayer:` links on restore.
- Refreshes BN display names after reload via `RefreshBNLinksInMessage`.
- Drops legacy BN rows on sanitize that lack `bnSenderID` and have no `|HBNplayer:` link in the stored message.

### Channel-aware restore
- Captures `channelName`, `channelBaseName`, `zoneChannelID`, and `channelIndex`.
- `FrameShowsChannelEntry` mirrors Blizzard `channelList` / `zoneChannelList` matching so Trade vs General (and similar) restore to the correct tab.

### Dedupe and direction
- Stored-entry dedupe keys: `p:CHATTYPE:NAME:payload` and `bn:CHATTYPE:ID:payload` (incoming/outgoing no longer collapse when body text matches).
- Frame-side dedupe (`BuildFrameMessageSet`, `PlayerLinkedDedupKeysFromStored`) uses the same key shape to avoid double-restore after `/reload`.
- Restore uses full chat types from events (`WHISPER_INFORM`, `BN_WHISPER_INFORM`, etc.) instead of stripping `_INFORM`.

### Capture quality
- `ChatMessageBody` and `ChatSenderName` resolve body/sender from `lineID` first, then event args, so stored `rawBody`/`playerName` match the displayed line.
- Legacy migration: parsers extract name/body from old plain-text saved lines; `whisperPlayerName` is still read for backward compatibility.

### Removed / unchanged behavior
- Emotes removed from capture (`CHAT_MSG_EMOTE`, `CHAT_MSG_TEXT_EMOTE`) and dropped on sanitize.
- Still open-world capture only; instance/restricted chat not persisted (existing limitation).

## Test plan
- [ ] `/reload` after `/say` and `/yell` — sender name is clickable
- [ ] Party/guild/raid messages restore with clickable names
- [ ] Trade channel message restores to Trade tab, not General
- [ ] Incoming and outgoing whispers with identical text both restore (no dedupe collapse)
- [ ] BN whisper in/out restores with clickable BN link and correct direction
- [ ] BN friend display name updates correctly after `/reload`
- [ ] Legacy BN rows without `bnSenderID`/links are dropped on sanitize (expected)
- [ ] Emote lines are not captured or restored

### DIFF
--- a/EllesmereUIChat/EllesmereUIChat_SessionHistory.lua
+++ b/EllesmereUIChat/EllesmereUIChat_SessionHistory.lua
@@ -14,6 +14,8 @@ if not ECHAT then return end
 
 local strsub = string.sub
 local gsub = string.gsub
+local format = string.format
+local strupper = string.upper
 local wipe = wipe
 local GetTime = GetTime
 local GetServerTime = GetServerTime
@@ -41,7 +43,6 @@ local UnarmDeferredRestore
 -- Capture in open world only (CaptureAllowed); instance chat still not storable.
 local CAPTURE_EVENTS = {
     "CHAT_MSG_SAY", "CHAT_MSG_YELL",
-    "CHAT_MSG_EMOTE", "CHAT_MSG_TEXT_EMOTE",
     "CHAT_MSG_PARTY", "CHAT_MSG_PARTY_LEADER",
     "CHAT_MSG_GUILD", "CHAT_MSG_OFFICER",
     "CHAT_MSG_CHANNEL",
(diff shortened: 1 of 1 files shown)


EllesmereGaming:

PR #351, opened 2026-05-23
Unit frames: independent left/right/center health and BTB text

liamcooper:
## Summary

Health bar and bottom text bar (BTB) text on **player, target, and focus** can use **left, right, and center** at the same time. Each slot has its own content, size, color, and X/Y offset.

**Mini frames** (pet, target-of-target, boss) use **center or left/right**: center can be used alone, or left and/or right together, but not all three at once.

Refactored duplicated positioning logic into a shared helper and aligned the options preview with live frame truncation.

### Full frames (Frame Display — player / target / focus)
- Left, right, and center health bar text can all be active together.
- Duplicate content across slots is prevented (the same value cannot occupy two slots).
- BTB supports the same combined layout; selecting center no longer clears left/right.

### Mini frames (Mini Frame Edit — pet / ToT / boss)
- Left, right, and center are all configurable in the UI.
- When center text is set, left and right are cleared, disabled in the UI, and hidden on the frame. When center is `none`, left and/or right work as usual.

### Code quality
- Added `ApplyBarOverlayTextPositions` for health bar and BTB layout.
- Fixed `ApplyClassColor` load order (nil on frame init).
- Preview applies width constraints to right text as well as left.

## Decisions

| Decision | Rationale |
|----------|-----------|
| All three slots on full frames | Normal-sized bars with per-slot X/Y offsets. |
| Mini frames: center or sides | Narrow bars; center and side text are mutually exclusive. |
| No center-aware width truncation | Left/right width still only accounts for each other; overlap is handled via offsets for now. |
| BTB only on full frames | Mini frames have no text bar settings or runtime bar. |
| Boss follows mini text rules | Same center-or-sides behavior as pet and ToT on Mini Frame Edit. |

## Test plan

- [ ] Player/target: left + right + center with X/Y offsets; `/reload`
- [ ] Focus: same with default right text `perhp`
- [ ] BTB: left, right, and center together
- [ ] Pet / ToT: center only vs left+right (not all three)
- [ ] Boss: same mini text rules as pet/ToT
- [ ] Options preview matches in-game layout

### DIFF
--- a/EllesmereUIUnitFrames/EUI_UnitFrames_Options.lua
+++ b/EllesmereUIUnitFrames/EUI_UnitFrames_Options.lua
@@ -284,6 +284,18 @@ initFrame:SetScript("OnEvent", function(self)
     }
     local healthTextOrder = { "none", "---", "name", "perhp", "perhpnosign", "curhpshort", "perhpnum", "both" }
 
+    local PREVIEW_TEXT_WIDTHS = {
+        both = 75, perhpnum = 75, curhpshort = 38, perhp = 38, perhpnosign = 30,
+        perpp = 38, curpp = 38, curhp_curpp = 75, perhp_perpp = 75,
+    }
+    local function EstimatePreviewTextWidth(content)
+        return (PREVIEW_TEXT_WIDTHS[content] or 0) + 10
+    end
+
+    local function IsCompactUFUnitKey(unitKey)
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
this was also added during 8.0 sorry about that LOL

PR #360, opened 2026-06-01
feat(unitframes): detached power bar border and layering/text-color fixes

Aderfi:
## Summary

A few improvements to the detached power bar on unit frames:

- Render the detached power bar above the unified border instead of underneath it.
- Give detached power bars their own border so they don't look unfinished next to the framed health bar.
- Fix the "Power Colored Text" option losing its color after a reload or loading screen.

## How it works

**Layering.** The unified border inherits the frame's (user-configurable) strata and never sets its own, so ordering within a unit frame should be decided by frame level, not strata. The detached power bar broke that in two places: `CreatePowerBar` hardcoded its strata to `MEDIUM` (so it fell under the border whenever `frameStrata` was MEDIUM or higher, and desynced from its own percent-text overlay), and `ReparentBarsToClip` re-asserted its frame level to `hpLevel+2` (= frame+4) on every reparent, below the border at frame+10. The fix lets the power bar inherit the frame strata like the border does and orders both by frame level: the detached bar now sits at frame+12 (above the border) in both code paths, while the attached bar keeps `hpLevel+2` (below it).

**Border.** The detached bar's border is a child frame anchored to the bar's edges, so it shares the bar's strata and moves with it. It's fixed at size 1 but pulls its texture, color and alpha from the frame's border settings, so it matches the rest of the frame.

**Power-colored text.** The percent text resolved its color once at build time via `PowerBarColor[UnitPowerType(unit)]`. After a reload the unit's power type isn't ready yet, so the lookup failed and the text stuck on white/mana. The color logic is now split out and re-resolved on power updates (the events oUF already uses for the percent value) via `GetPowerColor` — the same source the options preview uses, which also honors custom power colors — so it self-corrects once the real power type is known.

## Changes

- **EllesmereUIUnitFrames/EllesmereUIUnitFrames.lua**
  - `CreatePowerBar` — removed the hardcoded `MEDIUM` strata so the bar inherits the frame strata; added the detached-border call.
  - `ReparentBarsToClip` — detached power bar is re-leveled to frame+12 (above the border); attached path keeps `hpLevel+2`.
  - `ApplyDetachedPowerBorder` (new) — creates/refreshes the size-1 border for detached bars, inheriting the frame's border style; hidden when not detached.
  - `UpdateBordersForScale` — refreshes the detached power border on scale/texture changes.
  - `ApplyPowerPercentColor` (new) — token-cached color resolution shared by `ApplyPowerPercentText` and the power `PostUpdate`, so power-colored text re-resolves on power updates.

## Testing

Tested in-game on retail, both inside and outside instances:

- Detached power bar renders above the frame border, with its size-1 border matching the frame style.
- "Power Colored Text" keeps the correct color after `/reload`, relog and loading screens.
- No regressions on attached power bars (below/above) or on the other unit frames.

### DIFF
--- a/EllesmereUIUnitFrames/EllesmereUIUnitFrames.lua
+++ b/EllesmereUIUnitFrames/EllesmereUIUnitFrames.lua
@@ -1789,16 +1789,58 @@ local function ReparentBarsToClip(frame, powerPosition)
             if frame.Power:GetParent() == clip then
                 frame.Power:SetParent(frame)
             end
+            -- Detached power renders outside the frame, on top of the unified
+            -- border (frame level + 10, same inherited strata). SetParent
+            -- resets frame level, so re-assert above the border after every
+            -- reparent. This is the path that previously dropped the bar below
+            -- the border (it forced hpLevel+2 = frame+4 here too).
+            frame.Power:SetFrameLevel(frame:GetFrameLevel() + 12)
         else
             if frame.Power:GetParent() ~= clip then
                 frame.Power:SetParent(clip)
             end
+            -- Power bar must render above absorb overlay (health level + 1).
+            -- SetParent resets frame level, so re-assert after every reparent.
+            local hpLevel = frame.Health and frame.Health:GetFrameLevel() or clip:GetFrameLevel()
+            frame.Power:SetFrameLevel(hpLevel + 2)
         end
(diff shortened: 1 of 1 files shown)


Aderfi:
image

image

Aderfi:
Not compatible with 8.0 changes and no longer necesary

PR #362, opened 2026-06-07
feature: Tooltips for auras/buffs/debuffs in party and raid frames

pereira-a:
Show spell tooltips when hovering aura, buff, debuff, and defensive icons on raid frames. This is helpful to quickly learn or remember the effects of specific spells (either during combat or out of combat).

This implementation allows clicks or interactions to still affect the frame behind the icons. So no other features are broken (explanation below).

## Implementation

A single global throttled `OnUpdate` tracker maps cursor position to the visible icon underneath and calls `GameTooltip:SetSpellByID`. No scripts or `EnableMouse` are ever set on the child icon frames, so clicks pass through to the secure unit button behind them => targeting, right-click menus, and click-casting addons are completely unaffected.

When the cursor leaves the icon area but remains on the same unit frame, the unit tooltip is immediately restored.

## Settings

Respects both tooltip toggles from the **"Extas"** section from "Raid/Party" settings panel in "Raid Frames":

| Setting | Behavior |
|---|---|
| `showTooltip` | When off, no tooltips of any kind are shown. The tracker is never started. |
| `tooltipInCombat` | When off, spell tooltips are suppressed in combat (same as the existing unit tooltip toggle). |

### DIFF
--- a/EllesmereUIRaidFrames/EUI_RaidFrames_BuffManager.lua
+++ b/EllesmereUIRaidFrames/EUI_RaidFrames_BuffManager.lua
@@ -1105,17 +1105,20 @@ function ns.BM_UpdateSimpleGrid(button, unit, db, updateInfo)
         local iid = auraData.auraInstanceID
         local tex = auraData.icon
         local matched = false
+        local matchedSpellId
         if sid and iid then
             if not issecretvalue(sid) then
                 local psid = PRIMARY_BY_ALT[sid] or sid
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
this is a larger feature I am implementing in phases, I have now implemented debuff tooltips (with next patch) and private aura tooltips. I appreciate the help but need to be very careful with raid frames due to the complexity of the addon

PR #366, opened 2026-06-07
CDM: add "Hide Empty Healthstone" option

Trenchfoote:
Adds a per-bar toggle (CD/utility custom bars) that hides the Healthstone icon entirely when the player has none in bags, instead of showing it desaturated.

- Gated at injection in CollectAndReanchor so the layout reflows; only applies to the healthstone/demonic_healthstone presets, counting the preset's altItemIDs too. Layout changes are combat-locked, so in combat the icon simply stays desaturated until the next out-of-combat rebuild.
- A BAG_UPDATE_DELAYED listener triggers a rebuild when bag contents change (out of combat), guarded by a cheap scan (_AnyBarHidesEmptyHealthstone) so it only fires for users who enabled the option.


EllesmereGaming:

PR #367, opened 2026-06-07
Feat/class color background

Trenchfoote:


EllesmereGaming:
had to manually add due to a lot of conflicts, but it's in LETS GO

PR #370, opened 2026-06-07
Nameplates: add Border hover style + class-color option

Trenchfoote:
Extends the mouseover hover effect with a style choice and a class-color border, for enemy and friendly plates.

- New "Hover Style" dropdown: Health Bar Highlight (the existing bar-fill overlay) or Border (a pixel-perfect border around the plate). All show/hide sites route through a shared ns.ShowPlateHover(plate, shown) so the two elements never both show, and a style switch applies without a reload.
- The Hover Style cog is context-sensitive: Highlight exposes Opacity + Color (the former standalone "Hover Effect" controls, now consolidated); Border exposes Thickness + Color + Class Color.
- Class Color border uses the player's own class color (dims the custom Color picker while active); resolved at show time so it survives plate recycling. RefreshHoverEffect applies changes live.
- New DB keys (hoverStyle, hoverBorderSize, hoverBorderColor, hoverBorderClassColored) with defaults and per-profile sync.

### DIFF
--- a/EllesmereUINameplates/EUI_Nameplates_Options.lua
+++ b/EllesmereUINameplates/EUI_Nameplates_Options.lua
@@ -3567,7 +3567,7 @@ initFrame:SetScript("OnEvent", function(self)
             swatch:EnableMouse(not off)
         end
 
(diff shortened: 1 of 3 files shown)


EllesmereGaming:

PR #372, opened 2026-06-08
RaidFrames: Add dispel icon size

mexxman2:
Add a slider for dispel icon size because default was too big. Also found it weird that pretty much everything else is resizable except for this.


EllesmereGaming:

PR #374, opened 2026-06-09
Add option for absorbs growing from left edge of health bar.

RoastedRowlet:
absorb grows from left of frame to right.

image

At the moment its only absorbs. I couldn't seem to get healing absorbs on that side at the same time. Though it might be better this way?

image

made the option mutually exclusive with grow from right because they wouldn't make sense together.

For some reason this feature only works after a reload. I am not comfortable enough with the init code to feel comfortable changing it to refresh things.


EllesmereGaming:
this feature was actually already added in a prior update, but thank you for the PR!

PR #375, opened 2026-06-09
WIP RaidFrames: Implement change over time feature for health color

mexxman2:
Marking as WIP for now as I have yet to test it in a real environment.

It's a feature I've used a lot in Cell, but unfortunately it's kinda all over the place in Midnight, without any official updates.
Essentially it's an extension of the current Health Bar Color feature in the Buff Manager. With this you can assign 3 colors based on remaining duration. Can be a nice reminder to eg. refresh Atonements.
With current implementation:
1. Remaining duration is higher than 50%
2. Remaining duration is lower or equal to 50% (This was adjustable in Cell, but don't really think it matters too much)
3. Remaining duration is lower or equal to an adjustable number in seconds (1-5 seconds right now)


mexxman2:
Similar implementation was added, so I'm closing this

PR #378, opened 2026-06-12
QoL: Shows durability % on QoL Extras

xatzisktv:
## Summary

Adds an optional durability readout to the FPS counter so you can see your lowest equipped-item durability % alongside FPS, world MS, and local MS.

- Extends the FPS counter with a new `87% (dur)` segment that follows the same divider + value + label styling as the existing MS readouts
- Adds a **Show Durability on FPS Counter** toggle in the QoL **EXTRAS** section (disabled when the FPS counter is off)
- Extracts shared `GetLowestEquippedDurability()` logic used by both the FPS display and the existing Low Durability Warning, so both always report the same value
- Fixes missing `_lblWorld` / `_lblLocal` references so FPS label font sizing applies correctly

**Saved variable:** `EllesmereUIDB.fpsShowDurability` (defaults to off)

## Test plan

- [ ] Enable **Show FPS Counter** and confirm FPS (and optional world/local MS) still display correctly
- [ ] Enable **Show Durability on FPS Counter** and confirm `NN% (dur)` appears after the MS segments
- [ ] Damage gear or repair and confirm the % updates within ~1s
- [ ] Confirm the displayed % matches the value used by **Low Durability Warning**
- [ ] Toggle durability off — segment disappears, FPS counter remains
- [ ] Disable the FPS counter — durability segment is hidden and the durability toggle is dimmed with the "requires Show FPS Counter" tooltip
- [ ] Toggle world/local MS independently — durability always anchors after the last visible segment
- [ ] Confirm the FPS counter unlock/position entry still works with the wider frame

### DIFF
--- a/EllesmereUIQoL/EUI_QoL_Options.lua
+++ b/EllesmereUIQoL/EUI_QoL_Options.lua
@@ -590,6 +590,25 @@ initFrame:SetScript("OnEvent", function(self)
             end)
         end
 
+        -- Row 2: Show Durability on FPS Counter (left) | helper label (right)
+        _, h = W:DualRow(parent, y,
+            { type="toggle", text="Show Durability on FPS Counter",
+              tooltip="Shows lowest equipped-item durability % on the FPS counter.",
+              disabled=function()
+                return not (EllesmereUIDB and EllesmereUIDB.showFPS)
+              end,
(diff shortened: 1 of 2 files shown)


EllesmereGaming:

PR #392, opened 2026-06-18
Resource threshold colouring bugfix and UI update

asdfractal:
Bugfix: Resource threshold was not working correctly

Testing:
images of many specs tested with resource/power/health 
https://imgur.com/a/7EXW3dW

UI update:
Added a clearer text for toggle above/below for class resource bars that are not pip based
image

When creating a new threshold will set default below for spenders (eg; mana), default above for builders (eg; rage)


asdfractal:
If you just wanted the fix without any UI or above/below changes it was this
EllesmereUIResourceBars.lua:3539
```                            
curve = GetBarThresholdCurve(
                                _tsR or 1, _tsG or 0.2, _tsB or 0.2,   -- threshold color (above)
                                r, g, b,                                -- fill color (below)
                                _tsThreshCount or 30)
```

PR #411, opened 2026-06-20
[RaidFrames] introducing new frame level properties for individual components

maxzaleski:
This PR introduces user-configurable frame level control across most visual layers of the raid frames: name and health text, status overlays, indicators, aura icons, and Buff Manager indicators. 

Previously, every component's z-order was hardcoded to a fixed offset constant (such as `ns.LVL_AURA`, `ns.LVL_RAISE`), which meant users who repositioned the name text to a centre or bottom slot had no way to resolve the resulting overlap with ready-check icons, status text, or aura icons short of a source edit.

<details>
<summary>In-game example (old)</summary>

image

Targeted Spells, and indicators like Ready-Check / Summon Pending would overlap with the display name. Additionally, should one prefer to have debuffs or other auras outside of frame boundary, the hover/target border would cut these off.
</details>

<details><summary>In-game example (new)</summary>
<p>
image

Every component is displayed correctly. One will also notice on frame 2 (top-down, warrior) that both a private aura and a targeted spell are rendered at the same time; I set the TS to have a higher priority than PA. 

</p>
</details> 

## New scale

eui_rf_new_scale

| Key | Offset | Targets (by default) |
|--------|--------|--------|
| Lowest | +12 | Text (name, health) | 
| Low | +13 | Indicators | 
| Medium | +14 | Auras (defensives, debuffs, targeted spells) |
| High | +20 | Raised border(s) |
| Highest | +22 | Above border(s) | 

The philosophy behind the new scale was to maintain backwards compatibility as much as possible while providing some flexibility to the user. 

Text, indicators, and auras are now distributed over 3 distinct levels representative of their visual priority, +12, +13, and +14 respectively. The implication is that Indicators and auras no longer share the same `ns.LVL_AURA` level; auras still have a sufficient level buffer (from +14 to +20) to allow for flexible rendering. The ~~threat~~/hover/target border remains unchanged at +20. The raid-marker "always on top" behaviour also remains unchanged; however, the user can now set other elements to that same level (such as my debuffs in the prior section).

> Correction (21/06): threat border was originally set to +10; commit [bda1dd5](https://github.com/EllesmereGaming/EllesmereUI/commit/bda1dd535d79f3bbdd8d48d374881dd1092706b6) adds dropdowns for both threat and dispel borders, these are now set to "medium" and "high" respectively.

###   Buff Manager aligned, not redesigned. 

BM indicators already had a per-indicator frame-level dropdown (behindBorders → highest). Rather than adding a new pool-level key on top of that, `FRAMELVL_BASE`'s three shared tiers (medium, high, highest) is now aligned to `ns.FRAMELVL` (new scale), so BM indicators occupy the  same bands as every other component. The two BM-only fixed tiers (behindBorders=7, behindText=11) are intentional design positions - indicators explicitly placed there are exempt and unaffected. The text carrier offset was also made relative (off + 2) instead of pinned to the old fixed constant +18, which would have placed labels below their own icon at high or highest.

## Summary Changes

<details><summary>New profile properties</summary>

  | Property | Default | Function |
  |---|---|---|
  | `nameLevel` | `"lowest"` | `UpdateTextCarriersLevel` |
  | `healthTextLevel` | `"lowest"` | `UpdateTextCarriersLevel` |
  | `statusTextLevel` | `"low"` | `UpdateTextCarriersLevel` |
  | `roleIconLevel` | `"low"` | `UpdateIndicatorsLevel` |
  | `raidMarkerLevel` | `"highest"` | `UpdateIndicatorsLevel` |
  | `readyCheckLevel` | `"low"` | `UpdateIndicatorsLevel` |
  | `summonPendingLevel` | `"low"` | `UpdateIndicatorsLevel` |
  | `leaderIconLevel` | `"low"` | `UpdateIndicatorsLevel` |
  | `threatBorderLevel` | `"medium"` | `UpdateBordersLevel` |
  | `dispelBorderLevel` | `"high"` | `UpdateBordersLevel` |
  | `debuffLevel` | `"medium"` | `UpdateAurasLevel` |
  | `defLevel` | `"medium"` | `UpdateAurasLevel` |
  | `paLevel` | `"medium"` | `UpdateAurasLevel` |
  | `tsLevel` | `"lowest"` | `TS_ApplySettings` → `StyleIcon` |

</details> 

 Each controllable layer follows the same end-to-end flow: a new string key is added to the profile defaults using one of five named tiers that map to absolute frame level offsets via `s.FRAMELVL`. A corresponding Frame Level dropdown is added to the options UI, reading and writing that key through the standard SVal / SSet helpers and triggering a `ReloadAndUpdate`on change. 

On the render side, `StyleButton` creates a carrier frame for each layer and immediately calls a closure; `ns.FRAMELVL[s.<key>]` is read and applied relative to the unit button's current frame level. That closure is stored on `d` so both `ReloadFrames` and `ReloadPartyFrames` can re-invoke it when settings change without recreating any frames. The preview panel mirrors this: carrier frames are stored with `f._` prefixes at creation, and `ApplyPreviewData` re-levels them on every settings refresh so the panel stays in sync with the dropdowns in real time.

> [!IMPORTANT]
> Please note that `ReadyCheck` and `SummonPending` now have uncoupled texture frames.

## Screenshots

<details><summary>Settings</summary>
Screenshot_2026-06-20_185434
Screenshot_2026-06-20_185413

These patterns are repeated where applicable.
</details>

### DIFF
--- a/EllesmereUIRaidFrames/EUI_RF_TargetedSpells.lua
+++ b/EllesmereUIRaidFrames/EUI_RF_TargetedSpells.lua
@@ -204,6 +204,15 @@ local function StyleIcon(icon)
     local k = raid and 1 or (ns._partyIndicatorScale or 1)
     local sz = Setting(raid, "IconSize", 24) * k
     icon:SetSize(sz, sz)
+    -- Frame level: read from tsLevel/tsRaidLevel setting so the option takes effect.
+    local lvlKey = Setting(raid, "Level", "medium")
+    local parent = icon:GetParent()
+    if parent then
+        icon:SetFrameLevel(parent:GetFrameLevel() + ns.FRAMELVL[lvlKey])
+        if icon._borderFrame then
+            icon._borderFrame:SetFrameLevel(icon:GetFrameLevel() + 1)
+        end
+    end
     if icon._borderFrame then
         local PP = EllesmereUI and (EllesmereUI.PanelPP or EllesmereUI.PP)
         if PP and PP.UpdateBorder then
@@ -216,7 +225,6 @@ end
 local function CreateIcon(btn, raid)
     local icon = CreateFrame("Frame", nil, btn)
     icon._tsRaid = raid or false
-    icon:SetFrameLevel(btn:GetFrameLevel() + 12)
     icon:Hide()
 
     local tex = icon:CreateTexture(nil, "ARTWORK")
--- a/EllesmereUIRaidFrames/EUI_RaidFrames_BuffManager.lua
+++ b/EllesmereUIRaidFrames/EUI_RaidFrames_BuffManager.lua
@@ -105,7 +105,7 @@ local FRAMELVL_VALUES = {
 }
 local FRAMELVL_ORDER = { "behindBorders", "behindText", "medium", "high", "highest" }
 local FRAMELVL_BASE = {
-    behindBorders = 7,                -- below the main border (+8)
+    behindBorders = 7,                -- below the configurable threat/dispel borders (default high = +20)
     behindText    = 11,               -- below the name/health text carrier, above borders
     medium        = ns.FRAMELVL.medium,
     high          = ns.FRAMELVL.high,
@@ -855,7 +855,7 @@ function ns.BM_CreateIndicators(button, health, d, PP)
     -- Frame effect border
     local effectBorder = CreateFrame("Frame", nil, button)
     effectBorder:SetAllPoints(button)
-    effectBorder:SetFrameLevel(button:GetFrameLevel() + 11)
+    effectBorder:SetFrameLevel(button:GetFrameLevel() + ns.FRAMELVL.highest)
     effectBorder:Hide()
     if PP then PP.CreateBorder(effectBorder, 0, 1, 0, 1, 2) end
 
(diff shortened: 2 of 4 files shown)


EllesmereGaming:
wow this is an insanely well done PR, i'll look into this soon. quick question, does it make sure to set defaults to existing hardcoded values so user settings are completely unaffected unless they actively change one?

maxzaleski:
Hey Ellesmere, to my understanding `DeepMergeDefaults` will fill-in the new/missing keys at render time – I don't believe there to be any inherent conflicts with the existing layout (since most of strata setters used constants). The only caveat is that under the new scale, targeted spells are now "2" levels above what they were previously (text > indicators > TS/auras for visibility purposes). 

My understanding of the engine (EUI) as a whole is limited, but I have been consistently reloading my changes with an existing profile without any Lua errors. I will report back tomorrow with a fresh profile, and double-check that it behaves as expected. 

Edit: I would be happy to send you my current profile string should that be of any help to you.

maxzaleski:
Played with it for a reset now, feels stable. Only thing left now is to decide on the presets (notably around which border has priority over another).

EllesmereGaming:
Hey! Just wanted to say that I really appreciate your thorough work here i'm sorry EllesmereUI at the time just full launched and userbase skyrocketed and there were just too many things to do and this got lost in everything. but the PR and work was extremely appreciated, sorry this got buried in the chaos <3

PR #423, opened 2026-06-21
[RaidFrames] animated clock border and cast countdown timer for targeted spell icons

maxzaleski:
This PR introduces animated clock border and cast countdown timer for targeted spell icons. The former is a feature I enjoyed when Cell was still around (somewhat inspired), and the latter is a nice usability addition. 

Both features are showcased [here](https://youtu.be/CyYoaB3VDS0).

> [!NOTE]
>  While I have been tinkering with these features for a few days now, the landscape of targeted spells remains experimental. I would recommend spending a bit of time independently testing either feature to be somewhat confident with their stability. I have not encountered major downsides as of yet (degradation in performance, Lua errors, etc). 
> 
> If a more informed maintainer could look at my API utilisation that would be helpful for future development. I have a couple more PRs to submit; I hope to be able to finish them in the coming days. 

### Clock Border

A `CooldownFrameTemplate` frame is placed outside the icon edges by a configurable amount (default 3 px). The spell texture, rendered at a higher frame level, masks the interior such that only the outer ring is visible. The ring depletes clockwise as the cast drains, giving an immediate visual read on urgency without obscuring the icon.

- The default pie-sweep texture ignores `SetSwipeColor`, so `WHITE8X8` is used as the swipe texture and tinted to the user's chosen colour. Colour must also be re-applied after every `SetCooldownFromDurationObject` call, as that resets both the texture and colour.
- The black background for the elapsed arc lives on a separate sibling `Frame` at frame level 2 rather than on the `CooldownFrame` itself. WoW draws the cooldown swipe before standard textures on the same frame, so a `BACKGROUND` texture directly on the `CooldownFrame` renders above the sweep and obscures it.

> [!IMPORTANT]
> When `ShowClockBorder` is enabled, the static pixel-perfect border (`_borderFrame`) is
> hidden to avoid visual conflict with the ring.

#### UI Widgets

Toggle and colour swatch on the left, border size slider (1–10 px, default 3) on the
right. Per section (raid / party).

---

### Cast Countdown Timer

A second `CooldownFrameTemplate` is placed over the icon with the swipe disabled and the built-in countdown numbers enabled. It shows the remaining cast time as a ticking number.

- Reading `UnitCastingInfo` timestamps and doing the arithmetic in Lua is not viable as those values are secret. `SetCooldownFromDurationObject` passes the duration object directly to the C side, which handles all timing internally and never surfaces taint in Lua.
- Font size and colour are user-configurable. The internal `FontString` inside the`CooldownFrame` is not exposed via a named method, so it is discovered once at creation time via `GetRegions()` and stored as `_timerCdFS` for later use in `StyleIcon`. The `FontString` is also re-anchored directly to the icon frame to bypass any pixel-snapping offset baked into the template.

#### UI Widgets

Duration Timer toggle shares a row with the Show Targeted Spells dropdown, with its colour swatch and cog popup (text size, X/Y offset) inline on the right. Icon Size and Icon Spacing share the row below. Per section (raid / party).

## Screenshots

<details><summary>Settings</summary>
bvbbbb
</details> 

<details><summary>Stills from demo video</summary>
CleanShot 2026-06-21 at 21 09 28@2x
CleanShot 2026-06-21 at 21 09 07@2x
</details>

### DIFF
--- a/EllesmereUIRaidFrames/EUI_RF_TargetedSpells.lua
+++ b/EllesmereUIRaidFrames/EUI_RF_TargetedSpells.lua
@@ -413,7 +413,7 @@ local function LayoutButton(btn)
     local ox = Setting(raid, "OffsetX", 0) * k
     local oy = Setting(raid, "OffsetY", 0) * k
     local anchor = btn._health or btn
-    local spc = 2 * k
+    local spc = Setting(raid, "IconSpacing", 2) * k
     local spacing = sz + spc
 
     local shown = 0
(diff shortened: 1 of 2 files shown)


maxzaleski:
From reviewing the footage, I'm realising now that the new border bleeds onto the "personal space" of sibling icons. I will push a new commit tomorrow aiming to allow the spacing between icons to be customised. 

~~Edit (23/06): not had time to look into this just yet. Hoping to get this done in the next couple of days.~~

Added as part of  [753a535](https://github.com/EllesmereGaming/EllesmereUI/pull/423/commits/753a5351ee10146fc99d76fac58e958196b61c5e). Additionally, preview now includes new settings.

maxzaleski:
Paused for now. Exboss have come out with their own version of targeted spells; timers work where mine fall short.

Current state:
- Cast timers are consistent with non-elite trash.
- Cast timers are inconsistent on elite trash (e.g. in MT: Sentry's Ethereal Shackles won't work; Pyromancer's Pyroblast works)

PR #433, opened 2026-06-22
Here’s a fairly complete zhCN translation.

myxiaomenya-glitch:
I found that using the original zhCN code would cause most translations to fail to load. So I revised the code to load this Lua file automatically when it detects the WoW client language is set to zhCN.


Shiyan66666:
哥么求你别偷我代码乱发了，这还是我好几个版本前的代码了，很多翻译也是我自己都没核对的。

PR #440, opened 2026-06-23
Fix damage meters ignoring locale translations

myxiaomenya-glitch:
Previously, EllesmereUIDamageMeters had all display strings hardcoded in English, completely ignoring the L() locale system. zhCN, zhTW, and koKR locale files already contained translations for all damage meter strings, but they were never called.

Changes:
- Added `local L = _G.EllesmereUI.L` reference
- Replaced all hardcoded English strings with L() calls, including:
  - Window titles (Damage Done, Healing Done, etc.)
  - Right-click context menu (Damage, Healing, Actions, Off)
  - Mythic+ auto-start config menu
  - Session type labels and "Overall" prefix
  - Window management tooltips (New Window / Close Window)

No risk to other languages — L() returns the original English string as silent fallback when a key is missing from a locale file (deDE, esES, frFR, itIT, ptBR, ruRU).


EllesmereGaming:
couldn't auto merge due to formatting issue, but I manually merged so all the changes will be in next patch thank you!

PR #445, opened 2026-06-24
Add Mage Remove Curse to HoverCast dispel bindings

nulltyto:
## Problem

Mages have no entry in `DISPEL_SPELLS` (`EUI_RaidFrames_ClickCast.lua`), so a mage who binds the Raid Frames → HoverCast **dispel** action gets:

- An **empty macro** — `BuildBaseMacroText` loops the table for `sp.class == pClass`, finds no match, and returns `nil`, so the bind casts nothing.
- The **generic dispel icon** — `CC_GetBindingIcon` likewise finds no match and falls back to `ACTION_ICONS.dispel`.

So a mage could see dispellable debuffs highlighted on frames but their HoverCast dispel bind did nothing.

## Fix

Add Mage **Remove Curse** (spell ID 475) to `DISPEL_SPELLS`. It's baseline for all three specs, so a single entry covers the class. Both the macro builder and the icon resolver read the same table, so the bind now casts and shows the correct spell texture.

```lua
{ id = 475, name = "Remove Curse", class = "MAGE" },  -- All specs (Curse only)
```

## Notes

- Remove Curse only clears **Curse** debuffs (mages can't dispel Magic), which is correct/expected. The 'Show All' highlight will still surface Magic debuffs a mage can't act on — unchanged behavior; users can switch to 'By Me' if they prefer.
- Single appended row; no existing entries reordered. Other classes are unaffected since the read loops filter by `sp.class == pClass`.

## Testing

Verified in-game on a mage: the Q dispel bind now shows the Remove Curse icon and casts Remove Curse on a mouseover target.


EllesmereGaming:
added this locally since it was just one line and i'm in the middle of editing these files, ty!!

PR #449, opened 2026-06-24
Fix: debuff position Y offset ignored in unit frame preview

LoChinAn:
## Summary

The unit frame **options preview** hardcoded the debuff **Y offset to `0`**, so dragging the *Debuff Position* **Y** slider produced no visible change in the live preview thumbnail, while the **X** offset worked. The real in-game frames already honored the Y offset, so the preview disagreed with the actual result and looked broken.

## Change

In `EllesmereUIUnitFrames/EUI_UnitFrames_Options.lua` (debuff preview repositioning):

- Use `s.debuffOffsetY` for `dOffY` instead of the hardcoded `0`.
- For Boss *Simple Debuff Display* mode, read both X and Y from `ns.GetBossSimpleDebuffOffset(s)` (it already returns `x, y`).

`dOffY` now flows into `anchorMap` (the per-anchor `oy`) and the `anchorKey` cache, so the preview repositions live and matches the real frames.

## Testing

Verified in-game: dragging the Debuff Position Y slider now moves the debuff icons in the options preview, consistent with the actual unit frame.


EllesmereGaming:
I am implementing a more involved version of this to try and mitigate preview height issues while still reflecting accurately. thank you!

PR #454, opened 2026-06-24
CDM: Change cd ready glow function to also check for is spell usable

JuJuFX-dev:
## Summary
The CD Ready Glow (`pixelGlowReady` / `buttonGlowReady`) previously activated
whenever a spell came off cooldown, regardless of whether the player had enough
resources to cast it. This caused the glow to fire even when e.g. Fury was too
low, which was misleading.

## Changes

**`EllesmereUICdmHooks.lua`**

### 1. `HasEnoughResources(spellID)`
New helper that cross-checks the spell's resource cost (`C_Spell.GetSpellPowerCost`) against live `UnitPower()` / `UnitPowerMax()`. It is **ANDed with `C_Spell.IsSpellUsable()`**, never used standalone — `IsSpellUsable` stays the primary castability signal (cooldown/form/lockout), `HasEnoughResources` is a defensive second opinion for the one window where `IsSpellUsable` can't be trusted (see below).

### 2. Login/reload data settling: `ns._cdGlowPowerConfirmed`
`C_Spell.IsSpellUsable()` can briefly report a resource-gated spell as usable immediately after `PLAYER_ENTERING_WORLD`, before the client's power data has actually synced. Instead of guessing a fixed delay, a flag is set `false` on `PLAYER_ENTERING_WORLD` and only flips `true` on the next genuine `UNIT_POWER_UPDATE`. Resource-gated glow decisions are suppressed while the flag is `false`. If a resource never changes (e.g. stays at 0), the glow correctly never lights — which is the right outcome, not a bug.

### 3. Explicit cooldown-stop safety net
The event-driven update loop (`SPELL_UPDATE_COOLDOWN` / `UNIT_POWER_UPDATE`) only ever handled the "not on cooldown" branch; nothing explicitly stopped the glow once a spell *went* on cooldown if the `SetDesaturated`-based hook didn't happen to fire for that transition. Added an explicit `onCD` branch that always stops the glow, independent of resource state.

### 4. Per-frame spell-binding tracking
Blizzard's CDM viewer frames are pooled and get reassigned to a different spell over time. Glow state (`_cdStateGlowOn`) was tracked per **frame**, not per **(frame, spell)** — so a stale "on" state from a previously-displayed spell could bleed into a newly assigned spell and sit there until an unrelated event happened to trigger a re-check. A `_cdGlowBoundSid` tag now forces an immediate reset whenever the bound spell changes.

### 5. Perf: cached viewer-frame lookup
Added a small `GetViewerFrame(vi)` cache to avoid repeated `_G[...]` lookups in both the CD Ready Glow update loop and the existing 10 Hz active-aura ticker (`BuffTicker`) — the four CDM viewer frame objects never change identity for the session.

**`EllesmereUICooldownManager.lua`**

### 6. Closed a third, independent glow-start path
Beyond the two paths above, the icon decoration/rebuild logic (which runs during reanchor/reflow, e.g. on `/reload`) had its **own** glow-start logic that checked only "not on cooldown" — completely bypassing the gating in `EllesmereUICdmHooks.lua`. This was the actual cause of the glow staying lit permanently after a `/reload`: it was started by a code path with no event listener ever wired up to re-evaluate it. Both glow-start sites here now apply the same `IsSpellUsable` + `HasEnoughResources` + `ns._cdGlowPowerConfirmed` gate.

## Why `IsSpellUsable` *and* `HasEnoughResources`, not just one
- `IsSpellUsable` alone: correct almost all the time, but briefly wrong right after login/reload (the original bug).
- `GetSpellPowerCost` alone: unreliable for some spells — Void Ray reports `cost = 0` even outside Void Metamorphosis (where it should cost 100 Fury), so a pure cost-comparison fix isn't safe by itself.
- Together, with `IsSpellUsable` as the gate and `HasEnoughResources` as an additional AND-condition behind `ns._cdGlowPowerConfirmed`: keeps `IsSpellUsable` as the primary signal, while specifically covering the login-timing window where it can't be trusted.

## Behaviour
- Login/reload with insufficient resources → glow stays off, no flash, no stuck-on state.
- Spell comes off cooldown but resources are insufficient → glow stays off.
- Spell goes back on cooldown → glow stops immediately, regardless of resource state.
- A pooled viewer frame gets reassigned to a different spell → glow state resets immediately instead of inheriting the previous spell's state.

Recreated the pull request cause of some problems with forked and local repo.


EllesmereGaming:
hey! this is awesome and i really appreciate it, but while functional it has performance impacts that i'm not willing to compromise on at this time. Unconditional polling like this is something I try to avoid at all costs unless the feature is absolutely essential. if you can find a way to do this without impacting performance then I would be happy to add it!

JuJuFX-dev:
hey! thank you for the feedback. I will start working on a less performance impacting solution and come back to you asap.

JuJuFX-dev:
> hey! this is awesome and i really appreciate it, but while functional it has performance impacts that i'm not willing to compromise on at this time. Unconditional polling like this is something I try to avoid at all costs unless the feature is absolutely essential. if you can find a way to do this without impacting performance then I would be happy to add it!

Hey, I just uploaded a new version with performance improvements and an improved function for the check. Feel free to check it out. If I get some more time, I’ll keep working on it and see if there’s still room for further optimization.

JuJuFX-dev:
@EllesmereGaming Hey, I did some more testing, and I think this is the best iteration I can come up with for now. I don't think there's much room left for improvement. I'd love to see this get implemented. 🙂

EllesmereGaming:
taking a look now! thanks for the hard work :D

PR #455, opened 2026-06-25
feat: add public external module registration API

xanahopper:
## Why

I plan to port ElvUI_WindTools into EllesmereUI_WindTools, and that requires a supported way for third-party addons to integrate with EllesmereUI as native modules without relying on internal allowlists or fragile implementation details.

This PR adds a versioned extension surface so external modules can register their own sidebar entry, options pages, and lifecycle callbacks while keeping the host options UI protected from module errors.

## What Changed

- Added `EllesmereUI.API_VERSION` as the compatibility contract for external modules.
- Added `EllesmereUI_ExternalModules.lua` and loaded it from `EllesmereUI.toc`.
- Added `EllesmereUI.RegisterModuleCallback` and `EllesmereUI.FireModuleCallback` for host lifecycle events.
- Added `EllesmereUI.RegisterExternalModule(spec)` for third-party addons to register:
  - sidebar metadata
  - options page config
  - page builder callbacks
  - reset/cache-restore callbacks
  - API version requirements
- Added caller-folder validation so an addon can only register itself.
- Wrapped external module callbacks with `xpcall` so module failures surface through the normal error handler without breaking the EllesmereUI options panel.
- Added an `External` sidebar group for registered third-party modules.
- Marked external modules as profile-sync exempt to avoid misleading sync UI.
- Fired a `ProfileChanged` callback when EllesmereUI profiles are switched so external modules can rebuild cached profile state.
- Split the internal module config writer into `_RegisterModuleConfig` so both official modules and validated external modules can share the same registration path.

## Note

This PR was generated with AI assistance, and has been manually tested and reviewed.

Test Case like this:

e59dc0d86f2bca282bc0454b8bb61785

### DIFF
--- a/EllesmereUI.lua
+++ b/EllesmereUI.lua
@@ -1326,6 +1326,13 @@ EllesmereUI.ADDON_ROSTER = ADDON_ROSTER
 EllesmereUI.LOCALE_FONT_FALLBACK = LOCALE_FONT_FALLBACK
 EllesmereUI.EXPRESSWAY = LOCALE_FONT_FALLBACK or EXPRESSWAY
 
+-- External Module API version. Bumped whenever the RegisterExternalModule
+-- spec schema, buildPage contract, or callback event names change in an
+-- incompatible way. External modules declare the API they targeted via
+-- spec.apiVersion and are rejected/marked on mismatch. See
+-- EllesmereUI_ExternalModules.lua for the full implementation.
+EllesmereUI.API_VERSION = 1
+
 -- Taint-safe print. Uses AddMessage instead of the global print(), which
 -- routes through Blizzard's C-side handler and taints the chat frame
 -- execution context. Silently drops the message inside protected instances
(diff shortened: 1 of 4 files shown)


xanahopper:
hello~ can this PR be merged?

EllesmereGaming:
Thank you for the PR! The implementation itself looks solid, however, I'm not going to merge this one. Early on with EUI I had external modules integrated into the main options, and it caused a lot of confusion. Users saw them as a part of EUI, so I'd get constant bug reports and support requests for addons I didn't actually maintain.

I want to avoid that at least for now while EllesmereUI is still fairly new and people are still learning the system fully. I'd be much more open to it after a few months of the playerbase getting accustomed to it but at this time it's a bit too early for me to want to manage this. I don't mind people making external plugins, many already do, but I want them to stay separate from the core EUI panel so it's clear what's officially part of EUI and what isn't. I just don't have the time right now to support the extra confusion that comes with that more aggressive integration currently.

xanahopper:
Sure, I get it. I may maintain a branch/fork for my own and friends. or I just port WindTool function to EllesmereUI directly?

PR #458, opened 2026-06-25
fix: warrior stance reminders in Aura/Buff Reminders

nulltyto:
## Summary

Fixes the Aura/Buff Reminders warrior stance behavior reported on Discord: https://discord.com/channels/585577383847788554/1518276017930637432

Reported symptoms:
- **Fury** always showed Berserker Stance, even while in it
- **Protection** always showed Defensive Stance, even while in it
- **Arms** showed no stance at all

## Root cause

Warrior stances are **shapeshift forms**, not auras, so `C_UnitAuras.GetPlayerAuraBySpellID(386196/386208)` never detected the active stance. The reminder therefore reported the stance as permanently "missing" (Fury/Prot). Arms showed nothing because the gate used `IsPlayerSpell`/`IsSpellKnown` on the cast spell, which didn't report for it.

## Fix

Detect stances through the stance bar via `GetShapeshiftFormInfo`, which reports both whether a stance is **known** (present in the bar) and whether it is currently **active**. Each spec is mapped to its stance:

| Spec | Stance |
|------|--------|
| Arms | Battle Stance (386164) |
| Fury | Berserker Stance (386196) |
| Protection | Defensive Stance (386208) |

The reminder now hides once the desired stance is active, and is suppressed entirely when the stance isn't known.

## Notes
- `battle_stance` added to default `auras.enabled`; `DeepMergeDefaults` fills it for existing profiles. The options toggle auto-generates from the aura list.
- Stance checks remain out-of-combat only (existing behavior).
- `Locales/_keys.txt` regenerated (picked up a pre-existing `Server Time` key).

## Testing
Verified in-client that each spec shows its own stance and hides it once that stance is active.


nulltyto:
Not tested in PVP. Not sure if there is anything relevant here b/c I believe there is PVP fallback.

PR #460, opened 2026-06-25
fix pandemic glow for lifebloom, add marker on bar

asdfractal:
Fix inconsistent pandemic glow for Lifebloom on player.
Lifebloom on another target cannot glow because of secret value - instead add an optional marker enabled just for Lifebloom buff bar at 30% of bar width to know when pandemic/bloom is available.

Tested extensively with dungeon, raid, mythic plus, /euidev.
Screenshot 2026-06-25 232202
Screenshot 2026-06-25 232155
image


asdfractal:
reworking with better solution

PR #464, opened 2026-06-26
fix(aura-reminders): show spell names in the client's locale

LoChinAn:
## Problem

In **Aura Buff Reminders**, the names shown for raid buffs, auras, rogue poisons, paladin rites and shaman imbues — both the in-game reminder labels and the options checkbox grids — come from a hardcoded English `name` field on each data entry. That string is passed straight to `ShortLabel(...)` / `SetText(...)` without going through `EllesmereUI.L()` **or** the spell API, so the names always display in English even on non-English clients (and community locale translations for them never apply).

Everything else on the same panel is already localized (headers, toggles, zone names via `EllesmereUI.L(zoneName)`, talent names via `C_Spell.GetSpellName`) — only these spell labels are not.

## Fix

Resolve each label from its spell ID with `C_Spell.GetSpellName(spellID)`, keeping the existing hardcoded English `name` as a fallback when the API returns `nil` (e.g. the spell isn't cached). This matches the approach already used for flask names (`C_Spell.GetSpellInfo(buffID).name`).

The resolver is a single shared helper exposed as `_G._EABR_SpellName(spellID, fallback)` (same pattern as the existing `_G._EABR_*` shared functions), used by both `EllesmereUIAuraBuffReminders.lua` and `EUI_AuraBuffReminders_Options.lua`.

Result: on a zhTW / koKR / etc. client these now show the official localized spell names; on an English client they are unchanged.

## Scope

**Changed** (in-game labels + options grids): raid buffs, auras, rogue poisons, paladin rites, shaman imbues.

**Intentionally left as-is:**
- **Shaman shields** — custom labels (`Earth Shield (Self)`, `Lightning/Water Shield`) and `castSpellFn`; the class override already collapses them to `Shield`.
- **Flasks / foods / weapon-enchant items** — these are items, not spells (flask names are already localized elsewhere).
- Category words `Poison` / `Weapon` / `Shield` and the preview icon-row labels are EUI's own strings (not spell names) — possible follow-up.

## Implementation note

`_G._EABR_SpellName` is defined directly as a chunk-level global, **not** assigned from a local inside the init block. Assigning it from a local there left the global `nil` at runtime (the init closure is very large), which caused a nil-call error; defining it at chunk scope avoids that.

## Testing

Tested on a Rogue (`/reload`): the poison names in the options grid render correctly and there are no Lua errors.

This is a first-pass change and I've only verified a limited set of cases — I have **not** tested every class, an English client, or the API-`nil` fallback path.


EllesmereGaming:
there were significant changes to the codebase in the patch today, would you mind rebasing?

LoChinAn:
Thanks for the heads up! I'm currently a bit tied up, but I'll rebase and resolve the conflicts by June 30th.

LoChinAn:
Hi! I've successfully rebased onto the latest main branch and resolved the conflicts. It's ready for your review and merge!

LoChinAn:
Closing this one since the branch has become too desynced from the recent updates. I will submit a fresh, clean PR with the rebase changes shortly!

PR #491, opened 2026-06-28
Add group for tracking bars

senowak:
### Summary

Tracking Bars previously supported only a single grouped container. All grouped bars shared the same anchor, grow direction and spacing, making it impossible to separate different categories of buffs into independent layouts.

This change introduces support for multiple Tracking Bar groups. Each group behaves independently and can be positioned anywhere on the screen while preserving the existing grouped behaviour within that group.

### Changes

Tracking Bars now support multiple independent groups instead of a single global group.

### 1. Multiple Tracking Groups

Each Tracking Bar can now be assigned to one of several groups instead of only using a boolean "Group Tracking Bars" flag.

Each group maintains its own:

* anchor position
* grow direction
* spacing
* dynamic layout
* unlock/move frame

### 2. Independent Dynamic Layout

Visible bars are now reflowed independently for each group.

Within a group:

* configured bar order remains the priority order
* inactive/hidden bars are skipped
* active bars are packed together without leaving gaps

Groups do not affect each other, allowing different buff categories to live in completely separate areas of the UI.

### 3. Backwards Compatibility

Existing configurations continue to work.

Legacy grouped Tracking Bars are automatically treated as belonging to the default group, so existing layouts require no manual migration.

Behaviour

Example:

Group 1 (Defensives)

* Ironfur
* Barkskin
* Survival Instincts

Group 2 (Offensives)

* Berserk
* Tiger's Fury

If only Barkskin is active:

```
Barkskin
```

When Ironfur becomes active:

```
Ironfur
Barkskin
```

The Offensive group remains unchanged.

Likewise, activating or removing buffs in Group 2 has no effect on Group 1.

image

### Why

Many classes benefit from separating different categories of buffs rather than displaying every Tracking Bar in a single list.

Supporting multiple independent groups allows defensive, offensive, utility or encounter-specific buffs to be positioned separately while preserving the existing grouped behaviour and priority ordering within each group.

### DIFF
--- a/EllesmereUICooldownManager/EUI_CooldownManager_Options.lua
+++ b/EllesmereUICooldownManager/EUI_CooldownManager_Options.lua
@@ -2963,86 +2963,102 @@ initFrame:SetScript("OnEvent", function(self)
         parent._showRowDivider = true
 
         -------------------------------------------------------------------
-        --  BAR GROUPING (shared across all bars)
+        --  BAR GROUPING (multiple independent groups)
         -------------------------------------------------------------------
         _, h = W:SectionHeader(parent, "BAR GROUPING", y);  y = y - h
 
-        -- Group Tracking Bars (per-bar checkbox dropdown) | Grouped Grow Direction
-        -- The checkbox dropdown lists every bar; checked bars chain together and
-        -- share width/height, unchecked bars are independent. Grow/spacing apply
-        -- to the chain and only matter once 2+ bars are checked.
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
hey! I have actually been working on a tracking bars entire system for a bit now so unfortunately can't pull this but this feature will be a part of the new upgraded tracking bars! Thank you so much for the effort it is super appreciated

PR #494, opened 2026-06-28
Rework GCD bar: per-frame probe polling + spell-queue visuals

nulltyto:
## Summary

Reworks the GCD bar so it can no longer desync, and adds spell-queue-window visuals plus fill-style options.

Bug report: https://discord.com/channels/585577383847788554/1520554339482075176

### GCD detection: events → per-frame probe polling
The bar previously captured the GCD from `UNIT_SPELLCAST_*` events in an `OnEvent` handler. Spam, queued casts and late/out-of-order events could leave the bar stuck or restart it mid-fill. It now polls the GCD probe (spell `61304`) every frame in `UpdateGCDBar` — the same approach the cursor GCD ring and Blizzard's action-bar swipes use. A new window is detected **only** by the probe's `startTime` changing, so there's no event gate, dedup anchor or stop-handler left to desync. The true hasted GCD is derived as `duration / modRate`.

### New: spell-queue-window visuals
Highlights the final stretch of the GCD where pressing the next spell queues it (size from `C_Spell.GetSpellQueueWindow`):
- **Show Queue Window** — translucent overlay band, with colour
- **Two-Color Bar** — tints the queue slice of the fill itself (opaque, clipped) instead of an overlay
- **Show Queue Tick** — thin boundary line, with width + colour

### New: fill-style options
- **Smooth Fill** (default on) — eased fill; off = the fill tracks the true GCD position 1:1 so it lines up exactly with the queue marker
- **Reverse Fill** — starts full and drains; the GCD-completion (business) edge and queue window stay fixed while full↔empty swap. Spark and queue visuals follow correctly across all four orientations.

## Notes
- `Locales/_keys.txt` regenerated via `.tools/extract-locale-keys.sh` (drops one stale key; new option strings are wrapped at the widget level, consistent with the rest of the options file).
- Self-reviewed: fixed a falsy-zero colour-channel bug (a `0` red/green/alpha was discarded for the default) and hoisted a per-frame `pcall` closure out of the `OnUpdate` hot path.

## Testing
- [ ] In-client smoke: spam casts, queued casts, instant-only, instance-only, Always Show
- [ ] Queue overlay / two-tone / tick across all four orientations
- [ ] Smooth + Reverse fill combinations

### DIFF
--- a/EllesmereUIResourceBars/EUI_ResourceBars_Options.lua
+++ b/EllesmereUIResourceBars/EUI_ResourceBars_Options.lua
@@ -7126,6 +7126,59 @@ initFrame:SetScript("OnEvent", function(self)
         _, h = W:Spacer(parent, y, 16);  y = y - h
 
         -----------------------------------------------------------------------
+        _sec, h = W:SectionHeader(parent, "SPELL QUEUE", y);  y = y - h
+
+        local queueOff = function() local p = DB(); return not p or not p.gcdBar.enabled or not p.gcdBar.queueWindow end
+        local tickOff  = function() local p = DB(); return not p or not p.gcdBar.enabled or not p.gcdBar.queueTick end
+
+        -- Row: Show Queue Window | Queue Color
+        _, h = W:DualRow(parent, y,
+            { type = "toggle", text = "Show Queue Window",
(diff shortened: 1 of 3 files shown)


nulltyto:
Changes landed in main between my pull and PR.

PR #518, opened 2026-06-29
CDM trinket: drop CVar toggle, use HideShoppingTooltips only

Filpet96:
## Summary

Follow-up to #515. The CDM trinket tooltip previously suppressed the equipped-item comparison (shopping) tooltips by temporarily flipping the `alwaysCompareItems` CVar on the user's behalf. Per review feedback (avoid setting CVars for users), this drops the CVar toggle entirely and relies solely on `GameTooltip_HideShoppingTooltips` to hide the comparison for the player's own equipped trinket.

## Changes

- Removed the `GetCVar`/`SetCVar("alwaysCompareItems")` save/disable/restore blocks around `SetHyperlink` in the trinket frame `OnEnter` handler.
- Keep `GameTooltip_HideShoppingTooltips(GameTooltip)` after `Show()` to hide the comparison tooltips.

## Test plan
- [x] Hover CDM Trinket Slot 1/2 → equipped trinket tooltip shows real item level/stats, no comparison tooltip
- [x] Anchor to Cursor ON and OFF both behave correctly
- [x] `alwaysCompareItems` CVar is left untouched (normal item comparisons elsewhere unaffected)

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
+++ b/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
@@ -2052,18 +2052,7 @@ local function GetOrCreateTrinketFrame(slotID)
             -- SetItemByID is only a fallback if no link is available.
             local link = GetInventoryItemLink("player", self._trinketSlot)
             if link then
-                -- Suppress the equipped-item comparison (shopping) tooltips: this is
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
already added locally

PR #540, opened 2026-07-01
Bags: Refactor font handling in bank to match bags

Snsei987:
There was a mismatch between certain font (for example item count was using Frizqt regardless of the configuration) between bank and bags.

image


EllesmereGaming:
manually merged this PR since i merged the previous one, ty!

PR #543, opened 2026-07-02
Friends: add offline filter and sort options

ZiiMs:
Add Friends List settings for showing/hiding offline friends and sorting by status,
name, game, level, or zone.

Post-process Blizzard's native FriendsFrame data provider so existing groups
like Favorites stay intact while visible rows are sorted within each group.

image

### DIFF
--- a/EllesmereUIFriends/EUI_Friends_Options.lua
+++ b/EllesmereUIFriends/EUI_Friends_Options.lua
@@ -196,6 +196,14 @@ initFrame:SetScript("OnEvent", function(self)
         "blizzard", "modern", "pixel", "glyph",
         "arcade", "legend", "midnight", "runic",
     }
+    local FRIEND_SORT_VALUES = {
+        status = "Status",
+        name   = "Name",
+        game   = "Game",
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
hey! while this would be a great feature, and actually EUI has had it in the past, it along with friend grouping feature had to be removed due to lua taint errors. these are the exact same lua errors every addon that tries to add this functionality faces (trying to modify the friends list scrollbox by using setdataprovider). for more niche addons this is an acceptable limitation but for EUI, this would introduce significant maintenance and code complexity and reintroduce lua taint which I have a hard rule again, I will not integrate any feature that lua errors. 

If you'd like to test this out as an example, download the addon FriendGroups, zone into a follower dungeon with EUIDev mode enabled (or zone into an m+, euidev mode just simulates the secret environment of an m+), set your WoW settings to whisper = new tab, so that whispers create new tabs, and then while in combat open your friends list, scroll to someone and whisper them.

EllesmereGaming:
(euidev can be activated with /euidev)

PR #548, opened 2026-07-02
Fix unitframe detached powerbar border

JuJuFX-dev:
## Summary

Fixed the border not showing when selecting detached power bar on target unitframe

## Changes

- Added helper function **EnsurePowerBorder**


EllesmereGaming:
oops i already fixed this locally thank you though!

PR #561, opened 2026-07-03
feat(unitframes): add "Important Only" aura filter

Filpet96:
## Summary

Adds an **Important Only** option to the per-unit **Buff/Debuff Filter** dropdowns (Player / Target / Focus) and the **Boss Debuff Filter**.

When checked, only auras Blizzard flags as *important* — via `C_Spell.IsSpellImportant()` (encounter/boss debuffs) — are shown. This hides the clutter the user called out: Stagger, the lust/Sated debuff, and other self/utility auras, leaving just the debuffs that actually matter in an encounter.

This mirrors the "important only" filter people know from ElvUI, but implemented natively through Blizzard's aura-filter tokens rather than a curated spell-ID list.

## How it works

It's a new classification in the existing token-based filter system (secret-value safe, `IsAuraFilteredOutByInstanceID`), sitting right next to **Crowd Control**, **Big Defensive** and **External Defensive**. Checked classifications OR together at runtime as before.

- `ns.ResolveAuraFlags` now returns a 6th flag — `buffImportant` / `debuffImportant`
- `ns.ComposeAuraFilter` appends `|IMPORTANT` to the element's change-detection signature so a `ForceUpdate` fires when the toggle flips
- `ns.EUIAuraFilter` unions the `HARMFUL|IMPORTANT` / `HELPFUL|IMPORTANT` classification
- `ns.ApplyEUIAuraFilter` records the flag on the element
- Options: new `important` item added to the shared `filterItems` and the buff/debuff key maps in both the Player/Target/Focus block and the Boss block

`IMPORTANT` is a Blizzard aura-filter token added in patch 12.0.1.

## Compatibility

- New settings keys `buffImportant` / `debuffImportant` default to `false` — existing profiles are unaffected, no migration needed.
- Labels/tooltips fall back to English via `EllesmereUI.L` until translations are added, matching the existing filter labels.

## Testing notes

Static change only (no in-game run available in this environment). Behavior to verify in-game: with only **Important Only** checked on the Player Debuff Filter, boss/encounter debuffs show while Stagger and the Sated/lust debuff are hidden.


EllesmereGaming:
this filter was removed in 12.0.7

Filpet96:
> this filter was removed in 12.0.7

@EllesmereGaming  Yeah my bad was not supposed to PR this, i think claude code did that automatically while i was testing out claude desktop app xd

PR #564, opened 2026-07-04
Nameplates: tank role icon on group tank's friendly nameplate

rinleyg:
﻿## Summary
Adds an option to show a **tank role icon** above the nameplate of the party/raid member assigned the **TANK** role, making the tank easy to spot at a glance.

## Details
- Works in **both** friendly nameplate modes: name-only and health-bar.
- New controls in the Nameplates friendly settings:
  - **Show Tank Role Icon** toggle (default: on)
  - **Tank Icon Size** slider (default: 60)
- Icon is centered just above the unit name, using the `UI-LFG-RoleIcon-Tank` atlas (matching the raid-frame role icons).
- Role resolution uses `UnitGroupRolesAssigned`, with a **GUID-based fallback** to the roster token, since nameplate unit tokens don't always resolve the assigned role directly.
- Updates live on `PLAYER_ROLES_ASSIGNED` / `GROUP_ROSTER_UPDATE` and on plate add.

## Changed files
- `EllesmereUINameplates/EllesmereUINameplatesFriendly.lua` — role-icon rendering for both plate paths, event wiring, `IsGroupTank` helper, `ns.RefreshFriendlyRoleIcons`.
- `EllesmereUINameplates/EllesmereUINameplates.lua` — defaults `showTankIcon = true`, `tankIconSize = 60`.
- `EllesmereUINameplates/EUI_Nameplates_Options.lua` — toggle + size slider in the friendly section.

159 insertions, 0 deletions.

## Testing
Tested live in a Mythic+ group: the tank's nameplate shows the shield above their name in name-only mode; toggling the option and dragging the size slider both update live.

### DIFF
--- a/EllesmereUINameplates/EUI_Nameplates_Options.lua
+++ b/EllesmereUINameplates/EUI_Nameplates_Options.lua
@@ -2524,6 +2524,25 @@ initFrame:SetScript("OnEvent", function(self)
             refreshNameSwatches()
         end
 
+        _, h = W:DualRow(parent, y,
+            { type="toggle", text="Show Tank Role Icon",
+              tooltip="Show a tank role icon above the nameplate of your group's tank so you can find them easily.",
+              getValue=function() return DBVal("showTankIcon") ~= false end,
+              setValue=function(v)
+                DB().showTankIcon = v and true or false
+                if ns.RefreshFriendlyRoleIcons then ns.RefreshFriendlyRoleIcons() end
+              end },
+            { type="slider", text="Tank Icon Size", trackWidth=120,
+              min=16, max=100, step=1,
+              disabled=function() return DBVal("showTankIcon") == false end,
+              disabledTooltip="Enable Show Tank Role Icon",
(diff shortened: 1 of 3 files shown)


EllesmereGaming:
hey! appreciate the work, but a few issues. this would have a significant impact on performance, you're currently iterating through every friendly nameplate every refresh with a full GUID scan on each whenever the role isnt tank. you need a guid -> role map once per refresh although again anything that's iterating through nameplates is highly performance impactful, nameplates is by far the most cpu intensive module and we need to be hyper vigilant to keep it as lightweight as possible. also 2 other rules for PRs need to be fixed, 1 is that any PR submitting a new setting has to have 0 impact on existing behavior so you'll need to default this to disabled. 2 is that any new setting must be 0 cost unless enabled, that means no new events registered nothing at all done unless the user specifically enables that setting. 

sorry i gotta be so strict with PRs but if you make these changes feel free to resubmit and i'll take another look! thanks again

PR #571, opened 2026-07-05
Feat/resource bars upgrade

asdfractal:
**Major upgrade to resource bars and threshold**
[Short demo video of UI](https://streamable.com/tehk9q)

**New features**
  - Multi-band threshold editor: Enable multiple conditions with >= / <= logic
  - Talent gating for threshold, have multiple entries per spec with active talent choice
  - Threshold settings page (class resource only) for better management of settings and cleaner UI
    - Click threshold settings to open page, each entry on the left is a button, the right are the settings
    - Class resource hash/thresholds can toggle between value/percent on any applicable
  - Segmented "pill" switch widget adopted throughout (the pre-existing BuildSegmentedControl), with new square, height, and width options to better match UI style. Toggles between two always on values, visually clearly than using a switch toggle.
image

  - Simple/Advanced resource config: Allows more fine grained control of every resource setting (except any anchor/positional) per spec. 
  - Advanced page defaults to synced with simple (like raid/party), each section can be synced independently
  - Enable/disable will preserve per spec settings to easily toggle
  - Confirm-delete popup on the Advanced screen's Remove button
  - Druid form specific threshold colouring in advanced config
image

Technical aspect abstracts the section builders to a function, which decides whether it's using the simple or advanced page. The build logic itself is unchanged and uses function arguments to create the elements exactly the same on both pages.

**Bug fixes**
  - Fixed Ignore Pain (Prot Warrior) threshold, new secret-safe StatusBar-overlay coloring (IP isn't a real power type and its value is secret, like VDH).
  - Fixed Devourer hash meta-cap, hide hashes >39 in Void Meta. And fixed resource bar not displaying full when resource is full.
  - Fix generator thresholds not filling above value (evoker/warlock)

**New files**
  - EllesmereUIResourceBars/EUI_ResourceBars_Advanced.lua: Per-spec Advanced mode options
  - EllesmereUIResourceBars/EUI_ResourceBars_Helpers.lua: Abstract some repeated functions across resource bar options in an effort to reduce time to understand, mostly focused on changes I added with the option to add/reuse in the future

**Testing**
A large range of testing has been done by myself and other support team members across a variety of classes.
I have run mythic+ and raids with this config with no errors.
Class specific tests for edge cases:
  - Ironfur (Guardian) multi-band
  - Ignore Pain (Prot Warrior)
  - VDH
  - Enhancement shaman

UI testing has gone through various tests to ensure functionality:
- Redrawing after unlock mode (overlays especially)
- Dropdowns functioning and attaching to parent across multiple talent entries

**Current profile settings/backwards compatibility**
This has been designed to inherit current settings so there is no change to anyone's config.
Support members have validated that none of their settings have changed on loading these changes.
I have also imported profiles created prior to this with no changes, as well as loading profiles I had already created and not opened while developing this.

**Additional changes and fixes outside the feature**
 - General_options.lua: Fix sidebar highlight/accent on main UI not updating with theme change
image

 - Widgets:lua Fix a UI wide bug with dropdown parent and invalidation, did not seem to impact the UI outside my change but impacted the talent dropdown on multiple entries correctly attaching to the frame. This was the cause of the talent dropdown displaying behind the entry after first time it was destroyed - it lost the reference to the correct parent. Default behavior unchanged and tested across various elements in the UI. Code comments explain further.
 - Widgets: Adjust disabled overlay by 1px so it doesn't overlap popup borders.

**Future plans**
- Explore adding threshold/colouring on buff
- Possibly add multi-talent for a single entry
- Anything else we like to add for more fine grained control or expanding conditions to resource bars is enabled by this framework


asdfractal:
ah had an ignored folder the script was scanning - updated

PR #585, opened 2026-07-06
Update range crosshair item and treat Holy Paladin as melee

hakklo:
This PR implements two QoL changes for the range crosshair:

1. Replaced the 40-yard range check item **Happy Fun Rock** with **The Decapitator** (ID `28767`).
2. Treat **Holy Paladin** as melee (5-yard cutoff range) rather than a 40-yard range class.


EllesmereGaming:
i made it so both items work, and i made an optional toggle for hpal that you can now choose to show as melee instead of hardcoded.

PR #591, opened 2026-07-06
Update koKR 8.3.8, Shipping a locale-specific cut safely

labrie75:
Update koKR 8.3.8

```lua
if GetLocale() ~= "koKR" then return end   -- cut engine below is koKR-only; change this (and MAP) to reuse for another locale
```

Everything **below this guard** is a koKR-specific render/translation cut. The guard makes the
whole block a **no-op on any non-koKR client** — nothing below it runs unless the game client
is Korean — so shipping it is safe for every other locale.

To reuse the same cut for another language:

1. Change the locale code in this check (`"koKR"` -> target locale).
2. Update the `MAP` table it drives.

No other edits are required.


EllesmereGaming:
merged this but I don't want to install global hooks so i'm changing that engine to instead just be coded as an if block inside the minimap file

PR #593, opened 2026-07-06
feat(damagemeters): Add segment sync across windows and auto-switch to Current on combat

SamJin98:
## Overview

This PR adds two new per-window toggles to the Damage Meters window settings menu (gear icon), sitting next to the existing **Auto Swap Current/Overall**:

**What's new:**

- **Sync Segment Selection** — windows with this on form a sync group: pick a segment (or Current/Overall) in one and the others follow. Windows without it are unaffected either way.
- **Auto Current on Combat** — if a window is sitting on a past segment when combat starts, it switches back to Current. Same idea as `auto_current` in Details. Windows on Overall are left alone.

Both options are per-window, default **off**, and persist across reloads. zhCN translations included.

Recording 2026-07-06 020332 (1)

## Background

- With several windows open, checking one fight means setting each window to that segment by hand — and afterwards they stay stuck there, showing stale data on the next pull. 
- Auto Current makes sure a pinned window snaps back to live data the moment a new fight starts.

### New per-window properties (stored in each window's config)

| Property | Default | Read by |
|---|---|---|
| `syncSegments` | `false` | `ns.ApplySegmentSelection` |
| `autoCurrentOnCombat` | `false` | `ns.AutoCurrentOnCombat` |

## Testing

Tested in-game (retail):

- [x] Sync with 2 windows on + 1 off: synced pair follows each other, third window independent both ways
- [x] Auto Current fires on trash pulls and boss pulls; Overall windows untouched
- [x] Both toggles off = behavior identical to before
- [x] Settings persist through /reload, per window
- [x] Runs alongside Auto Swap Current/Overall through an M+ start/finish without conflicts
- [x] Clean load, no Lua errors


EllesmereGaming:
ooooh i like this one, nice work!

PR #600, opened 2026-07-06
fix(cdm): trinket CD-ready sound on login + stop overlapping false CD-ready sounds

Filpet96:
## Summary

Two fixes to the Cooldown Manager's **Audio Effect on CD Ready** feature.

### Trinket CD-ready sound silent after login (until manually re-armed)

Trinket settings are keyed by the equipped item, but the trinket icon lives in an inventory slot (`-13`/`-14`). `FakeActive_Rearm` baked the slot into the rule once, but equipment is not queryable during the first CDM rebuild at login, so the rule kept the raw item key. The icon match and cooldown read then silently missed, so no sound ever fired — until the option was toggled off/on, which re-armed while equipment was readable (matching the reported workaround).

Fix: resolve the live trinket slot from the stable source item key on every poll in `EvalCdStateNow`, so it also survives trinket swaps without a full re-arm. The armed state still keys off the item id, so the Mythic+ rebuild-survival fix is preserved.

### Random overlapping CD-ready sounds "for no reason"

The spell path armed a sound on any "not ready" read and only gated the *fire* behind the loading/settle window. Across a boundary (zone/login and — newly — spec/talent swaps and full rebuilds), the cooldown API briefly reports ready spells as on-cooldown, which false-arms a whole batch that then fires together the instant the window ends — heard as several unrelated CD-ready sounds overlapping.

Fix:
- Gate **arming** (not just firing) behind the suppression window and clear stale arms while suppressed. A spell genuinely on cooldown across the window simply re-arms from its ongoing cooldown once the window ends, so no real ready edge is lost.
- Open the same settle window on every CDM full rebuild (`FullCDMRebuild`) so the re-prime that runs while the API is in flux can't false-arm a batch.

## Test plan

- [x] Set a CD-ready sound on an on-use trinket, `/reload` and relog — the sound fires when the trinket comes off cooldown (no toggle needed).
- [x] Swap trinkets between slots — the sound follows the item.
- [x] Set CD-ready sounds on several spells, then zone / take a portal / swap spec — no phantom burst of overlapping sounds.
- [x] Normal play: each spell still plays its sound once, exactly when it comes off cooldown.


EllesmereGaming:
merged manually

PR #605, opened 2026-07-07
Color wrapped cast icon border

Nnoggie:
Summary:
- Color the full-size cast icon border with the active target border color while wrap border is active.
- Reset the icon border to black when the wrap/full-size target-border condition is not active.

Before:
image
After:
image


EllesmereGaming:
manually merged this as well as i turned it into a setting to preserve user behavior. it'll be opt in within the inline cog of the "Spell Icon" setting

PR #625, opened 2026-07-08
Add fixed buff slots to raid frames

aizuon:
## Summary
- Add a Fixed Slots toggle for icon and square buff indicators in Raid Frames > Buff Manager > Display.
- When enabled, buffs keep their configured slot positions and missing buffs leave empty space instead of shifting later buffs forward.
- Keep existing compact behavior when the option is disabled.

## Testing
- Tested live in-game and it works.
Wow_lRAm2xGLh6


EllesmereGaming:
for this to work in 12.1 you need to make them separate indicators so i will be leaving that as the same functionality for live so people get used to setting it up that way

PR #630, opened 2026-07-08
feat(unitframes): per-unit frame source (EllesmereUI / Blizzard / Hidden)

SamJin98:
## Overview
A lot of players would rather use Blizzard's native unit frames, or keep some of them, instead of the EUI's styled ones. This PR adds a Frame Source dropdown on each unit frame (Player, Target, Focus, Pet, Target of Target, Focus Target, Boss). Each frame can be set to one of:

- EllesmereUI: the addon's styled frame (the default, same as now)
- Blizzard: Blizzard's default frame, left as-is
- Hidden: no frame

This lets you run EllesmereUI on some frames and Blizzard's default on others, or hide individual frames, without affecting the rest. Changing a source needs a /reload, since Blizzard frames can't be swapped in combat.

Target of Target and Focus Target are the one special case: Blizzard draws them as part of the parent Target/Focus frame, so "Blizzard" is only offered for them when the parent is also Blizzard. Mixing them (parent on Blizzard, child on EUI or Hidden) has a combat-only quirk, see Known issue at the bottom.

## Frame Source dropdown
image

## Disable other configs when 'Blizzard Default' or 'Hidden' is selected
image

## Each unit frame in each source
Hidden shows nothing, so no shot needed.

| Unit | EllesmereUI | Blizzard |
|------|-------------|----------|
| Player | image | image |
| Target | image | image |
| Focus | image | image |
| Target of Target | image | image |
| Focus Target | image | image |
| Boss | image | image |

## Known issue (not recommended)
Target on Blizzard with Target of Target on EllesmereUI (or Hidden). The addon hides Blizzard's built-in target-of-target out of combat, but it's a protected frame that can't be hidden during combat, so Blizzard shows it again for the whole fight, on top of your EllesmereUI one. There's no addon-side fix, so this mix is not recommended: keep Target and Target of Target on the same source. Focus and Focus Target behave the same way.

| Out of combat | In combat (Blizzard ToT reappears, not recommended) |
|---------------|-----------------------------------------------------|
| image | image |

## Testing
- Every unit (Player, Target, Focus, Pet, Target of Target, Focus Target, Boss) switched through all three sources; each takes effect after a /reload.
- EllesmereUI source: frames look and behave as before, no change to their existing settings.
- Blizzard source: the native frame shows and keeps its own behavior (unit menu, cast bar, buffs).
- Hidden source: the frame does not appear, and its settings page shows the inactive notice.
- Target of Target / Focus Target dropdown: offers all three options only when the parent is Blizzard, and drops to EllesmereUI / Hidden otherwise.
- Native Target of Target renders with Target and Target of Target both on Blizzard (against a target that has a target); Focus / Focus Target the same.
- Known issue confirmed: Target on Blizzard with Target of Target on EllesmereUI shows one ToT out of combat and both in combat, clearing when combat ends. Not recommended.
- Switching a source back and forth (EllesmereUI to Blizzard to EllesmereUI) restores cleanly after reload, with no leftover or duplicate frames.
- Existing profiles from before this change load unchanged (default is EllesmereUI on every unit).
- No Lua errors on load; luac -p passes on both changed files.

## Implementation notes
- ns.GetUnitFrameSource(unit) and ns.SetUnitFrameSource(unit, source) in EllesmereUIUnitFrames.lua handle resolution and persistence, and keep the existing enabledFrames[unit] flag and per-unit "never" visibility in sync.
- SuppressBlizzardChildFrame hides Blizzard's native ToT/FoT (unregister events, hide, re-hide on OnShow). The in-combat limitation is noted in a comment there.
- No CVars or FrameXML globals are written. An earlier version set showTargetOfTarget / showTargetOfTargetState; that was removed because the state cvar no longer exists in retail and writing those globals taints secure frames.
- Two files changed. Will rebase onto main before merge; no overlap with recent main commits.


SamJin98:
I've tested this quite a bit, but a few rough edges (mainly the in-combat Target of Target) come from Blizzard's protected-frame restrictions, not the addon, and can't be fully worked around. If you hit anything odd in an edge case, let me know.

EllesmereGaming:
this is awesome, amazing work. can you explain the rough edges a bit more? what exactly breaks?

SamJin98:
> this is awesome, amazing work. can you explain the rough edges a bit more? what exactly breaks?

Sure, here's what I found last night:

Blizzard's native ToT display logic is tied to `TargetFrame`, not to whatever we've set as the ToT source. `TargetFrameToT`/`FocusFrameToT` are children of `TargetFrame`/`FocusFrame`, not independent frames, so as long as Target's source is Blizzard, Blizzard keeps trying to show its own ToT regardless of what the ToT dropdown says. Whether that's actually visible depends entirely on combat state, because `TargetFrameToT`/`FocusFrameToT` inherit the same protected-frame restrictions as their parents - once combat starts, addon code can't Show/Hide/resize/reposition a protected frame, full stop (https://warcraft.wiki.gg/wiki/API_InCombatLockdown).

Out of combat, `SuppressBlizzardChildFrame` does exactly what it should, hides Blizzard's native ToT/FoT and re-hides on `OnShow`, so only our frame shows. 

The moment combat starts:

- Blizzard's own frame code still calls `Show()` on `TargetFrameToT` as part of its normal update cycle
- our `Hide()` in response is now an insecure call against a protected frame during lockdown, and gets blocked
- the native frame reappears for the rest of the fight, stacked on the EllesmereUI ToT if that's the chosen source, or just showing up on its own if set to Hidden
- the instant combat ends, `OnShow`'s re-hide logic works again and it clears

So: Target on Blizzard + ToT on EllesmereUI (or Hidden) looks correct out of combat, breaks the second you pull, self-corrects the second combat ends. Same for Focus/Focus Target. Not a logic gap on our end - it's combat lockdown doing exactly what it's designed to do, just somewhere visible. No real workaround short of reimplementing ToT as its own secure frame, which isn't worth it for what it'd buy.

(Related, same cause: ToT/FocusTarget are children of the parent frame, not standalone, that's why "Blizzard" only shows up as an option when the parent's also on Blizzard.)

So, if Target's source is Blizzard, keep ToT on Blizzard too (same logic for Focus/FocusTarget) - mixing a Blizzard parent with an EllesmereUI or Hidden child is the one combination we'd flag as not recommended

PR #633, opened 2026-07-08
Player debuff colored border

voidlol:
Add colored border (based on debuff dispel type) to player's unitframe debuff frame


EllesmereGaming:
have to integrate this into the new ptr merged version which is in a new path so closing this and adding it manually, thank you!

PR #647, opened 2026-07-09
Feature: Add chat tab font size option

JuJuFX-dev:
### Summary
Adds a new option letting users adjust the font size of the chat tab labels (e.g. "General", "Trade", ...) independently from the actual chat text size.

### Changes

**`EllesmereUIChat.lua`**
- `UpdateTabStyle()`: tab font size is now read from `Cfg("tabFontSize")` instead of being hardcoded to `11` (the DB default was already `10`, but was never actually used anywhere).
- New function `ECHAT.ApplyTabFontSize()`: updates the font size on all currently skinned tabs live, no reload required.

**`EUI_Chat_Options.lua`**
- New **"Tab Font Size"** slider (range 8–20) in the cog popup next to the "Font" dropdown (alongside "Outline Mode").
- `RefreshAll()` now also calls `ApplyTabFontSize()`, so the size is correctly re-applied on e.g. profile switch.

### Behavior
- Change applies instantly, no `/reload` needed.
- Only affects the tab label text, not the chat message text size (which is still controlled via Blizzard's right-click "Font Size" menu).

### DIFF
--- a/EllesmereUIChat/EUI_Chat_Options.lua
+++ b/EllesmereUIChat/EUI_Chat_Options.lua
@@ -27,6 +27,7 @@ initFrame:SetScript("OnEvent", function(self)
     local function RefreshAll()
         if ECHAT.ApplyBackground  then ECHAT.ApplyBackground()  end
         if ECHAT.ApplyFonts       then ECHAT.ApplyFonts()       end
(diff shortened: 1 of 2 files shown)


JuJuFX-dev:
already got implemented with another pr

PR #652, opened 2026-07-09
Render Tracking Bar text above overlapping buff icon displays

Filpet96:
## Summary

Fixes the [eui-bug-reports thread](https://discord.com/channels/585577383847788554/1518302204509950014) "Tracking Bars text has lower strata than the border" (reported by Homing, still reproducing on v8.4.1).

The tracking bar and the buff-icon cooldown displays both sat on `MEDIUM` frame strata. Because a buff icon's border frame is created at `icon +13` and its cooldown swipe rides on the icon frame, an overlapping buff-icon display drew its border and darkened cooldown swipe on top of the tracking bar's text **and** bar body.

This pins the whole tracking bar `wrapFrame` to `HIGH` strata, so the entire bar (fill, background, border, and text) renders above the `MEDIUM` buff-icon displays and their borders/cooldown swipes wherever the two overlap. The text overlay inherits the wrap's strata (keeping its `+6` level so it stays above the bar's own border and pandemic glow). The options preview builds bars with the same live constructor and continues to lift them into the popout's `FULLSCREEN_DIALOG` strata.

## Changes

- `EllesmereUICdmBuffBars.lua`: tracking bar `wrapFrame` now uses `HIGH` strata; text overlay inherits it instead of pinning its own.
- `EUI_CooldownManager_Options.lua`: updated preview comment; text overlay re-level unchanged.

## Test plan

- [x] Overlap a Tracking Bar with the Buffs (buff icon) display — the bar text, fill, and border all render above the icons, their borders, and cooldown swipes (no dark swipe bleeding onto the bar).
- [x] Original repro: border 1, bar height lower than text, text size larger than bar — text stays above the border.
- [x] Options popout preview still shows the bar text above the bar/border correctly.
- [x] No regression when a tracking bar is not overlapping any icon display.


EllesmereGaming:
manually merged

PR #659, opened 2026-07-09
Fix: Detached bar strata not re-applied on live settings refresh (Target/Focus/Boss)

JuJuFX-dev:
Bug Report: https://discord.com/channels/585577383847788554/1524025329122017300

## Problem
When **Custom Bar Strata** is enabled and the detached Power Bar strata (e.g. `HIGH`) differs from the main **Frame Strata** (e.g. `MEDIUM`), the frame's border would render on top of the detached power bar for the **Target**, **Focus**, and **Boss** unit frames.

This was **not reliably fixed by `/reload`** either — for this specific `Frame Strata = Medium` / `Detached Power Bar Strata = High` combination, the border was still visible over the power bar even on a freshly reloaded UI, not just after live settings changes.

The **Player** frame was unaffected in both cases.

### Root Cause
`frame:SetFrameStrata(strata)` on the main unit frame re-stacks all descendant frames (Health, Border, Power, Castbar, etc.) down to the frame's own strata — this is already acknowledged elsewhere in the code (see the existing comment on the Castbar strata restore).

The **Player** branch of the settings-apply function already re-applies the detached Power Bar's custom strata *after* this reset:
```lua
if ppIsDetached and db.profile.enableCustomBarStratas then
    frame.Power:SetFrameStrata(db.profile.detachedPowerStrata or "HIGH")
elseif ppIsDetached then
    frame.Power:SetFrameStrata("MEDIUM")
end
```

This restore was **missing** in the **Target**, **Focus**, and **Boss** branches — the "is detached" flag (`ppIsDetached2` / `fPpIsDet` / `bPpIsDet`) was computed in each of these but never actually used to set the strata.

Crucially, `ReloadFrames()` (the function containing this logic) is not just the live-settings-refresh handler — it is also called as the final step of `InitializeFrames()`, i.e. on every `/reload` and initial login. So even though `CreatePowerBar` correctly sets the detached Power Bar to `HIGH` at creation time, `ReloadFrames()` runs immediately afterward as part of the same initialization sequence, cascades Power's strata back down to `MEDIUM`, and then — for Target/Focus/Boss — never restores it. This is why the bug persisted even right after a reload, not just on live changes.

### Changes
- Added the missing detached Power Bar strata restore to the **Target**, **Focus**, and **Boss** branches of the settings-apply logic, mirroring the existing Player-branch logic 1:1.
- (Earlier, exploratory commit) Added explicit strata re-sync for `frame.Health`, `frame.unifiedBorder`, and attached (non-detached) `frame.Power` in both strata-refresh loops (`ReloadFrames` / `InitializeFrames`) for defensive consistency — functionally redundant given the cascade behavior, but kept as a safety net against future changes to that cascade logic. No behavior change on its own.
- Pet / Target-of-Target / Focus-Target frames were checked and confirmed not to support detached power bars, so no fix needed there.

### Testing
- Set **Frame Strata = Medium**, **Detached Power Bar Strata = High**, `Custom Bar Strata` enabled.
- Verified on a fresh `/reload` that the border no longer renders over the detached Power Bar for Target, Focus, and a Boss frame.
- Toggled strata settings live in the options menu (no reload) and confirmed the fix applies immediately.

### Files changed
- `EllesmereUIUnitFrames.lua`

### DIFF
--- a/EllesmereUIUnitFrames/EllesmereUIUnitFrames.lua
+++ b/EllesmereUIUnitFrames/EllesmereUIUnitFrames.lua
@@ -1266,7 +1266,7 @@ local function ApplyDarkTheme(health)
         -- Check for custom fill/bg colors on this unit
         local unitKey = health._euiUnitKey
         local unitSettings = unitKey and db.profile[unitKey]
-        -- Pet class coloring is opt-in. Pets never match oUF's colorClass (that
+                -- Pet class coloring is opt-in. Pets never match oUF's colorClass (that
         -- path requires a player unit), so the pet frame uses colorClassPet, gated
         -- on the pet page's Class Colored Fill toggle.
         if unitKey == "pet" and unitSettings and unitSettings.healthClassColored then
@@ -6785,6 +6785,7 @@ local function StylePetFrame(frame, unit)
     bg:SetColorTexture(0, 0, 0, 0.5)
     health.bg = bg
 
+    health.colorClass = true
     health.colorReaction = true
     health.colorTapped = true
     health.colorDisconnected = true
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
merged another PR that did the same thing

PR #667, opened 2026-07-10
feat(damagemeters): fade-out time for icon history

ExoJdi:
Adds a Fade-Out Time slider for the spell icon history (0 = keep forever). Each icon expires individually once its cast is older than the configured time, then fades its alpha to zero over 1.5s instead of vanishing abruptly. Expiry only counts down out of combat so the history stays readable through a fight. Options-page preview placeholders no longer appear in slots vacated by faded icons. Includes locale entries.


EllesmereGaming:
Thank you this is a great idea! I'm down to merge this, but it has a performance issue that needs to be addressed first. I'd want that ticker replaced with event based player_regen_enabled instead. fast tick should only happen when an icon is actually fading, otherwise just use a c_timer.after set to the next expiring icon. As a general rule for PRing in the future for EUI, any new feature needs to be event driven unless absolutely necessary.

EllesmereGaming:
a version of this feature has been implemented and is live

PR #668, opened 2026-07-10
feat(resourcebars): full bar stack preview with real offsets

ExoJdi:
The options preview only showed the class resource pips. It now renders the whole stack the player sees in game - class resource, power bar and health bar - positioned at their actual configured offsetX/offsetY values, so spacing changes (including overlaps and negative gaps) are visible immediately. The group is shifted as one unit to stay centered in the header, and texture, colors, border, dark mode and text settings update live.

### DIFF
--- a/EllesmereUIResourceBars/EUI_ResourceBars_Options.lua
+++ b/EllesmereUIResourceBars/EUI_ResourceBars_Options.lua
@@ -210,18 +210,162 @@ initFrame:SetScript("OnEvent", function(self)
         return 5
     end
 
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
there are a lot of factors to this with click to scroll interactions, header height when offsets are large, vertical bars, etc making the header either awkward or way too large so the options section is shrunk, you could make it scrollable but this is a lot of factors to consider and work to do for something that doesn't feel super necessary compared to other previews. if you want to try and account for all of that you're welcome to but i am closing this for now due to all the additional complexities it would introduce

PR #670, opened 2026-07-10
feat(nameplates): raid marker opacity and friendly plate markers

ExoJdi:
Adds opacity sliders for both the name raid marker and the slot raid marker, and shows the name marker on friendly plates: in health-bar mode it attaches to the custom friendly plate, in name-only mode a standalone marker anchors to the visible name text and stays in sync via RAID_TARGET_UPDATE. Shared enable/size/opacity settings drive all cases. Includes locale entries.


EllesmereGaming:

PR #672, opened 2026-07-10
feat(chat): abbreviate channel names option

ExoJdi:
Adds an Abbreviate Channel Names toggle that shortens channel prefixes like [2. Trade] to [2] in the chat frame, keeping the clickable channel link intact. Includes locale entries.


EllesmereGaming:

PR #673, opened 2026-07-10
feat(chat): timestamp color source and bracket style

ExoJdi:
Adds two color-source swatches next to the Timestamps dropdown: chat color (timestamp inherits the message color, default) and a custom fixed color that opens the color picker. The cog menu gets a Brackets dropdown (none / [15:27] / (15:27)) with example labels rendered from the selected time format. Includes locale entries for the new strings.


EllesmereGaming:

PR #675, opened 2026-07-10
Feat/chat remove name brackets

ExoJdi:


EllesmereGaming:

PR #676, opened 2026-07-10
Add minimap rotation and facing cone options

aizuon:
## Summary

- add a toggle that rotates circular minimap textures with the player
- add an independent blue facing cone for circular and square minimaps
- place both controls at the end of the Minimap Display section

## Why

These options make the player's facing direction easier to read at a glance. Rotation uses WoW's native minimap behavior for circular maps, while the facing cone also supports square maps without exposing the native rotating renderer's empty corners.

## Demo
https://github.com/user-attachments/assets/f2abecfe-65b4-4a25-9396-eaeaa6a8e810


EllesmereGaming:

PR #681, opened 2026-07-11
Action bar text color bug

dfrisone:
## Summary
Fixes custom action-bar hotkey text color (e.g. yellow) reverting to Blizzard's default white on several triggers: target change, combat enter/exit, dragging a spell onto a slot, and usable-state changes. The custom color only "returned" whenever something else happened to re-trigger the addon's own font-application pass (e.g. combat ending), making it feel completely unpredictable.

## Root Cause
Blizzard's `ActionBarActionButtonMixin` resets `HotKey`'s text color on native refreshes (`UpdateHotkeys`, `Update`, `UpdateAction`, `UpdateUsable`), and this addon only ever applied the custom color once. It had no mechanism to re-assert the color after Blizzard's own code touched it again.

## Changes

**`EllesmereUIActionBars/EllesmereUIActionBars.lua`**

Layered fix, from most to least targeted:
- `hooksecurefunc` on all four `ActionBarActionButtonMixin` refresh methods (`UpdateHotkeys`, `Update`, `UpdateAction`, `UpdateUsable`) to re-apply the configured hotkey color immediately after Blizzard's own update — covers target change, combat transitions, and usable-state flips with no visible flash.
- A debounced `ACTIONBAR_SLOT_CHANGED` handler for the drag-between-slots case, combined into the existing registration rather than added as a second one (`EAB:RegisterEvent` only keeps one handler per event name — registering twice would have silently broken the pre-existing "Always Show Buttons" visibility feature).
- A `self:ApplyFonts()` call added to the existing `PLAYER_TARGET_CHANGED` handler for the usable-state-on-target-change case specifically.
- A 0.2s polling backstop (`EnforceHotkeyColors`) that compares each button's current hotkey color against the configured value and corrects any mismatch — a guaranteed catch-all for cases not covered by the above (some buttons were observed staying persistently white out of combat with no discrete trigger any hook caught).

## Bug Fixes
- Fixed hotkey text color reverting to white on target change, combat enter/exit, drag-and-drop between action slots, and ability usability changes.

## Code Review Notes
Ran a full multi-angle review after the fix stabilized. Two confirmed, low-severity items worth a follow-up pass (not blocking): a missing `IsForbidden()` guard before touching Blizzard's `HotKey` fontstring (matches an existing convention elsewhere in this file), and two call sites using the heavier `self:ApplyFonts()` where a lighter color-only reapply would suffice. Neither affects correctness of the actual fix.

## Verification
- Confirmed color persists through target changes, combat entry/exit, drag-and-drop between action bar buttons (including empty-slot drag transitions), and ability usability changes.
- Confirmed performance impact is negligible (ticker cost estimated and compared against existing tickers of similar cadence already in this codebase).
actionbarfix1
actionbarfix2

These pictures show the action bar text no longer default to a white state on use, on drag, or on target.


EllesmereGaming:
this would reintroduce a local issue that was resolved in regards to mouseover performance so can't merge but i am going to address this with a different fix that should accomplish the same thing

PR #682, opened 2026-07-11
feat(damagemeter): register windows with UnlockMode

bmfurtado:
Damage meter windows previously moved via a private drag overlay, bypassing the core UnlockMode mover system that every other module uses. Register each window as an unlock element (EDM_Win1..N) so UnlockMode's cog panel, anchor-to-element, size matching, and arrow-key nudge all work on them.

- Add ns.RegisterDMUnlock(): registers/refreshes EDM_WinN elements on login, window add/remove, and profile swap.
- Add ns.ApplyWinPosition() to apply saved positions in both the legacy TOPLEFT-drag format and UnlockMode's point/relPoint format, so existing saved layouts load unchanged.
- Remove the private unlock overlay; unlock-mode dragging, resizing, and anchoring now go through the core mover system.
- Re-key anchor/size-match links on window deletion via ShiftIndexedAnchorKeys, fixing a stale-winIdx bug in the DB compact step along the way.
- Manual (outside-unlock) dragging and resizing behavior is unchanged, and now also clears any unlock-mode anchor on the window and reapplies anchored children.
- Per-window lock button (wdb.locked) still blocks outside-unlock drag/resize as before; UnlockMode can still move a locked window.

### DIFF
--- a/EllesmereUIDamageMeters/EllesmereUIDamageMeters.lua
+++ b/EllesmereUIDamageMeters/EllesmereUIDamageMeters.lua
@@ -452,6 +452,27 @@ local function WinDB(idx)
     return cfg.windows[idx]
 end
 
+-- Applies a window's saved position to its frame. Two stored formats:
+--   unlock format: { point, relPoint, x, y } -- relative to UIParent, written
+--                  by unlock mode's Save & Exit (CENTER/CENTER canonical form)
+--   legacy format: { x = left, y = top } -- TOPLEFT offset from UIParent's
+--                  BOTTOMLEFT, written by header drags outside unlock mode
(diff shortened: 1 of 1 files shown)


bmfurtado:
I couldn't find any guidelines around contributions so let me know if I missed something. 

I didn't open an issue but my rationale for this is that I play on the same machine on monitors with very different aspect ratios. On my 32:9 monitor I moved all the UI elements closer to the middle of the screen roughly where they'd be on a 16:9 monitor. Since Unlock Mode is center-anchored, this works perfectly for most UI elements but the damage meter windows being top-left aligned meant I had to move them every time I switched.

PR #688, opened 2026-07-11
feat(cdm): fill bar by stacks

uNBEx:
Added an ability to fill the CDM bar by the number of buff/debuff stacks instead of by duration. Especially useful for frost mage's shatter/freezing stacks.

This was highly requested in [discord](https://discord.com/channels/585577383847788554/1483614440707457055)

This feature was vibecoded with some educated guidance, I have checked the code and the feature ingame by hand - everything seems to be working fine. Feel free to edit this to your liking.


uNBEx:
removed tostring as I think it might error with 12.1.0 API changes

uNBEx:
Unfortunately this is broken on PTR, I'll continue to work on it and will reopen once I get a working solution.

PR #689, opened 2026-07-11
fix(blizzard-skin): avoid secret-boolean errors on Group Finder checkboxes

Filpet96:
## Summary
- Fixes Lua errors thrown by the Group Finder reskin under LFG taint (12.0 secret values). With `/euidev` on, creating/browsing a group raised `attempt to perform boolean test on a secret boolean value` and `bad argument #1 to 'SetShown'` from the custom checkbox skin.
- `CheckButton:GetChecked()` can now return a *secret boolean*; boolean tests on it and `SetShown(secret)` are disallowed on tainted paths.
- The custom checkmark is now driven with `Region:SetAlphaFromBoolean` (the sanctioned secret-safe API). Remaining checked-state reads (`ResolveRailSelection`, `SaveAppRoles`, and the hover wash) are guarded with `issecretvalue`, degrading gracefully instead of erroring.
- Same fix applied to the identical checkbox helper in `EllesmereUIBlizzardSkin_WindowPacks.lua`.

## Notes
- Behavior is unchanged outside restricted content. The only difference under LFG taint: the faint hover-wash won't light up purely from the checked state (still lights on hover); the green tick still reflects checked state correctly.
- No localized strings changed, so `Locales/_keys.txt` is untouched.

## Test plan
- [x] With `/euidev` on, open Group Finder and create a group — no `EllesmereUIBlizzardSkin_GroupFinder.lua` errors.
- [x] Browse Premade Groups and apply — no secret-boolean / SetShown errors.
- [x] Category-rail checkboxes and the sign-up role checkboxes render correctly (tick + hover) outside instances.

### DIFF
--- a/EllesmereUIBlizzardSkin/EllesmereUIBlizzardSkin_GroupFinder.lua
+++ b/EllesmereUIBlizzardSkin/EllesmereUIBlizzardSkin_GroupFinder.lua
@@ -344,9 +344,9 @@ local function SkinCheckbox(cb)
     wash:Hide()
     local hovering = false
     local function updState()
-        local checked = cb:GetChecked() and true or false
-        tick:SetShown(checked)
-        wash:SetShown(hovering or checked)
+        local checked = cb:GetChecked() -- may be a secret boolean under LFG taint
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
oh i have already fixed this locally, thank you though D:

PR #691, opened 2026-07-12
Party Buff Stack Show Fix

dfrisone:
## Fix: Buff Manager "Simple Setup" never showed stack counts

**Bug report:** Reported by @DreadMesh (Party Frames, v8.4.1) — Simple Setup in Buff Manager doesn't show stacks, and there's no option to enable/adjust stack text.

### Root cause
The Buff Manager has two independent buff-rendering systems: **Custom Buff Display** (per-indicator manual setup) and **Simple Setup** (auto-tracks the active spec's full buff whitelist in a grid). Custom Buff Display already had full stack-count support (its own toggle, color, size, and offset controls). Simple Setup's icon pool, however, was built with only a texture, cooldown swipe, and border — it never had a stack-count text element at all, and `BM_UpdateSimpleGrid` never called the stack-count API. There was nothing to enable because the feature didn't exist in that code path.

### Fix
- Added the missing stack-count `FontString` to the Simple Setup icon pool (mirrors the Custom Buff Display pool's existing `_count` text).
- Wired up `C_UnitAuras.GetAuraApplicationDisplayCount` (secret-value-safe) in `BM_UpdateSimpleGrid` to populate it.
- Added the same stack-count rendering to the Simple Setup live preview in the options panel.
- Added a new **Show Stacks** row to the Simple Setup options (toggle + color swatch + size/offset cog), matching the existing Show Duration Text row's layout.
- Added matching defaults (`showStacks = true`, `stacksTextColor`, `stacksTextSize`, `stacksOffsetX`, `stacksOffsetY`) so existing profiles get sensible values automatically.

### Testing
- Verified the new **Show Stacks** row appears in `/eui` → Raid Frames → Buff Manager → Simple Setup, with a live "3" preview on the sample icon.
- Verified real stacking buffs tracked via Simple Setup now show their stack count on raid/party frames.
- Verified toggling off hides the count, and the color/size/offset cog updates live.
-During code review found one issue with toggle not dimming fixed.


afrisone:
Slow down, bro!

EllesmereGaming:
please don't LOL

PR #719, opened 2026-07-13
Feature: Add Player Aura Frames (Buffs/Debuffs) to UnlockMode

JuJuFX-dev:
## What does this PR do?

**See https://github.com/EllesmereGaming/EllesmereUI/pull/729 for implementation as hook variant with minimal changes to EUI_UnlockMode.lua**

Makes the Player Aura frames (Buffs/Debuffs) movable exclusively through EllesmereUI's UnlockMode instead of Blizzard's Edit Mode. `PlayerBuffs`/`PlayerDebuffs` are now registered as ordinary unlock elements with their own position DB (save/load/clear/apply), and a post-hook on `ApplySystemAnchor` re-applies the custom position after Blizzard's own layout refresh so it no longer gets silently reset. Blizzard's Edit Mode drag is locked for both frames using the same pattern as `LockCooldownViewerFrames` for CDM, so there is exactly one way to move them going forward.

Also fixes the Buffs frame's right-edge padding: it now reserves space for the Collapse/Expand arrow only when the `collapseExpandBuffs` CVar indicates the arrow is actually shown, instead of always reserving it.

## How was it tested?
Move Buffs and Debuffs in unlock mode, anchor them, move them with keyboard (1px).
Try to move buffs and debuffs in Blizz Edit Mode -> see chat message.
Enable CollapseExpandBuffs -> Reload -> Check unlock mode and move the buffs. Same for disable.

## Screenshots
grafik

## Checklist
**_Default **ON**_**

**_N/A: always active; cost is one hooksecurefunc + one SetScript-nil per frame at login_**

**_hooksecurefunc fires only when Blizzard's own ApplySystemAnchor runs; one deferred C_Timer.After(0) re-apply per call, not a repeating timer_**

**_Same established exception as LockCooldownViewerFrames: SetScript(nil) on frame.Selection is required to reliably block Blizzard's own drag delegation; a weak-table side-store wouldn't be visible to Blizzard's OnDragStart chain_**

- [x] Tested in-game, works on live retail; **PTR NOT TESTED**


JuJuFX-dev:
will be reworked for 12.1

PR #721, opened 2026-07-14
feat(raidframes): add in-combat icon for party and raid members

Filpet96:
## Request

- Connected to the Discord feature request: https://discord.com/channels/585577383847788554/1525856569173086260

## Summary

- Adds an opt-in **Combat Icon** on party and raid frames that marks members currently affecting combat — handy for spotting who is still pulling during a Mythic+ skip.
- Configurable position (9-point anchor + None to disable), size, style (Standard, Class Theme, and the combat0-5 sprite styles), custom/class coloring, and X/Y offsets.
- Full options integration: options row with inline cog (style, class-colored toggle, offsets), color swatch, and live preview.
- Respects party/raid indicator sync via the party key section map.
- `UNIT_FLAGS` is registered dynamically only while the feature is enabled (per-unit for raid/party/player and Extra Frames), **so no combat-tracking code runs when the option is off.**

## Screenshots
image
image

## Test plan

- [x] Enable the combat icon on raid frames; confirm it appears only on members in combat and clears when they leave combat.
- [x] Repeat for party frames (with and without party/raid sync).
- [x] Verify each style renders (Standard, Class Theme, combat0-5) and custom vs class coloring.
- [x] Verify position dropdown (including None) and size/offset controls, and that the options preview matches.
- [x] Confirm no Lua errors on entering/leaving combat, and that disabling the option unregisters UNIT_FLAGS.

### DIFF
--- a/EllesmereUIRaidFrames/EUI_RaidFrames_Options.lua
+++ b/EllesmereUIRaidFrames/EUI_RaidFrames_Options.lua
@@ -2800,6 +2800,111 @@ initFrame:SetScript("OnEvent", function(self)
             cogBtn:SetScript("OnClick", function(self) cogShow(self) end)
         end
 
+        -- Row: Combat Icon Position (includes "None" to disable) | Combat Icon Size.
+        -- Shows on members currently in combat -- e.g. to spot who's still pulling
+        -- during a Mythic+ skip. Style/color/offset live in the inline cog.
+        local combatPositionValues = {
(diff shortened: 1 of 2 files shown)


Filpet96:
@EllesmereGaming This is tested in multiple M+ tonight and it works very good, checked replay and it detects combat instantly.. very nice feature for skips in m+ etc.. to see if someone got in combat

PR #756, opened 2026-07-15
fix(visibility): druid flight form now triggers dragonriding hide

danvernon:
## Summary
- `IsAirborneSkyriding()` used `IsMounted()` which returns false for druid flight form — switched to `IsPlayerMountedLike()` which covers druid mount-like shapeshift forms (Travel Form, Flight Form, Aquatic Form)
- Removed the `IsFlying()` requirement so bars hide as soon as the skyriding mount/form is active, not only after takeoff (addresses the "bars should hide when dragonriding bar pops up" feedback)
- Updated CDM's `UPDATE_SHAPESHIFT_FORM` handler to also re-evaluate visibility when any bar uses dragonriding visibility modes, not just `visHideMounted`

## Test plan
- [ ] Set a CDM or resource bar to "When Not Dragonriding" visibility
- [ ] As a druid, shift into flight form — verify bars hide
- [ ] Land and shift out — verify bars reappear
- [ ] Mount a regular skyriding mount — verify bars hide before takeoff (when dragonriding bar appears)
- [ ] Non-druid: mount a skyriding mount — verify bars hide
- [ ] Verify bars still show correctly in combat druid forms (Cat/Bear/Moonkin)

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICooldownManager.lua
+++ b/EllesmereUICooldownManager/EllesmereUICooldownManager.lua
@@ -8347,17 +8347,21 @@ eventFrame:SetScript("OnEvent", function(_, event, unit, updateInfo, arg3)
         return
     end
     if event == "UPDATE_SHAPESHIFT_FORM" then
-        -- Bail fast if no bar actually uses visHideMounted: druids shift
-        -- constantly in combat (Bear/Cat) and we don't want to re-run the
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
this is actually an intentional design decision to only have the skyriding ones show/hide when in the air vs when landed

PR #769, opened 2026-07-16
fix(cdm): lower tracked buff bar strata from HIGH to MEDIUM

danvernon:
## Summary
- v8.4.3 elevated tracked buff bar wrapper from `MEDIUM` to `HIGH` strata to render above buff-icon borders/swipes
- `HIGH` strata renders above the dragonriding HUD and resource/power text
- Changed to `MEDIUM` strata with frame level 100, which clears the icon display stack (levels 5–27) without overlapping unrelated UI

## Test plan
- [ ] Verify tracked buff bars still render above buff-icon borders when overlapping
- [ ] Dragonriding with a tracked bar active — bar should not cover the HUD
- [ ] Confirm bar does not clip over resource/power text


EllesmereGaming:
ended up adding customizable strata instead, but also changed default to medium 100. ty!

PR #770, opened 2026-07-16
Blizz UI Enhanced: Merchant show as list

Snsei987:
## What does this PR do?

Add an option to display the merchant UI as a list. This is OFF by default and option requires a reload to be fully applied. Not reloading when asked will revert to the current value as it breaks the UI immediately when the user opens merchant UI. It also fully supports the current "Show Item Level" option. The click zone (to buy, preview etc...) is the whole row with this option enabled.

The code leverages a lot of the default UI. For example, the buttons are using the `MerchantItemTemplate` template so we don't have to implement the clicks handlers and we are modifying the children it brings by default instead of creating our own elements.
Source: https://github.com/Gethe/wow-ui-source/blob/live/Interface/AddOns/Blizzard_UIPanels_Game/Mainline/MerchantFrame.lua

Frames are created when needed then cached and modified as needed when the UI updates (`MerchantFrame_Update`).

Supports stack count (bottom right, same as default), item quantity left (top left, same as default). It also supports overlay Blizzard might add, such as azerite armors and quest starter items.

image

## How was it tested?

Both 12.0.7 and 12.1.0 were tested. On multiple merchants: repair mammoth, ritual site / void assaults, heirlooms, azerite armor...
Tested in M+ + combat state only in 12.0.7.

## Screenshots

image
image
image
image
image
image

(Show Item Level option)
image

(Forced quest starter item flag to test behavior)
image

## Checklist


Snsei987:
Still wondering if we should allow to change the height of the row (just need to change `rowHeight`) with a slider or dropdown with different mode (standard, small...) or maybe just decrease the default value. What's your opinion ?

`rowHeight = 24`
image

Other possible improvement: change the background to also be red if the icon is red.

EllesmereGaming:
my opinion is this makes me pretty nervous about taint D: this is extremely aggressive customization to blizzard frames. how thoroughly was this tested for taint? can you send me your EUI version on discord so I can have it more aggressively tested? I have fairly hard rules against writing directly to blizzard frame fields and this kinda blows all of that out instantly so i'd want to be very thorough before i brought this into the codebase. taint can be invisible and not propagate until something else happens to trigger it so sometimes its not always obvious you've introduced taint when building something. doing an actual taint log assessment would go a long way here to confirm nothing is in any danger, would have to be tested under a few different environments

EllesmereGaming:
it does look sick though lol i like the concept. and row height def seems like we should make it customizable

Snsei987:
I'll update it with the latest released version (8.4.8) and send the `EllesmereUIBlizzardSkin` folder

Snsei987:
Added the option to change the row height as discussed. Description is updated to reflect this change. When the merchant frame is open and you change the row height option, it will be updated instantly.

PR #774, opened 2026-07-16
fix(options): clamp config panel to screen

danvernon:
## Summary
- Config panel could be dragged off-screen with no recovery other than relogging
- Adds `SetClampedToScreen(true)` on the main frame

## Test plan
- [ ] Drag the config panel toward all screen edges — should stop at the boundary
- [ ] Multi-monitor setup — panel stays within the game window

### DIFF
--- a/EllesmereUI.lua
+++ b/EllesmereUI.lua
@@ -6321,8 +6321,7 @@ local function CreateMainFrame()
     clickArea:EnableMouse(true)
     clickArea:SetMovable(true)
     clickArea:RegisterForDrag("LeftButton")
-    -- No SetClampedToScreen -- the whole window (bg + content) moves as one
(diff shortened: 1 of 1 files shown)


Snsei987:
If I can voice my opinion on that, I'm against it. The config frame is quite large and sometimes it's very important to have like 50% of the frame outside the screen when tinkering your CDM config for example. I'm probably not the only one.

While there is a problem when it's not forced to be fully on the screen, I would prefer a solution where I must have at least X% of the config in the view, or having a slash command to reset the position or maybe force the default position whenever it's opened.

EllesmereGaming:
yeah this one is actually intentional, you can always /reload if somehow you get it all the way offscreen

PR #785, opened 2026-07-17
(Raid frames - Buff Manager) Add on option to show a square when an aura is missing on raid frame

Xelia-wow:
## What does this PR do?

Adds a "Show When" option to the **Square** buff indicator in Raid Frames > Buff Manager (Core section), letting you choose between:
- **When Present** (existing behavior, unchanged) — the square shows while the tracked buff is on the unit.
- **When Missing** (new) — the square shows instead while the tracked buff is **absent**, so you can spot an unapplied buff at a glance.

When "When Missing" is selected, the following fields are greyed out since there's no aura instance to drive them: Duration Swipe, Duration Text, Show Stacks, Max Duration, and the whole Threshold section. Only the square's color/border/position/opacity still apply.

This only applies to the Square indicator type (Icon and Bar are unaffected). On the 12.1 PTR client, "When Missing" is inert (the new aura-container engine only ever builds slots for present auras) — the dropdown stays usable but is marked with the existing "removed/inert on 12.1" overlay, matching the precedent already used for Max Duration.

## How was it tested?

Tested in-game on live retail (12.0.7): configured a Square indicator tracking multiple spells, toggled "Show When" between Present/Missing, confirmed:
- Duration Swipe/Text, Show Stacks, Max Duration, and Threshold grey out and re-enable immediately (no reopen needed) when switching modes.
- Each tracked spell's square shows/hides independently based on that spell's own aura presence, with correct color/border/position.
- Each tracked spell's square hides/shows on unit death/res.
- No Lua errors during normal use.

### When Present
image

### When missing:
image
image

### Demo on raid frame
image
image

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) — `showWhen` defaults to `"present"`, matching current behavior exactly; nothing changes for existing profiles.
- [x] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built — purely a new branch in the existing UNIT_AURA-driven update path, only taken when `showWhen == "missing"` is explicitly set.
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations) — reuses the existing pooled indicator frames and the existing UNIT_AURA-driven rescan; no `OnUpdate`/timers added.
- [x] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames — N/A, this only touches EllesmereUI's own pooled indicator frames, not Blizzard frames.
- [x] Tested in-game, works on live retail; no load errors on the 12.1 PTR client — verified on retail; "When Missing" is inert by design on 12.1 (aura-container engine limitation, see description) with no error path (falls back silently, same as Max Duration).


Xelia-wow:
Using it in raid I realized that the square show also if the player is dead. I will fix it later tonight.

EllesmereGaming:

PR #815, opened 2026-07-18
fixed issues in French Localstrings

DlargeX:
## Checklist


EllesmereGaming:
already fixed ty!

PR #830, opened 2026-07-18
Actionbars: Add new section "BAR BACKGROUND" and add more bar background options

Ricoder92:
### What does this PR do?

**Adds expanded per-action-bar background customization:**

- Dedicated Bar Background section below Icon Effects
- Optional background color and opacity
- Configurable spacing around the icons
- Independent X/Y size multipliers with selectable growth directions
- Border style, size, color, and texture-specific offset controls
- Live configuration preview
- “Apply to all” and selective multi-apply support for:Enable
  - Spacing
  - Border Style
  - Border Size and Color

The feature remains disabled by default, preserving existing behavior unless explicitly enabled.

### How was it tested?

Verified modified files with git diff --check
Confirmed the addon stays within Lua 5.1 local-variable limits
Reviewed disabled-state behavior to ensure runtime background frames are not created while the feature is disabled
In-game testing on live retail and the 12.1 PTR is still required

### Screenshots

Before:
image

After: (Example with shadow border)
image

New Settings "BAR BACKGROUND" 
image

### Checklist

- [x] New settings default OFF (no behavior change without opt-in)
- [x] No writes onto Blizzard-owned frames; background elements are created on EllesmereUI-owned bar frames


Ricoder92:
tested on PTR, no errors :)

PR #842, opened 2026-07-19
fix(cdm): stop saved icon order snapping back to Blizzard order on reload

dfrisone:
## Problem

Reported by @woaw (8.4.9, Survival Hunter): Raptor Strike and Kill Command swap places in the CDM after every `/reload`, regardless of how often they're reordered in EUI.

## Root cause (confirmed with a live event dump)

Blizzard's spell data links **SV Kill Command (259489) to Steady Shot (56641) as its base spell** -- a different ability. The options normalize pass therefore stores Kill Command's slot as `56641` (a live `/euiorder` dump literally shows "56641:Steady Shot" on a Survival cooldowns bar).

The login reseed pass compared live icons against stored ids with an **exact-match** seen set. Live `259489` never matched stored `56641`, so every reload re-inserted Kill Command at its **Blizzard layout position** -- in front of Raptor Strike for layouts where Blizzard places it leftmost. The next options-panel open then deduped in favor of that copy, permanently deleting the user's saved slot. Re-dragging could never stick. A second latent bug rode along: the insert cursor was looked up **by value**, which fails the same way on a variant mismatch and dumps inserts at slot 1.

## Fix

In both materializer passes (the automatic login reseed in `ReseedAssignedSpellsFromLiveIcons` and the options append in `EnsureAssignedSpells`):

- **Presence is variant-aware** via `ns.FindVariantIndexInList`, so any stored variant form (base or override) counts as already-slotted. FindVariantIndexInList handles negative trinket/item markers by exact scan internally, so no separate fallback is needed.
- **The insertion cursor is a position index**, immune to id-form mismatches.
- **The reseed now stores the base form**, matching the options normalize pass, so the two writers can no longer persist divergent forms of the same spell (previously exports shipped whichever form the last-run pass wrote).

All guards (ghost, other-bar owner, hosted-buff/overflow/placeholder skips, claimed-elsewhere) are unchanged, and behavior for non-variant spells is identical.

Also extends the permanent `/euiorder` diagnostic with per-icon identity probes (sid/base/override/canonical/linked/layoutIndex/sortOrder) and a dump of the bar's cached order map -- this instrumentation is what pinned the root cause from a single paste.

## Review

High-effort multi-angle review before submitting: a removed-behavior audit traced every invariant of the old code as re-established (including out-of-order live-icon sequences and negative-marker handling), and dead fallback loops flagged by the review were removed. Known pre-existing latents deliberately left out of scope (candidates for follow-ups): the `NormalizeToBase` writer itself has no round-trip guard for cross-spell base links, and a few manual-add/preset-overlap guards elsewhere still use exact id matching.

## Testing

Verified in-game on a Survival Hunter: with the fix loaded, stored order and rendered order agree (`/euiorder` shows the bridged order map ranking Kill Command at its stored Steady-Shot slot), reordering Raptor Strike / Kill Command persists across `/reload` in both directions, and non-variant spells reorder exactly as before. Users bitten by the old bug need one final re-drag (their saved order was already destroyed); it sticks from then on.

### DIFF
--- a/EllesmereUICooldownManager/EUI_CooldownManager_Options.lua
+++ b/EllesmereUICooldownManager/EUI_CooldownManager_Options.lua
@@ -6331,10 +6331,6 @@ initFrame:SetScript("OnEvent", function(self)
                           and ns.cdmBarIcons and ns.cdmBarIcons[barKeyE]
         if liveIcons then
             if not sd.assignedSpells then sd.assignedSpells = {} end
-            local seen = {}
-            for _, existing in ipairs(sd.assignedSpells) do
-                seen[existing] = true
-            end
             local removed = sd.removedSpells
             -- Never materialize a spell that's currently HIDDEN (in the ghost bar)
             -- onto a visible bar -- that recreates a both-state and the spell would
@@ -6378,7 +6374,14 @@ initFrame:SetScript("OnEvent", function(self)
             -- position instead of being appended at the very end (which piled
             -- talent-swap spells after the trinket/racial slots and never survived
             -- a /reload cleanly).
-            local insertAfterSid = nil
+            -- Variant-aware presence + POSITION cursor (mirrors the reseed pass in
+            -- EllesmereUICooldownManager.lua): the old exact-match seen set missed a
+            -- stored entry when the live icon resolved to a different variant form,
+            -- re-inserting a duplicate at Blizzard's position -- and the normalize
+            -- pass above then deduped keeping that copy, deleting the user's saved
+            -- slot (the Raptor Strike / Kill Command order swap, 8.4.9). The
+            -- by-value cursor lookup failed identically and dumped inserts at 1.
+            local insertPos = nil
             for _, icon in ipairs(liveIcons) do
                 -- Resolve the live icon's DISPLAYED spell the same way the picker
                 -- does (canonical = GetSpellID-first, with the active-frame cache),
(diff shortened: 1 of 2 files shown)


dfrisone:
Second report of the same family landed (@Doppler, 8.5.1): on a bar mixing CD/utility spells with hosted buffs (Vengeful Retreat, Moment of Craving, Voidray, Soul Fragments, Collapsing Star, Shift), Shift always snapped back to right after Voidray -- jumping its three buffs -- mid-combat and permanently after /reload.

Same mechanism, two halves:
1. **The presence miss** (why Shift got re-inserted at all): id-form drift, the same class as the Kill Command/Steady Shot case -- Shift being a form-transform ability makes its live id change in combat, which is why it "happens at some point during combat". Covered by this PR's variant-aware matching.
2. **The placement** (why it landed right after Voidray specifically): the reconcile passes skip hosted-buff frames entirely (their membership is a negative marker, not a spell id), so the insert cursor never advanced past the buffs. A missed spell was re-inserted at cursor+1 = immediately after the last matched CD spell. New commit ce5f70f1 closes this: skipped hosted-buff icons now advance the cursor to their marker's index (variant-tolerant, forward-only, overflow-diverted frames excluded).

With both halves, mixed-bar orders should hold across combat and reload.

dfrisone:
Third independent report of the same mechanism (Skolodac, Balance Druid): Wrath jumps from 3rd to 1st on every relog/reload, but ONLY when Moonfire and Starfire aura dots (marker entries, not spells) sit before it. An identical setup on their spriest is unaffected.

This one confirms both halves of the fix cleanly:

- **Presence miss**: Wrath is a spec/base id-variant pair, so the stored form and the live icon's form differ -- the old exact-match reconcile re-inserted it as "new". The spriest bar is the natural control case: same aura-dot layout, but stable spell ids, so no miss and no reversion.
- **Placement at slot 1**: with only skipped marker entries ahead of Wrath, the old cursor was still nil at the miss, so the "no anchored predecessor -> insert at front" branch fired. That's why the dots looked causal to the reporter.

Running tally of reports this PR resolves: Raptor Strike / Kill Command (Survival, cross-spell GetBaseSpell link -- verified in-game), Shift on a mixed spell+buff bar (Devourer DH -- verified in-game across /reload), and now Wrath behind aura dots (Balance). Three specs, one mechanism. Affected users need one final re-drag after the fix (their saved order was already corrupted by the old code); it persists from then on.

PR #846, opened 2026-07-19
Unitframes: Add solid border for attched power bar

Ricoder92:
# Add Divider Borders to Attached Unit Frame Power Bars

## Summary

This change adds configurable divider borders to attached Unit Frame power bars.

When a power bar is attached to the health bar, only the shared edge is rendered:

- A power bar attached below the health bar displays a top border.
- A power bar attached above the health bar displays a bottom border.
- No left, right, or outer border edges are displayed.

Attached power bar borders always use the `Solid` texture. The existing border size slider supports values from `0` to `4`, and the existing color picker controls the divider color and opacity. A size of `0` disables the divider.

Detached power bars retain their existing full-border behavior and configurable border textures.

## Implementation

- Extended the shared Unit Frame power border update function to support both attached and detached positions.
- Reused the existing per-unit power border size, color, and opacity settings.
- Forced the `Solid` border style for attached power bars.
- Hid all border edges except the edge shared with the health bar.
- Preserved the selected full border style for detached power bars.
- Enabled the border size control whenever the power bar is visible.
- Updated the Unit Frame preview to render the attached divider consistently with live frames.
- Updated the disabled Border Style tooltip to explain that border textures are configurable only for detached power bars and that attached power bars always use `Solid`.
- Added locale keys and translations for all supported non-English locales.
- Added the new keys to the generated locale key catalog in alphabetical order.

## Behavior Matrix

| Power bar position | Border behavior |
| --- | --- |
| Below Health Bar | Solid top divider only |
| Above Health Bar | Solid bottom divider only |
| Detached Bottom | Existing configurable full border |
| Detached Top | Existing configurable full border |
| None | Border controls are disabled and no border is shown |

## Verification Checklist

- [x] Attached power bars support border sizes from `0` to `4`.
- [x] Border size `0` hides the attached divider.
- [x] A power bar below the health bar renders only its top edge.
- [x] A power bar above the health bar renders only its bottom edge.
- [x] Attached power bars always use the `Solid` texture.
- [x] Border color and opacity are applied to the attached divider.
- [x] Detached power bars retain their configurable full-border styles.
- [x] Selecting `None` hides the power bar border and disables its size control.
- [x] The options preview matches the live Unit Frame behavior.
- [x] The Border Style tooltip explains the attached and detached behavior.
- [x] Locale entries exist for `deDE`, `esES`, `esMX`, `frFR`, `itIT`, `koKR`, `ptBR`, `ruRU`, `zhCN`, and `zhTW`.
- [x] The locale key catalog contains both new tooltip keys in alphabetical order.
- [x] `git diff --check` passes.

## Suggested In-Game Testing

- [x] Test Player, Target, and Focus frames with the power bar attached below.
- [x] Test Player, Target, and Focus frames with the power bar attached above.
- [x] Verify every border size from `0` through `4`.
- [x] Verify custom border colors and opacity values.
- [x] Switch between attached and detached positions without reloading.
- [x] Confirm detached border textures remain unchanged after switching positions.
- [x] Confirm the preview updates when position, size, or color changes.
- [x] Confirm the localized tooltips on at least one non-English client.

## Screeshots

Border Size 0
image

Border Size 1
image


EllesmereGaming:
these are great additions thank you!! just a heads up i already fixed it so no need for a followup but this would've made it so any user that customized their border on a detached version and then swapped back to attached (so their slider was still at 2 for example) would login to a border on their power bar which would be an unexpected behavior change for users. I added a migration but just something to keep in mind for future PRs regarding 0 changes to existing user behavior. Thanks again!

Ricoder92:
Thanks for your feedback! In the future, I will pay more attention to the fact that the existing profiles or settings of the users are not affected or creates weird behavior. Cant wait to see it in the release! Thanks.

PR #851, opened 2026-07-19
fix(actionbars): repair Blizzard Style size tooltip and block junk width-match writes

dfrisone:
## Problem

Reported by @Zylann (8.5.1): after width-matching an action bar and undoing it under Blizzard Style Action Bars, the Icon Size slider stayed unusable with no explanation. The report asked whether it's intended and requested a tooltip -- because none ever appeared.

## Three defects on one seam

**1. The disabled slider's tooltip has NEVER rendered -- a Lua scoping bug.**
`BLIZZ_DIS_TIP` is a `local` inside `BuildMenuBagsXPPage`; the Icon Size rows live in `BuildSharedBarSettings`, where the same name resolves as a **nil global**, so `ResolveDisabledTip` silently returned nil. Confirmed with live hover instrumentation: the hit frames fired (`disabled=true`) but `resolved tip = nil`, while other tooltips rendered normally. The constant is now declared in the scope that uses it, with a comment marking the trap.

**2. Misleading tooltip precedence.**
With a match link present, the disabled slider's tooltip (had it rendered) said "Size matched to X. Unmatch in Unlock Mode to edit" even in Blizzard Style -- where unmatching cannot re-enable the slider (Blizzard Edit Mode owns sizing). That dead-end errand is exactly the reported confusion. The Blizzard Style message now always wins.

**3. Width/Height Match wrote junk settings in Blizzard Style.**
Unlock Mode offered size matching for action bars under Blizzard Style, where EUI does not control sizing: the match apply and the unmatch width-persist both wrote `buttonWidth`/`_matchExtraPixels` derived junk into the EUI-style settings -- visually inert until the user switches styles back, then their icon sizes have silently changed. Action-bar elements now declare `matchUnavailable` (a new optional `MakeUnlockElement` field): Unlock Mode refuses NEW matches with an explanatory tooltip ("Size matching is unavailable with Blizzard Style Action Bars."), while clearing an existing match stays allowed so nobody gets stuck. The elements' `setWidth`/`setHeight` also no-op under Blizzard Style, mirroring the options slider's own gate.

## Testing

Verified in-game across two rounds:
- Blizzard Style: hovering the disabled Icon Size label or track now shows the Blizzard Edit Mode message (previously nothing, reproduced and then confirmed fixed).
- Unlock Mode in Blizzard Style: Width Match on an action bar refuses with the tooltip; clearing an existing match still works.
- EUI style: slider, match and unmatch all behave exactly as before.


Zylann:
Hello

PR #863, opened 2026-07-19
fix(qol): fully serialize auto-open while a container is resolving

dfrisone:
Tester on a post-#827 build: opening 9 patron-order payout bags still stranded one greyed until relog. The single-flight cycle serialized same-slot uses but kept walking the candidate list: after an unresolved 0.5s recheck it opened the NEXT slot, and a payout container's loot window can take longer than that to appear -- a UseContainerItem issued while the previous open's loot session is still pending collides with it and strands the container (the known mechanic from the loot-window fix). The per-slot flag also cleared unconditionally at 0.5s, leaving a slow or looted-but-lingering container unprotected against same-slot re-use by a later cycle.

Keep the in-flight flag (now item GUID + timestamp) until the slot's contents actually change, halt the cycle instead of walking past an unresolved open, skip the halted cycle's timed re-scan (the resolution's own bag update or LOOT_CLOSED restarts), and refuse to start any cycle while an open is unresolved. A 30s ceiling clears a dead attempt so it cannot halt the feature; the GUID identity frees a slot when a different item (even the same itemID after a bag collapse) moves in.


dfrisone:
Superseded by #924. This branch's serialize rework (finish(skipRescan)/SlotIdentity) predated and now conflicts with the _cycleGen single-flight cycle already merged to main. Rather than re-litigate the serialize design, #924 ports only the tester-verified pacing (churn-aware delay + burst backstop) — the part this branch had that main still lacks — onto current main.

PR #868, opened 2026-07-20
Fix/databar currency empty state

Absol3m:
## The bug

Adding a Currency block to a bar produces a block that says **"Select a currency"** and cannot be configured from anywhere the player is likely to look.

The placeholder invites a click. The click was wired to `C_CurrencyInfo.OpenCurrencyPanel()` — Blizzard's currency panel, which has no way to assign a currency to a DataBar block. Clicking a currency there does nothing to the block. The currency is only selectable from a dropdown in the DataBars options, and nothing on screen said so: the unconfigured block had no tooltip at all.

So the block shipped with an empty state that leads nowhere. A player either already knows where the dropdown is, or gives up.

## The fix

Left click now branches on the block's state:

- **no currency selected** → opens the options deep-linked to that block's Currency dropdown: the right bar pre-selected, scrolled into view, and the row marked
- **currency selected** → opens the currency panel, exactly as before

The marker pulses until a currency is actually chosen, instead of the one-shot 0.75s flash the deep-link machinery normally uses — long enough to say "here", too short to survive the time it takes to read the page. It survives page rebuilds, and follows the row when reordering blocks relocates it (options sections are laid out in bar order). It clears the moment `settings.currencyId` is set.

The same marker now also fires when the block is **added** from the preview strip's "+" tile, which is the moment the intent is freshest. That path already navigated; it just aimed at the sizing row instead of the choice.

Tooltips got the other half of the gap: the unconfigured block had none, and the configured one never mentioned what a click does (the Gold block already did this — Currency didn't).

## Implementation notes

The deep link reuses what the preview strip already relies on — `EllesmereUI:NavigateToElementSettings` plus the per-build `_edbClickTargets` map — so the new plumbing is small:

- `ns.OpenBlockSettings(barId, blockId, settingKey)` — the one call that runs options → runtime, documented as such in the API header of `EllesmereUIDataBars.lua`, since that header otherwise states the traffic is one-way
- one extra click target for the picker row, keyed `block:<id>:currency` (a namespace extension of the existing `block:<id>`)
- an optional `holdWhile` predicate on `PlaySettingGlow`; every existing caller passes nothing and keeps the one-shot fade

**Blizzard's currency panel is not touched.** Hooking its rows to capture the clicked currency was the other possible route and was rejected: `EllesmereUIBlizzardSkin_WindowPacks.lua` already documents that touching those rows taints the warband currency transfer and gets the protected `RequestCurrencyFromAccountCharacter` forbidden, which is why the skin module leaves them 100% stock.

## Known scope limit

The pulse re-arms on page **rebuilds**. Reopening the options panel takes `SelectPage`'s warm path, which re-shows the cached wrapper without running the page builder, so the pulse stays dead until the next rebuild. Noted in the code with the seam that would close it (`onPageCacheRestore`, the way `_edbStripRelayout` is driven). Left out deliberately to keep the diff to the reported problem.

## Notes for translators

Adds one source string: `Open Settings`. It reaches `EllesmereUI.L` through the DataBars `ns.L` table, so it is invisible to `.tools/extract-locale-keys.sh` and `Locales/_keys.txt` is unchanged (verified — the script is a no-op on this branch).

## Testing

In-game, on a live bar:

- add a Currency block → the picker is marked immediately, and keeps pulsing
- click the placeholder on the bar → lands on the picker, marked
- reorder the block on the bar before choosing → the view follows it to its new position and it keeps pulsing
- toggle an unrelated setting on another block → the view does **not** jump
- choose a currency, then reorder → nothing pulses any more
- tooltips in both states

`luac -p` clean on all three files.

## Checklist

### DIFF
--- a/EllesmereUIDataBars/EllesmereUIDataBars.lua
+++ b/EllesmereUIDataBars/EllesmereUIDataBars.lua
@@ -101,6 +101,7 @@ local L = {
     DELVE_JOURNEY        = "Delver's Journey",
     COMPANION_LEVEL      = "Companion Level",
     SELECT_CURRENCY      = "Select a currency",
+    OPEN_SETTINGS        = "Open Settings",
     WHISPER              = "Whisper",
     WHISPER_BNET         = "Whisper BNet",
     INVITE               = "Invite",
--- a/EllesmereUIDataBars/EllesmereUIDataBars.lua
+++ b/EllesmereUIDataBars/EllesmereUIDataBars.lua
@@ -43,6 +43,15 @@
 --   ns.UpdateAllBarVisibility()
 --   ns.MakePreviewBackdrop(host, themeCfg)
 --   ns.BLOCK_TYPES / ns.BLOCK_DEFAULTS / ns.EDB_VIS_CAPS / ns.EDGE_PAD
+--
+--  The one call that runs the other way -- options file -> runtime, so a block
+--  whose empty state invites a click can send the player to the control that
+--  fills it in. Defined from the options file's PLAYER_LOGIN handler, hence
+--  the nil guard at its call site:
+--
+--   ns.OpenBlockSettings(barId, blockId, settingKey)
+--       deep-links to the "block:<blockId>:<settingKey>" click target the
+--       options page registers for that row (see _edbClickTargets).
 
 local ADDON_NAME, ns = ...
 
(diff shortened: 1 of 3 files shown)


EllesmereGaming:
love it ty, will go well with the new databars upgrade i'm rolling out in 15 minutes

PR #877, opened 2026-07-20
[feature] Glow color swatches widget + default tint using true atlas textures

lolswirl:
## What does this PR do?

Glow options now define a 3 swatch choice: "default", "custom", "class color". Custom and class color continue their standard approach, while adding the brand new "default" mode: atlas/texture glows from FlipBook apply no coloration, keeping their standard true-gold color as opposed to applying the color mask of a "fake" gold coloration that was default for (most) glows.

Since this adds a new 3 swatch choice, this also adds class color options to AuraBuff Reminders and Pandemic Glows in the cooldown manager, as opposed to only custom coloration.

Note: some places have been unchanged, as the scope was ever-increasing with these changes that were initially created to just use the default non-tinted proc glow for cooldown manager. These places include action bars -> bar animations -> custom proc glow, and cooldown manager -> tracking bars pandemic glows.  (that both should likely follow suit with the pandemic glow display in cooldown manager). Quick search through the code also has some of that in Nameplates, but as I don't have nameplates enabled, it's skipped under this PR.

## How was it tested?

Tested in 12.0.7.68453.
Used all versions of glows (AuraBuff Reminders, Cooldown Manager's CDM Bars Pandemic Glow & right click button's glow effect color, Bar Glows, etc) to verify their states utilize the appropriate color mode ("default", "custom", "class color") in previews and out on their respective UI elements

## Screenshots

Default new:
image
image

Default old:
image
image

### Cooldown Manager -> CDM Bars -> Extras
Before:
image

After:
Note, I just have a custom color of #ffffff set here, the default is unchanged from the base EUI.
image

### Cooldown Manager -> CDM Bars -> (Buffs dropdown) -> Icon Display
Before:
image

After:
image

### Cooldown Manager -> Bar Glows
Before:
image

After:
image

###  AuraBuff Reminders -> Auras, Buffs & Consumables
Before:
image

After:
image

## Checklist

The first setting I've kept unchecked, simply because this technically does change default behavior - where the old "default" was a fake-gold coloration and now the default mode actually does set it to its default coloration via the atlas/texture component of the custom glows. For glow types outside of FlipBook types, they will continue to use the default coloration of #FFC923, except for Pandemic Glow, that had a default of #FFFF00.

### DIFF
--- a/EllesmereUIAuraBuffReminders/EUI_AuraBuffReminders_Options.lua
+++ b/EllesmereUIAuraBuffReminders/EUI_AuraBuffReminders_Options.lua
@@ -895,65 +895,61 @@ initFrame:SetScript("OnEvent", function(self)
         -- Inline color swatch on Glow Type (right of row 3)
         do
             local rgn = row2._rightRegion
-            local swatch = EllesmereUI.BuildColorSwatch(rgn, rgn:GetFrameLevel()+5,
-                function()
-                    local d = DDB()
-                    local gc = d and d.glowColor or {r=1, g=0.776, b=0.376}
-                    return gc.r, gc.g, gc.b, 1
-                end,
-                function(r, g, b)
-                    local d = DDB(); if not d then return end
-                    d.glowColor = {r=r, g=g, b=b}
-                    d.glowColorMode = "custom"
-                    RefreshAll(); UpdatePreviewHeader()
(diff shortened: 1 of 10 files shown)


EllesmereGaming:
excellent work swirl

PR #880, opened 2026-07-20
Unitframes and Blizzard Auraframe: Add border options to player-, target-, bossunitframe and blizzard auras

Ricoder92:
## Description

This PR expands aura border customization for Blizzard Player Auras and Unit Frames on both Live 12.0 and PTR 12.1.

### Changes

- Added configurable **Border Style** and **Border Size** for:
  - Blizzard Player Auras
  - Defensive Buffs
  - Player Unit Frame auras
  - Target Unit Frame auras
  - Boss Unit Frame auras
- Added border popover options:
  - Offset X/Y
  - Shift X/Y
  - Show Behind
  - Behind Unit Frame
  - Border color
- Selecting **Shadow** automatically enables both **Show Behind** and **Behind Unit Frame**.
- Added an option to hide the Blizzard buff-frame expand button.
  - All buffs remain expanded when the button is hidden.
- Added profile migration for the previous numeric border-size setting.
- New Unit Frame aura borders default to **Solid** and **Thin**.
- Moved border settings to the end of each Buffs and Debuffs section.
- Fixed border rendering for pooled oUF aura buttons on Live 12.0.
- Consolidated duplicated border rendering into a shared implementation.
- Added all required localizations.

### PTR 12.1 Secret-Value Workaround

PTR 12.1 protects parts of the new aura system with secret values. Reading aura-button dimensions or using `BackdropTemplate` on frames anchored to protected aura geometry can cause taint or secret-number arithmetic errors.

As a workaround, textured aura borders no longer use Blizzard's backdrop renderer. The addon creates an owned border frame with an explicit public size and renders the border manually from eight texture slices: four corners and four edges.

This avoids reading protected aura dimensions while preserving border textures, offsets, colors, and layer settings. AuraKit also uses a style fingerprint and deferred restyling so pooled or protected aura buttons can be updated safely when settings change.

### Performance Considerations

This workaround may have a small performance impact because textured borders require eight additional texture regions per aura button and may need to be restyled when aura settings or frame layers change. The implementation already caches and reuses these regions and avoids unnecessary updates through style fingerprinting and deferred restyling. However, further profiling and optimization may still be useful, and an additional review would be appreciated to confirm whether this approach is appropriate for the new secret-value restrictions.

## Screenshots

Unitframe Aura Borders

image

Example 1
image
Example 2
image

Blizzard Auras

image

Example 1
image

Example 2
image

## Testing

- [x] Tested on Live 12.0
- [x] Tested on PTR 12.1
- [x] Tested in combat both Live and PTR 
- [x] Tested with the new PTR aura system
- [x] Existing profile migration tested
- [x] Immediate border restyling tested
- [x] No Lua errors encountered

### DIFF
--- a/EllesmereUI.lua
+++ b/EllesmereUI.lua
@@ -3013,16 +3013,15 @@ do
         return false
     end
 
-    --- Border color + Show Behind to apply when the user picks a border style.
-    --- Shadow -> black + behind on; everything else -> behind off (Shadow is the
-    --- only style that renders behind). Solid/Shadow default to black, other
-    --- textured styles default to white. Returns (colorTable, behindBool).
+    --- Border color + layer defaults to apply when a border style is selected.
+    --- Shadow is the only style rendered behind both its icon and Unit Frame.
+    --- Returns (colorTable, behindBool, behindUnitFrameBool).
     function EllesmereUI.GetBorderStyleSelectDefaults(textureKey)
-        if textureKey == "shadow" then return { r = 0, g = 0, b = 0 }, true end
+        if textureKey == "shadow" then return { r = 0, g = 0, b = 0 }, true, true end
         if not textureKey or textureKey == "" or textureKey == "solid" then
-            return { r = 0, g = 0, b = 0 }, false
+            return { r = 0, g = 0, b = 0 }, false, false
         end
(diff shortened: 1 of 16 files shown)


EllesmereGaming:
excellent work! you're killin it

PR #893, opened 2026-07-21
fix(blizzskin): guild rank SetGuildRankOrder taint from menu/roster skin

Filpet96:
## Bug report
(https://discord.com/channels/585577383847788554/1528850388860534785)
https://discord.com/channels/585577383847788554/1529094117404180633
## Summary

Fixes `ADDON_ACTION_FORBIDDEN` on `SetGuildRankOrder()` when changing a guild member's rank from the roster detail popup (`Menu.Pick` → `GuildRoster.lua`).

### Why we leave the member detail popup unskinned (instead of "fixing" its skin)

`GuildMemberDetailFrame` owns `RankDropdown`, which builds `MENU_GUILD_RANKS` and calls the protected `C_GuildInfo.SetGuildRankOrder` from `Menu.Pick`. Any addon write on that frame tree (backdrop `CreateTexture`, `AddBorder`, `FadeRegions`/`SetAlpha` on BG/Border, `CloseButton` skin, recursive `ButtonsIn`/`ControlsIn`) can leave the rank pick path tainted — BugGrabber then blames `EllesmereUIBlizzardSkin`.

We tried narrower approaches first (skip only the rank menu tag, owned chrome on the menu, fade-only detail chrome). They either still errored or broke the popup visually (no background) while still tainting. Blizzard's menu frames are also **pooled**, so parenting skin chrome onto a menu used for a safe context menu can poison a later rank dropdown pick.

So for this surface we deliberately **do not skin** `GuildMemberDetailFrame` at all and mark it `artExempt` so guild-window sweeps skip it. Stock Blizzard chrome here is preferable to a broken promote/demote action. A future taint-safe restyle would need an overlay that never writes Blizzard regions on that frame (same idea as the context-menu UIParent overlay) and would need in-game bisection before shipping.

### What this PR changes

- **Context menus**: stop mutating pooled `Menu` frames / `AddMenuAcquiredCallback`; paint via a UIParent overlay. Do **not** hook dropdown `OpenMenu` (guild ranks use that path).
- **Guild member detail**: remove `SkinGuildPopup(GuildMemberDetailFrame)`; `artExempt` so `ButtonsIn` / `ControlsIn` / art sweeps leave it alone.
- **Not reverted**: roster widen / ScrollBox lift / column header skin from #853 stay — those were the taint-*safe* layout path for notes/ranks and were not shown to be the `Menu.Pick` failure mode.

## Test plan
- [x] `/reload` with Guild & Communities skin + Reskin Popups/Menus enabled
- [x] Open guild roster → click a member → detail popup is **stock Blizzard** (background intact)
- [x] Change rank via the rank dropdown — applies with no `SetGuildRankOrder` / `ADDON_ACTION_FORBIDDEN`
- [x] Spot-check a normal right-click context menu still gets the dark overlay/border
- [x] Roster widen / Achievement Points column alignment still behaves as on main (#853 path)

### DIFF
--- a/EllesmereUIBlizzardSkin/EllesmereUIBlizzardSkin.lua
+++ b/EllesmereUIBlizzardSkin/EllesmereUIBlizzardSkin.lua
@@ -758,45 +758,85 @@ end
     -- Back-compat full init (data + visual), used by the live re-apply path.
     local function _ttInit() _ttInitData(); _ttInitVisual() end
 
-    -- Context menu skinning
-    local _menuSkinned = {}
+    -- Context menus: UIParent overlay only (menu frames are pooled; parenting taints picks)
+    local _menuOverlay, _menuOverlayTarget
 
-    local function _menuSkinFrame(frame)
-        if not frame or frame:IsForbidden() or not _pmEnabled() then return end
-        for i = 1, _select("#", frame:GetRegions()) do
-            local region = _select(i, frame:GetRegions())
-            if region and region:IsObjectType("Texture") and not GetFFD(region).owned then
-                local RS = EllesmereUI.RESKIN
-                region:SetColorTexture(RS.BG_R, RS.BG_G, RS.BG_B, 1)
-                region:SetAlpha(RS.CTX_ALPHA)
-                region:ClearAllPoints()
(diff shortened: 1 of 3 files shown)


EllesmereGaming:

Issue #871, opened 2026-07-20
[Feature Request] CDM anchor point / row expansion behavior


KyleYueye:
Hi! First of all, thank you for all the work you’ve put into EllesmereUI. I’ve noticed a small issue with the CDM’s positioning.

## Problem
Currently, the CDM only allows configuring the **anchor direction**, but there is no option to define the **frame’s own anchor point**. As a result, when the number of cooldowns changes between one row and two rows, the entire frame shifts position. For example, I may place the CDM exactly where I want it while there is only a single row of icons. Once a second row appears, the whole frame moves because its origin changes.

I noticed there is already an option to **keep the first row fixed when a second row appears/disappears**, but this introduces another problem: **what exactly is considered the “first row”**? Some players expect the top row to remain fixed, while others prefer the bottom row to remain fixed. Since this is a matter of personal preference, the current option doesn’t fully solve the positioning issue.

### Example
Set perfect position for two row CDM
<img width="1736" height="451" alt="Image" src="https://github.com/user-attachments/assets/51368d41-4138-4a73-a83e-c2e62fb6c164" />

Position shifted for one row
<img width="1686" height="452" alt="Image" src="https://github.com/user-attachments/assets/1a561678-8e22-47fb-a725-bd6f73739862" />

### Option 1 (preferred):
Allow users to configure the CDM frame’s own anchor point (TOP, BOTTOM, CENTER, etc.). The selected anchor would remain fixed in UI space, and the icons would expand relative to that anchor. This would completely eliminate the frame shifting when rows are added or removed.

### Option 2:
Improve the existing “**Keep first row fixed when the second row appears/disappears**” option by allowing users to define which row is considered the “first row” (top row or bottom row). This would make the current behavior much more flexible for different UI layouts.

I think either of these changes would make positioning much more predictable and customizable.

Thanks again for all your work!

KyleYueye:
Some additional technical context after digging into the implementation, in case it helps with evaluating the two options:

## Where the current behavior lives

The anchor point resolution is handled by `ns.ResolveGrowAnchorPoint(barData, ignoreFirstRow)`. It works in two steps:

1. The grow direction is mapped to its opposite edge (`RIGHT` → `LEFT`, `DOWN` → `TOP`, etc.), so the fixed edge stays put while the bar resizes along its growth axis. The perpendicular axis is left unpinned (centered) — which is exactly why a horizontal bar re-centers vertically when a second row appears.
[...]

PR #912, opened 2026-07-22
feat(cdm): replace row pin with a Row Growth direction option

KyleYueye:
## What does this PR do?

Fixes the CDM bar position-shifting problem described in #871 by remodeling how players control multi-row expansion.

**Why remodel instead of just adding a mirror option?**

When the icon count changes at runtime (procs, spec swaps, cooldowns spilling into a second row), a player's placement intent has exactly two dimensions:

1. **Which row is the anchor** — the row they positioned against their action bars / resource bars and expect to never move; and
2. **Which way extra rows grow** — up, down, or symmetrically from the center.

The previous model expressed neither directly. "Anchor First Row" forces the player to first understand what the "first row" is (an internal fill-order concept), and only covers the "top row stays" case — players who park the CDM *above* their character get the opposite of what they want. Extending it with a mirrored "Anchor Last Row" toggle would double the checkbox surface and still leave two mutually-exclusive toggles encoding what is really one three-state choice.

**User-facing changes:**

- **Row Growth** dropdown replaces the "Anchor First Row" toggle (Number of Rows cog, 2-row layouts):
  - *Grow Centered* — legacy default, bar re-centers symmetrically;
  - *Grow Down* — the base row is pinned on top, extra rows grow downward;
  - *Grow Up* — the base row is pinned at the bottom, extra rows grow upward (**new capability**: pinned-row icons never move on screen, implemented via visual row-order reversal since icons lay out from the frame's TOPLEFT);
  - Labels adapt to vertical bars (*Grow Right / Grow Left*).
- **Custom Base Row Count / Base Row Icons** replace the four Top/Bottom row count controls. The *base row* is direction-neutral: it fills first and it is the row Row Growth keeps in place — so its meaning never flips when the growth direction changes. Per-row size controls are relabeled *Base / Second Row* accordingly.
- The **options preview** now mirrors the live bar's visual row order.

**Compatibility:**

- `anchorFirstRow` migrates to `rowGrowDirection` ("DOWN"/"RIGHT") via `cdm_row_grow_direction_v1`; profile *imports* get the same conversion via an import-path forward-copy (registered migrations are skipped for imported profiles due to inherited migration flags).
- `topRowCount` already stored the base row, so existing top-row users keep their exact layout with zero conversion. Legacy `bottomRowCount` fields are still honored at runtime; only their UI was removed.
- Partner-sync exclusions cover the new key (positioning-adjacent, like the fields around it).

## How was it tested?

Live retail client (8.5.3 codebase):

- Horizontal bars, all three Row Growth values × 1↔2 row transitions (adding/removing tracked spells): pinned row's icons hold position pixel-exact; Grow Centered matches legacy behavior.
- Custom Base Row Count with reversed growth: the pinned row keeps the configured count regardless of direction.
- Base/Second Row Size offsets with reversed growth.
- Options preview row order matches the live bar in all combinations; drag-reorder in the preview still targets the correct slots.
- Vertical Orientation flip resets Row Growth to Centered (same policy as Grow Direction).
- Migration verified on a profile carrying `anchorFirstRow`; `/eui migrations` reports `cdm_row_grow_direction_v1` applied, bar renders identically.

Not yet verified on the 12.1 PTR client (no PTR access; changes are pure layout math with no new API surface).

## Screenshots
Before new modifications. We had set a perfect position for 1 row CDM, but it shifted vertically downward when it becomes 2 rows
屏幕截图 2026-07-22 210833
屏幕截图 2026-07-22 210908

After new modifications. Option sets to "Grow Up". 1 row versus 2 rows
屏幕截图 2026-07-22 211021
屏幕截图 2026-07-22 211042

After new modifications. Option sets to "Grow Down". 1 row versus 2 rows
屏幕截图 2026-07-22 222917
屏幕截图 2026-07-22 223118

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) — Row Growth defaults to *Grow Centered* (stored as nil), which is the exact legacy layout; migrated `anchorFirstRow` users keep their pin.
- [x] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built — the option is a stored field read inside the existing layout pass.
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations) — reversal is an index remap (`effRows - 1 - row`) computed once per layout.
- [x] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames — this PR only changes offsets fed to the existing layout code.
- [x] Tested in-game, works on live retail;


realmsg:
Thank you ,This is exactly what i need. Hope it gets merged into main soon!

PR #917, opened 2026-07-22
Fix Edit Mode secret-value taint warnings (two injection points)

dfrisone:
## Summary

Toggling Blizzard **Edit Mode** with the suite loaded emitted large `LUA_WARNING` storms in instanced content: `attempt to compare/perform arithmetic on a secret number value (execution tainted by 'EllesmereUIUnitFrames')` from `CompactUnitFrame_UpdateHealthColor`, the encounter warnings refresh, `SecureUtil` selection math, and Cooldown Viewer item data. Worse, when Edit Mode's party frame setup pass ran tainted, the compact party frames stayed poisoned for the rest of the session, erroring on every subsequent health update (observed 100+ occurrences). Reported against v8.5.3 (Chat setup + Edit Mode positioning was the trigger scenario; the Chat module was not involved).

## Root causes (two independent injection points)

**1. Inline re-parent hook in the embedded oUF library's `DisableBlizzard`.**
When the library disables a Blizzard frame it re-parents it to a hidden frame and installs a `SetParent` `hooksecurefunc` that yanks the frame back **inline inside the hook**. Blizzard's Edit Mode layout pass calls `SetParent` on managed containers, confirmed live with a print hook: `BossTargetFrameContainer -> UIParent` on every Edit Mode enter/exit. The inline yank therefore ran in the middle of that secure execution, tainting everything Edit Mode touched afterwards, and every 12.x secret value downstream became a warning.

Because the library is a packager external (`.pkgmeta`), the fix is a tracked-code override in `EllesmereUIUnitFrames.lua`: `DisableBlizzard` is replaced for the units this addon actually disables (player/pet/target/focus/boss) with the same disable steps, but the re-parent is **deferred to a timer callback** (a fresh addon-owned execution), further postponed while the Edit Mode manager is shown (SetParent fires Blizzard's synchronous layout handlers in the caller's execution), with the stock combat-safe regen sweep preserved. Units this addon never disables fall through to the stock implementation.

**2. Module `OnEnable` running inside Blizzard's secure demand-load executions.**
`EllesmereUI_Lite`'s enable queue flushed synchronously inside any `ADDON_LOADED` dispatch, including ones fired from Blizzard's own secure executions, e.g. UIParent demand-loading `Blizzard_CombatLog` during login (captured in `taint.log`: module `OnEnable` bodies executing under `UIParent.lua -> CombatLog_LoadUI -> LoadAddOn`). Everything those bodies wrote became tainted fields born mid-secure-chain, which Edit Mode's enter/exit passes later read. The flush is now deferred one tick via `C_Timer.After(0)` so `OnEnable` always runs in a fresh addon-owned execution. Enable order is unchanged; `OnEnable` lands one tick later at login (reviewers should be aware of that timing change, no behavioral difference was observed in testing).

## Diagnosis method

Bisected in-game across ~10 controlled sessions in a follower dungeon (secret values are active in instances, not the open world, open-world tests are silent even when tainted): module isolation, a minimal dummy addon, live `SetParent` instrumentation, and `taintLog 2` capture. Notably, `taint.log` showed **no** named-global reads at the toggle, which is what pointed at tainted **table fields** (invisible to the logger) and ultimately at the enable-queue flush.

## Verification

In a follower dungeon, after both fixes:
- ESC/game-menu Edit Mode open/close: clean, repeated toggles.
- Typed `/editmode` open/close: clean (previously the worst path, 100+ warnings).
- No session-long compact party frame error spam afterwards.
- Player/target/focus/boss frame replacement behavior unchanged (the override reproduces the stock disable steps).

Both files pass a Lua 5.1 syntax check.

### DIFF
--- a/EllesmereUI.lua
+++ b/EllesmereUI.lua
@@ -12514,7 +12514,11 @@ function EllesmereUI.SetPlayerCastBarSuppressed(owner, suppressed)
         end
         EllesmereUI._GetFFD(blizzBar).castBarSuppressed = true
 
-        if blizzBar:GetParent() ~= hiddenParent then
+        -- Skip the re-parent while Edit Mode is open: SetParent fires
+        -- Blizzard's synchronous layout handlers in this execution, and the
+        -- UnitFrames Edit-Mode-close hook re-applies this state afterwards.
+        if blizzBar:GetParent() ~= hiddenParent
+            and not (EditModeManagerFrame and EditModeManagerFrame:IsShown()) then
             blizzBar:SetParent(hiddenParent)
         end
 
@@ -12549,10 +12553,14 @@ function EllesmereUI.SetPlayerCastBarSuppressed(owner, suppressed)
             if not EllesmereUI._GetFFD(selection).showHooked then
                 EllesmereUI._GetFFD(selection).showHooked = true
                 hooksecurefunc(selection, "Show", function(self)
-                    if PlayerCastingBarFrame and EllesmereUI._GetFFD(PlayerCastingBarFrame).castBarSuppressed then
-                        self:SetAlpha(0)
-                        self:EnableMouse(false)
-                    end
+                    -- Deferred: Show fires inside Edit Mode's secure
+                    -- ShowSystemSelections pass; write nothing there.
+                    C_Timer.After(0, function()
+                        if PlayerCastingBarFrame and EllesmereUI._GetFFD(PlayerCastingBarFrame).castBarSuppressed then
+                            self:SetAlpha(0)
+                            self:EnableMouse(false)
+                        end
+                    end)
                 end)
             end
         end
@@ -12562,7 +12570,8 @@ function EllesmereUI.SetPlayerCastBarSuppressed(owner, suppressed)
 
     EllesmereUI._GetFFD(blizzBar).castBarSuppressed = false
 
-    if hiddenParent and blizzBar:GetParent() == hiddenParent and EllesmereUI._GetFFD(blizzBar).origParent then
+    if hiddenParent and blizzBar:GetParent() == hiddenParent and EllesmereUI._GetFFD(blizzBar).origParent
+        and not (EditModeManagerFrame and EditModeManagerFrame:IsShown()) then
         blizzBar:SetParent(EllesmereUI._GetFFD(blizzBar).origParent)
     end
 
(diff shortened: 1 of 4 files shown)


dfrisone:
Pushed a follow-up commit (90a1099f): a tester on the fixed build in a real public group still saw a small exit-only residue (3x CompactUnitFrame + 3x SecureUtil; the storms and session poisoning were confirmed gone). This pass applies the same deferral rule to every remaining Edit-Mode-adjacent site that did inline work in a hook or event dispatch: the EDIT_MODE_LAYOUTS_UPDATED cast bar handler, the Blizzard-style class power SetParent/Hide hooks (armed only when Class Resource is set to Blizzard, which the earlier test matrix did not cover), and the shared cast bar suppression's SetParent/Selection writes. Awaiting tester re-verification.

dfrisone:
Follow-up dca9c4fc: the tester's config has Resource Bars disabled, ruling out the cast bar suppression cluster for their residue, which pointed at frame-source-gated paths instead. Found and fixed a fourth inline site: the native ToT/FoT child suppression (armed when the parent target/focus frame is on the Blizzard source with the mini frame set to EUI/hidden) re-hid the child inline from its OnShow hook; with a Blizzard-source parent that frame is a live Edit Mode system whose preview pass shows the child inside the secure execution. The re-hide is now deferred with the same manager-open postpone used elsewhere in this PR. Blizzard-source parents also bypass the DisableBlizzard override entirely, which fits the tester-only repro.

dfrisone:
## Additional verification: hard-error reproduction with a deterministic A/B

Follow-up testing found a configuration where this taint manifests as a genuine **error**, not just a cosmetic `LUA_WARNING`, and confirmed the fix clears it with a controlled before/after on a single machine.

**Reproduction (stock, without this branch):** with **Arena Frames** enabled in Blizzard Edit Mode, entering Edit Mode inside instanced content throws:

```
attempted to iterate a table that cannot be accessed while tainted
(execution tainted by 'EllesmereUIUnitFrames')
CastingBarFrame.lua:722  StopFinishAnims
CastingBarFrame.lua:86   UpdateShownState
EditModeSystemTemplates.lua:1645  SetIsInEditMode
EditModeManager.lua:2375  RefreshArenaFrames
...
EditModeManager.lua:88   EnterEditMode
```

The frame is the arena castbar preview (`CompactArenaFrame`), and the blocked iteration is over its forbidden stage-pip table. This addon does not hook that frame, so it is the same underlying cause as the previously reported warning storms (a UnitFrames field tainted at login and later read during Edit Mode's enter pass), just reaching a Blizzard code path that iterates a secret table and therefore errors outright instead of logging a warning.

**Classification:** disabling only the UnitFrames module on the affected client makes it clean, confirming the taint originates from a UnitFrames field rather than generic slash-command taint shared across the ecosystem.

**Before/after, same account, config, and instance:**

| Entry path | Stock | This branch |
|---|---|---|
| Typed `/editmode` | error reproduces | clean |
| ESC menu | error reproduces | clean |

Both entry paths trip `RefreshArenaFrames` on Edit Mode enter, and both are clean on this branch across repeated enter/exit toggles.

**Regression recipe:** stock build, enable Edit Mode Arena Frames, enter a dungeon, open Edit Mode.

PR #925, opened 2026-07-23
[Translation] Adding a missing string

Crazyyoungs:
Hello,

I noticed a couple of missing translations from the post I submitted this morning, so I'm adding them here.

## What does this PR do?

This PR adds the missing Korean (koKR) translation for the `Swap Light/Reckless Pots When Missing` option, which was omitted from this morning's update.

```lua
L["Swap Light/Reckless Pots When Missing"] = "물약 부족 시 주능력치/보조능력치 물약 교체"
```

## How was it tested?

Verified the localization file syntax and checked that the text displays correctly without errors.

## Screenshots

N/A (Text-only localization update)

## Checklist


EllesmereGaming:
i believe these were added in the patch ~20 minutes ago

PR #928, opened 2026-07-23
Fix oversized AuraBuff reminders after dungeon reload

Nnoggie:
## Summary
- Normalize AuraBuff Reminder button scaling during layout.
- Prevent secure reminder icons from rendering oversized after reloading inside a dungeon.
image
image

## Why this is needed
When the UI is reloaded inside a dungeon, the protected reminder buttons can keep an effective scale of 1 even though UIParent and the reminder anchor use the configured UI scale. Their logical icon size remains correct, but they render much larger than intended. Reloading outside and then entering the dungeon does not trigger the problem.

The fix avoids the unreliable inherited scale on these secure buttons and explicitly applies UIParent's current effective scale. This keeps icon sizing consistent in both reload paths.

The screenshots show the oversized icons after an in-dungeon reload and the expected icon size after a normal outside reload.


EllesmereGaming:
oh i actually fixed this locally right before this lol sorry! pushing the fix out in a few

PR #962, opened 2026-07-24
i18n: Add localization support

Sanxy09:
## What does this PR do?
1. Fallback to client spell name when localization is missing
2. Uppercase section names
3. Add localization support

## How was it tested?
Tested on live retail

## Checklist


EllesmereGaming:
hey thank you for the help! some brief notes before i can merge:

1. Revert the four page = "Display" -> "DISPLAY" edits in EUI__General_Options.lua. That's the Nameplates page name, not a section header, it's registered as "Display" (EUI_Nameplates_Options.lua:18) and SelectPage is case-sensitive, so these deep-links break.

2. SPELL_NAME_BY_ID: on English clients loc == nm always, so the resolver now returns GetSpellName(id), which can be nil for uncached IDs, where we previously always had the stored name. End with ... or nm.

3. The section-header renames are locale key renames. SectionHeader doesn't uppercase, so "Display" -> "DISPLAY" etc. orphans every existing translation of the old keys (same for "no buff assigned" -> the new composite key). The all-caps style is fine, but please rename the keys in the locale files too, or list the orphaned keys so locale maintainers can follow up.

4. Small: keep the type(v) == "string" guard in the Locale.lua reverse-map loop, and Lf("Overlay Preview") with no args should be L(...).

5. Please split unrelated fixes into separate PRs: the tooltip/border fixes should be separated from this, makes it much harder to merge PRs when they combine multiple things.

thank you again for your contributions <3

Sanxy09:
Thanks! I have cleaned up this PR to focus **strictly on localization / i18n support**.

### Changes made to this PR:
1. **Addressing Review Feedback:**
   - Synchronized all section header keys and translated string values to **ALL-CAPS** across all locale files
   - Added `or nm` fallback to `SPELL_NAME_BY_ID` logic for missing localized spell names.
   - Reverted `page = "Display"` back to its original state as requested.

2. **PR Scope Separation:**
   - Reverted all non-localization UI & rendering changes (slider tooltips, debuff border visibility, preview icon offset, and disabled tooltips).

---

### Dedicated Follow-up PR:
The separated UI/border fixes have been submitted in a dedicated PR here: #981

PR #969, opened 2026-07-24
feat(databars): latency block shows home, world or both

Absol3m:
## What it does

The latency block showed **one** link — home or world, chosen by a `World Latency` on/off toggle. To see both, you had to add a second block.

It now has a **Latency** dropdown: **Home**, **World**, or **Both**. In *Both* mode each value carries its own icon — a house for home, a globe for world — so the two numbers are self-labelling rather than relying on order:

```
Home    🏠 45ms
World   🌐 92ms
Both    🏠 45ms  🌐 92ms
```

The **icon is opt-in** (`Show Icon`, off by default), so an existing block updates without its bar suddenly widening. The old `useWorldLatency` setting migrates in place: a block that showed world keeps showing world.

## Colours live in the tooltip

The bar stays in the block colour, like every other module — no red/yellow/green on the bar itself. Connection quality is shown where it belongs, in the tooltip, which now:

- colours **home and world** green / yellow / red by latency, from the client's own `GREEN`/`YELLOW`/`RED_FONT_COLOR` globals (thresholds 100 / 250 ms)
- adds **download / upload** bandwidth — the two `GetNetStats` values that were previously thrown away and appear on no bar at all

So a hover always adds what the bar cannot: the quality read and the bandwidth. `GetLatColor` used to fade to white at the good end, which never read as "good"; it is now three discrete bands.

## Implementation

Showing one or two icon-labelled segments is beyond the shared single-icon `MakeStatBlock`, so the latency block gets its **own factory**. `MakeStatBlock` is left at exactly its current form — the icon lookup, icon-off-by-default and value-colour hooks that this rework briefly added to it were removed, so `fps` and `durability` are untouched. A single `ns.LatencyMode(settings)` resolver backs both the block and its options row, so they can never disagree on what is selected.

## Notes for translators

Two new option strings via the module's `ns.L` table (so invisible to the key extractor, `Locales/_keys.txt` unchanged): the `Latency` dropdown label and its `Both` value. `Home` / `World` already exist. The tooltip's `Download` / `Upload` / `KB/s` were added in the same table.

## Testing

Home / World / Both on a horizontal bar and a vertical bar; `Show Icon` on and off; an existing block (migrated from `useWorldLatency`); the tooltip's colours and bandwidth.

`luac -p` clean on all three files.

## Checklist


EllesmereGaming:
hmmmm so the thing is, this was a deliberate choice by me to allow users to space things out how they wanted it to be. can you screenshot the new settings options and the visuals in game so I can get a preview of how spacing works?

Absol3m:
Hi, I fully understand what you’re saying, but one doesn’t rule out the other – we can still choose whether or not to display them, just as you used to. We can choose whether or not to display the icons, and whether or not to display one, the other, or both latency values.
From what I’ve read on Discord, lots of users would like to see the features they had in XIV DATABAR restored, and this change is a step in that direction.
I’m trying to respond as fully as possible to the DATABAR requests without compromising your original work.

image

image

image

image

EllesmereGaming:
that's perfect thanks! screenshots like that help a lot with visualizing new features. please try to include them anytime you introduce a new setting. excellent work :D

PR #972, opened 2026-07-24
feat(castbars): Add cast bar customizations. Also added canceled / interupted feedback for cast bars.

Ricoder92:
## Summary

This PR expands cast bar customization across Player, Target, Focus, and Boss unit frames while improving border alignment and cast outcome feedback.

## Changes

- Added border customizations for Player, Target, Focus, and Boss cast bars.
- Added selectable cast bar textures for Player, Target, Focus, and Boss cast bars.
- Unified border behavior between the cast bar and detached cast icon.
- Added an optional pixel-perfect divider between the icon and bar when the icon is part of the cast bar (for all types of cast bars).
- Fixed some border alignment on the cast bar frame.
- Added configurable cancelled and interrupted cast feedback:
  - Player, Target, Focus, and Boss support
  - Separate enable toggles
  - Separate colors
  - Configurable display durations
  - Interrupted casts use red by default
  - Cancelled casts use orange by default
- Temporarily overrides `Hide When Idle` while a cancelled or interrupted cast message is displayed.
- Prevents cancelled or interrupted feedback from appearing when Target, Focus, or Boss dies during a cast.
- Ensures new casts immediately replace pending cast outcome messages.
- Added and updated localization entries for the new settings and status messages.

## Result

Cast bars now provide more consistent visual customization across unit frames, improved pixel-perfect border rendering, and clearer feedback when casts are cancelled or interrupted.

### Settings
image
Player, Target, Focus and Boss are the same

### Interrupted 
Screenshot 2026-07-25 012825
(target castbar)

### Cancelled
image
(bottom cast bar is player, middle one is target)


EllesmereGaming:
Yo another banger PR from you man thank yoU! The core of what i need fixed is that any PR I intake needs to guarantee nothing changes for existing users unless it's specifically a bugfix. Currently there are a few issues on that front:

1. `castbarTexture = "none"` in defaults kills the health-bar fallback. Defaults get written into the profile, so `s.castbarTexture` is always truthy and cast bars lose the health bar texture they used to match. Leave the key out of defaults so it stays nil.
2. `if inWidth then offX, offY = 0, 0 end` silently zeroes existing cast icon offsets and makes the sliders no-ops (in-width is the default).
3. 'The outcome listener isn't gated on the toggles' - it registers UNIT_HEALTH for everyone, and its death/target-change branch hides the cast bar before the enabled check.
4. `interrupted and s.interruptedCastEnabled or s.cancelledCastEnabled` falls through, so "Show Cancelled" alone also shows interrupted casts. Needs an if/else. Same line in both files.

Ricoder92:
Thank you for your feedback and your code review! :)

I created a new commit:

1. removed castbarTexture = "none" from the defaults
2. Changed `if inWidth then offX, offY = 0, 0 end` to:
  ```
local layoutOffX = inWidth and 0 or offX
local layoutOffY = inWidth and 0 or offY
```
and changed following offX/Y to the new variables

3. The listener now registers health, death, and spellcast events only when either Cancelled or Interrupted display is enabled. A second guard inside the handler prevents the feature from hiding cast bars when both toggles are disabled.

4. Replaced the and/or expression with an explicit if/else. This prevents a disabled Interrupted setting from falling back to the Cancelled setting and incorrectly showing interrupted casts. The fix was applied in both files.


Im hoping its fine now :)

DoGi114:
Is this still coming into EUI ?
Im missing those things too. Tho at this moment I've just inherit border from health frame.

MarioCatuogno:
+1 for this feature!

DoGi114:
Not coming anymore ? :(

MarioCatuogno:
Hi @Ricoder92 So? It has been added or discarded?

DoGi114:
Hm?

Ricoder92:
I will rework it in the upcomming days and reopen the pullrequest.

Ricoder92:
see #1987

PR #973, opened 2026-07-25
Play the CDM cooldown-ready sound on the real cooldown-end edge

nulltyto:
### Reported issue

A "Cooldown Ready Sound" set on **Temporal Anomaly** (Preservation Evoker) never played when the cooldown finished. Instead it played at the exact moment the spell was cast again. The same setting on Verdant Embrace and Dream Breath worked correctly, so it looked spell-specific.

### Cause

Two defects meeting on one spell.

**The expiry edge was never seen.** The per-spell "Audio Effect on CD Ready" was driven only by `SPELL_UPDATE_COOLDOWN` / `SPELL_UPDATE_CHARGES`, but the client dispatches no event when a cooldown expires. Blizzard says so in `CooldownViewerItemMixin:TriggerAvailableAlert` itself:

> Need to refresh the entire button state after the cooldown finishes, this is what simulates the client sending a final SPELL_UPDATE_COOLDOWN event

and drives that edge from the viewer's `OnUpdate` against the real end time instead. So the sound waited for the next *unrelated* cooldown event — which, for a spell the player sits on, is their next cast — and the arm stayed set the whole time.

**The stale arm then fired on the cast.** `CdReadyIsReady` treated "on GCD" as ready, and a cast-time spell only puts its own cooldown up on cast SUCCESS. From cast start until then, `GetSpellCooldown` reports the GCD alone (`isActive` + `isOnGCD`), which read as ready and fired the armed sound. Temporal Anomaly has a 1.5s cast, so it hit this every time; instant spells like Verdant Embrace start their cooldown and GCD together and never show the state, which is why only one spell looked broken.

### Change

- Post-hook `TriggerAvailableAlert` as the primary driver for non-charge spells — the same taint-safe idiom `EnsureBuffSoundHook` already uses on `TriggerAuraAppliedAlert`, and it needs no secret duration read.
- Charge spells stay on `SPELL_UPDATE_CHARGES`: the alert fires when the *first* charge returns, while CDM charge readiness means back at max.
- The event pass keeps arming, charge readiness and cooldown resets, and still backs up the expiry in the gaps between global cooldowns.
- Firing takes a strict ready test (`not isActive`) so a GCD can no longer be mistaken for an expiry; arming keeps the loose read so a GCD still never arms a spell that was never on cooldown.
- Both drivers play through one throttled `PlayCdReadySound`, so an alert and an event landing on the same cooldown end a frame apart cannot double-fire.

EUI only alpha-fades the Blizzard viewers rather than hiding them (`EllesmereUICooldownManager.lua:6812`), so `CooldownViewerMixin:OnUpdate` — and therefore the alert — keeps running behind EUI's own bars.

### Testing

Reporter confirms Temporal Anomaly now sounds at cooldown end and is silent on the cast. `/cdmreadydbg` tags each fire with the driver that won the edge (`FIRE(alert)` / `FIRE(event)`), which is the quickest way to check a given spell.

Worth a second pair of eyes on: cooldown resets, a 2-charge spell, spec/talent swaps, and zone/login boundaries (the loading-screen settle gate is unchanged but the alert is a second path into the sound). One intentional behaviour change: the alert path does not apply the 1.6s minimum-arm floor the event fallback uses — Blizzard requires only >0.75s — so a very short cooldown with a sound assigned can now actually play.

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
+++ b/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
@@ -1196,9 +1196,11 @@ end
 --  fires on login (already ready -> never armed), on a GCD ending (a GCD never
 --  makes a spell "not ready" here), or per render tick.
 --
---  Driven ENTIRELY by the authoritative cooldown/charge events
---  (SPELL_UPDATE_COOLDOWN + SPELL_UPDATE_CHARGES) via the WatchCdReadySoundIfEnabled
---  set -- NOT by the SetDesaturated visual hook. That hook fires at moments unrelated
+--  The expiry edge itself comes from Blizzard's TriggerAvailableAlert (see
+--  HookCdReadyAvailableAlert): the client sends no event when a cooldown runs out,
+--  so the cooldown/charge events below cannot see it on their own. They still drive
+--  arming, charge readiness, and resets via the WatchCdReadySoundIfEnabled set
+--  -- NOT the SetDesaturated visual hook. That hook fires at moments unrelated
 --  to the real cooldown (rage/range/GCD repaint) where C_Spell.GetSpellCooldown can
 --  report a transient isActive=true / isOnGCD=false GCD race; sampling there
 --  false-armed non-charge spells that were never on cooldown (Odyn's Fury etc.),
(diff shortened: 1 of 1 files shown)


nulltyto:
I did test reset, 2-charge spells, spec swaps, and zone changes, but doesn't hurt to have another round of tests.

PR #977, opened 2026-07-25
Nameplates: add No Tint option for Target/Focus Texture

anthonycosson:
## What does this PR do?

Adds a "Don't tint (keep bar's own color)" toggle inside the existing Target Texture / Focus Texture cog popups (Nameplates -> Display -> "Target, Focus & Hover Effects" section). When enabled, the pattern picked in the Target Texture / Focus Texture dropdown is applied directly as the health bar's own fill texture (like the main "Bar Texture" setting) instead of being drawn as a tinted stripe overlay on top of it. This lets you give your current target/focus a distinct pattern without also forcing a separate overlay color onto it -- the bar keeps its normal reaction color (or the custom target/focus color, if that toggle is enabled).

Both existing behaviors (overlay tint, custom target/focus color) are untouched when the new toggle is off, which is the default.

## How was it tested?

Tested in-game on live retail: enabled "Don't tint" in the Target Texture cog on a training dummy target, picked a fill pattern, and confirmed the targeted nameplate shows that pattern while its health bar color stays the normal reaction color (screenshot below). Also verified toggling it back off reverts to the normal overlay behavior, and that the Opacity / "Full alpha on empty part of bar" rows in the same popup grey out while No Tint is active. Not tested on the 12.1 PTR client.

## Screenshots
image

## Checklist

- [x] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames -- N/A, this only touches EllesmereUI's own health bar / overlay frames
- [ ] Tested in-game, works on live retail; no load errors on the 12.1 PTR client -- tested on live retail only, PTR not verified

### DIFF
--- a/EllesmereUINameplates/EUI_Nameplates_Options.lua
+++ b/EllesmereUINameplates/EUI_Nameplates_Options.lua
@@ -7481,6 +7481,14 @@ initFrame:SetScript("OnEvent", function(self)
         local isTargetTextureNone = function()
             return (DBVal("targetOverlayTexture") or defaults.targetOverlayTexture) == "none"
         end
+        -- No Tint: the target texture pattern becomes the bar's own fill
+        -- texture (SetStatusBarTexture) instead of a tinted overlay drawn on
+        -- top, so the color swatch/opacity below stop applying.
+        local isTargetNoTint = function()
+            local v = DBVal("targetOverlayNoTint")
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
well done PR, thank you!

PR #978, opened 2026-07-25
Stop the DataBars gold tooltip leaking characters through shared profiles

dfrisone:
## Problem

Reported by @TeoDaTank and reproduced locally: importing someone's profile showed
**their** characters and gold in the DataBars Gold tooltip.

## Cause

The cross-character gold store was declared in `defaults.profile`:

```lua
characters = {},  -- cross-character gold store, keyed "Name-Realm"
```

So it lived in the DataBars addon blob, and exports deep-copy those blobs
wholesale. Every exported profile carried the author's character names and gold,
and because the tooltip read the store from the active profile, importers saw the
author's characters instead of their own.

This is observed character data, not a setting, so profile scope was wrong to
begin with.

## Fix

Moved to `EllesmereUIDB.global.dataBarsCharacters`, the account-wide store the
Friends module already uses. Every read and write funnelled through `GoldStore()`,
so the move is contained to that accessor.

Three touch points, because stale data and old export strings are already out
there:

- **Migration** drops the per-profile copies, so nothing stale can re-export.
- **Export** strips the key (both export paths), so a string taken before the
  migration runs is clean too.
- **Import** strips it, so a sharer's characters never reach the recipient's
  SavedVariables at all.

The three call one `StripSharedCharacterData` helper, which documents the key
ordering it depends on: local folder keys, so before `AddonsToCanon` and after
`CanonToLocal`. That matters because the folder literal is rewritten by the
standalone packager, so getting the order wrong would no-op on standalone builds
only.

## Deliberate decision: the migration drops rather than merges

It would be nicer to merge the old per-profile stores into the new account-wide
one so no history is lost. It is not safe: any profile may be an imported one and
nothing durably records that, so a merge cannot tell the user's own characters
from a sharer's and would copy strangers' names and gold into permanent
account-wide storage -- worse than the bug being fixed. Little is lost in
practice, since each character records itself again on login via the money
watcher.

## Side effects

- Switching or copying profiles no longer loses gold history (account scope).
- Resetting a profile no longer clears it either, which follows from the same
  change.
- A full UI Reset still wipes it, unchanged from today. The reset preserves
  specific `global` keys (friend groups/notes/assignments) and this is not among
  them; happy to add it if you'd rather it survived like friend data does.

## Testing

- Import the reporter's profile, hover Gold: only your own characters listed.
- Export via the options Export button, import elsewhere: only that account's
  characters listed.
- Log two characters, switch profiles: history survives (it did not before).
- Ctrl+Alt+left-click a character row still forgets it.
- `/eui migrations` shows `databars_gold_store_account_wide_v1` applied cleanly.

### DIFF
--- a/EllesmereUIDataBars/EllesmereUIDataBars.lua
+++ b/EllesmereUIDataBars/EllesmereUIDataBars.lua
@@ -159,7 +159,10 @@ local defaults = {
     profile = {
         nextBarId  = 0,   -- monotonic bar id counter; only ever incremented
         bars       = {},  -- ordered array of barCfg
-        characters = {},  -- cross-character gold store, keyed "Name-Realm"
+        -- No cross-character gold store here: it is account-wide, at
+        -- EllesmereUIDB.global.dataBarsCharacters (see GoldStore in
+        -- EllesmereUIDataBars_Blocks.lua). Keeping it per-profile shipped the
+        -- sharer's character names and gold inside exported profiles.
     },
 }
 
--- a/EllesmereUIDataBars/EllesmereUIDataBars_Blocks.lua
+++ b/EllesmereUIDataBars/EllesmereUIDataBars_Blocks.lua
@@ -1125,11 +1125,21 @@ local function GoldCharKey()
     return (UnitName("player") or "Unknown") .. "-" .. (GetRealmName() or "Unknown")
 end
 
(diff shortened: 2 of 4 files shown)


EllesmereGaming:
manually merging parts of this to combo with dereks fix

PR #997, opened 2026-07-25
[Feature] Delete the entire CDM.

Kirihasio2:
## What does this PR do?

It deletes the CDM

## How was it tested?

12.0.7, 12.0.1

## Screenshots

image

## Checklist

- [ ✔] New settings default **OFF** (no behavior change without opt-in)
- [✔ ] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built
- [ ✔] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations)
- [✔ ] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames
- [✔ ] Tested in-game, works on live retail; no load errors on the 12.1 PTR client


EllesmereGaming:
if only it were that easy

PR #998, opened 2026-07-25
Make releasing Blizzard's cast bar actually restore it

dfrisone:
Follow-up to #991. Turning EUI's cast bar off left Blizzard's cast bar unusable, in two different ways depending on how you got there. Both halves are here because neither alone makes the behaviour correct: one fixes where the frame goes, the other fixes whether its events are alive.

## 1. Released into a hidden parent

The release path handed the frame back to the parent it was captured from, and that parent is `PlayerFrame` — which EUI hides whenever it renders its own player frame. Edit Mode does the same from the other side, re-parenting the bar into a layout frame that gets hidden on exit. Either way the bar ends up armed and unreachable, indistinguishable from the suppression never lifting.

The unconditional `SetUnit("player")` that #991 removed had been masking this by making Blizzard re-home the frame as a side effect.

Confirmed in-game by dump, in this order:

| Check | Result |
|---|---|
| `IsEventRegistered("UNIT_SPELLCAST_START")` | `true` — #991's ownership code working |
| `_playerCastBarSuppressors` / `castBarSuppressed` | `{}` / `false` — release ran, ownership cleared |
| `_GetFFD(PlayerCastingBarFrame).origParent` | `PlayerFrame` — restored into a hidden parent |

Release now falls back to `UIParent` when the captured parent is not visible. Edit Mode positions the bar from `UIParent` anyway and `SetParent` leaves anchors intact, so it still lands where the user had it. The un-park is gated on EUI having parked it — still in our hidden parent, or in the captured parent while that parent is hidden. A bar parked under some other frame belongs to another addon and is left alone, which is the case #991 protects.

## 2. The re-arm never re-armed anything

Logging in with the cast bar already off produced a different failure: events unregistered, frame in its normal container. The re-arm called `SetUnit` with the unit the frame already held, and `CastingBarMixin:SetUnit` guards its entire body on `self.unit ~= unit`, so nothing was registered.

Forcing that guard open is not an option — `SetUnit` runs `StopAnims` → `StopFinishAnims`, which iterates a table that cannot be accessed while tainted:

```
attempted to iterate a table that cannot be accessed while tainted (execution tainted by 'EllesmereUI')
[Blizzard_UIPanels_Game/Mainline/CastingBarFrame.lua]:722: in function 'StopFinishAnims'
[Blizzard_UIPanels_Game/Mainline/CastingBarFrame.lua]:119: in function 'SetUnit'
```

That guard is the only reason the original no-op call ever looked harmless. The events are now registered directly, mirroring the set oUF's Castbar element restores when it disables — the one path that has ever actually re-armed these frames, and one that runs no Blizzard code. The pet bar is covered too; oUF silences it alongside the player's.

`RestoreBlizzCastBarEvents` also treated "EUI does not own this" as "another addon owns this" and re-silenced the frame, so the release re-registered the events on the way past and the restore tore them straight back down. A foreign claim is now recorded explicitly when an `UnregisterAllEvents` lands outside a capture window, and only that re-silences. Registered-but-unclaimed means EUI's own bookkeeping lost track, and is left armed: a visible Blizzard cast bar is something the user can switch off, a dead one is not.

`DisableBlizzard("player")` on the hidden player-frame source is now wrapped in a capture window too — it unregisters PlayerFrame's cast bar child, so without the window EUI recorded its own silence as a third-party claim.

Adds an opt-in trace behind `EllesmereUIDB._CBDEBUG`, persisted so it can be armed before a relog, which is when these decisions happen.

## Testing

- Standalone cast bar addon enabled → it keeps the frame, Blizzard's does not reappear beside it (#991's case still holds).
- EUI's cast bar turned off mid-session → Blizzard's cast bar appears and casts normally.
- Logged in with EUI's cast bar already off → Blizzard's cast bar works.
- EUI's cast bar re-enabled → Blizzard's goes away again, no doubling.

Supersedes #999, which carried the second half on its own.

### Required Giphy
Greatest Of All Time Goat GIF by Nutrena Feed

### DIFF
--- a/EllesmereUI.lua
+++ b/EllesmereUI.lua
@@ -12549,13 +12549,27 @@ function EllesmereUI._BlizzCastBars()
     for i = 1, #bars do
         local bar = bars[i]
         -- An UnregisterAllEvents outside one of our own Capture/Restore windows
-        -- is another addon claiming the frame, so EUI drops its claim and stops
-        -- treating that silence as something it may undo.
+        -- is another addon claiming the frame: EUI drops its claim and records
+        -- the claim as foreign. "Foreign" is tracked separately from "not ours"
+        -- because those are very different states -- see RestoreBlizzCastBarEvents.
         hooksecurefunc(bar.frame, "UnregisterAllEvents", function()
-            if not EllesmereUI._blizzCastBarSnapshot then bar.owned = false end
+            if not EllesmereUI._blizzCastBarSnapshot then
+                bar.owned = false
+                bar.foreign = true
+                EllesmereUI._CastBarDbg("foreign claim", bar.unit)
+            end
         end)
     end
     return bars
+end
+
+-- Opt-in trace for cast bar ownership. Persisted in EllesmereUIDB so it can be
+-- armed before a relog, which is when the interesting decisions happen:
+--   /run EllesmereUIDB._CBDEBUG = true
+function EllesmereUI._CastBarDbg(...)
+    if EllesmereUIDB and EllesmereUIDB._CBDEBUG then
+        print("|cff33ff99[CastBar]|r", ...)
+    end
 end
 
 function EllesmereUI.CaptureBlizzCastBarEvents()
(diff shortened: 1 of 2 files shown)


dfrisone:
Re-verified in-game on the combined branch (both commits together, not just each half on its own): third-party cast bar keeps the frame with no Blizzard bar beside it; EUI's cast bar turned off mid-session brings Blizzard's back; logging in with it already off works; re-enabling hides Blizzard's again with no doubling.

PR #999, opened 2026-07-26
Re-arm Blizzard's cast bars, and stop killing them on release

dfrisone:
Logging in with EUI's cast bar already off left Blizzard's cast bar dead: events unregistered, frame sitting in its normal container, nothing able to bring it back short of a reload. Two causes, both pre-dating #991.

## The re-arm never re-armed anything

It called `SetUnit` with the unit the frame already held, and `CastingBarMixin:SetUnit` guards its entire body on `self.unit ~= unit`, so nothing was registered.

Forcing that guard open is not an option either — `SetUnit` runs `StopAnims` → `StopFinishAnims`, which iterates a table that cannot be accessed while tainted, so from addon execution it throws:

```
attempted to iterate a table that cannot be accessed while tainted (execution tainted by 'EllesmereUI')
[Blizzard_UIPanels_Game/Mainline/CastingBarFrame.lua]:722: in function 'StopFinishAnims'
[Blizzard_UIPanels_Game/Mainline/CastingBarFrame.lua]:119: in function 'SetUnit'
```

That guard is the only reason the original no-op call ever looked harmless.

The events are now registered directly, mirroring the set oUF's Castbar element restores when it disables — the one path that has ever actually re-armed these frames, and one that runs no Blizzard code. The pet bar is covered too: oUF silences it alongside the player's, and only that same `Disable` ever brought it back.

## Restore treated "not ours" as "someone else's"

`RestoreBlizzCastBarEvents` re-silenced any frame it did not own, which is what killed the bar here — the release re-registered the events on the way past and the restore tore them straight back down.

A foreign claim is now recorded explicitly when an `UnregisterAllEvents` lands outside a capture window, and only that re-silences. Registered-but-unclaimed means EUI's own bookkeeping lost track, and is left armed: a visible Blizzard cast bar is something the user can switch off, a dead one is not.

The `DisableBlizzard("player")` call on the hidden player-frame source is now wrapped in a capture window too. It unregisters PlayerFrame's cast bar child, so without the window EUI recorded its own silence as a third-party claim and would never hand the bar back.

Adds an opt-in trace behind `EllesmereUIDB._CBDEBUG`, persisted so it can be armed before a relog, which is when these decisions happen.

## Testing

Verified in-game, both directions:

- Standalone cast bar addon enabled → it keeps the frame, Blizzard's bar does not reappear next to it (#991's case still holds).
- EUI's cast bar disabled → Blizzard's cast bar comes back and casts normally.

## Note on #998

Independent of #998, but touches the same function. #998 fixes the release re-parenting into a hidden `PlayerFrame`; this fixes the event side. Both are needed for "release actually restores the bar" to be true. Whichever merges second may want a trivial rebase, since the changed hunks are adjacent.

### DIFF
--- a/EllesmereUI.lua
+++ b/EllesmereUI.lua
@@ -12549,13 +12549,27 @@ function EllesmereUI._BlizzCastBars()
     for i = 1, #bars do
         local bar = bars[i]
         -- An UnregisterAllEvents outside one of our own Capture/Restore windows
-        -- is another addon claiming the frame, so EUI drops its claim and stops
-        -- treating that silence as something it may undo.
+        -- is another addon claiming the frame: EUI drops its claim and records
+        -- the claim as foreign. "Foreign" is tracked separately from "not ours"
+        -- because those are very different states -- see RestoreBlizzCastBarEvents.
         hooksecurefunc(bar.frame, "UnregisterAllEvents", function()
-            if not EllesmereUI._blizzCastBarSnapshot then bar.owned = false end
+            if not EllesmereUI._blizzCastBarSnapshot then
+                bar.owned = false
+                bar.foreign = true
+                EllesmereUI._CastBarDbg("foreign claim", bar.unit)
+            end
         end)
     end
     return bars
+end
+
+-- Opt-in trace for cast bar ownership. Persisted in EllesmereUIDB so it can be
+-- armed before a relog, which is when the interesting decisions happen:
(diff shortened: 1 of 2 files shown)


dfrisone:
Folded into #998 so the cast bar release ships as one tested change. The event-side fix here only makes the behaviour correct together with #998's parent fix, and each was verified without the other -- merging them separately would have produced a combination nobody had run. Same commit, cherry-picked onto #998.

PR #1003, opened 2026-07-26
Fix Blizzard tracked buff bars reappearing over Tracked Buff Bars

dfrisone:
### Problem

Reported by @Fexe (8.5.8, Cooldown Manager): EUI Tracked Buff Bars work correctly, but a duplicate Blizzard tracking bar draws at the same time. Intermittent, mostly in combat, both in and out of instances, and not reproducible on demand.

### Cause

Tracked Buff Bars do not replace Blizzard's bars, they **mirror** them: `UpdateTrackedBuffBarTimers` reads min/max/value straight off Blizzard's Bar frames. So `BuffBarCooldownViewer` can never be `Hide()`n, a hidden viewer stops updating its children. `HideBlizzardCDM` instead set the viewer's alpha to 0 and parked it at -10000, 10000, once, with nothing re-asserting either half.

Both halves come undone in normal play:

- Blizzard's hide-when-inactive fade animates the viewer's alpha back to 1 whenever a tracked buff goes active. Animations bypass `SetAlpha`, so no hook can observe it. This is the same lesson already recorded in `CollectAndReanchor` for unclaimed CD/utility pool frames.
- Anything that re-applies the Edit Mode layout (a layout apply, `SaveLayouts`, a zone-in) re-anchors the viewer back to its saved position, undoing the park.

Once both have happened, the next buff proc in combat draws Blizzard's bars on top of ours. The tell is an asymmetry that was already in the file: unclaimed CD/utility pool frames get alpha + park **plus a `SetPoint` re-park hook**, while the bar viewer got the park with no hook.

### Fix

Route the secondary viewer's suppression through `ns.ParkSecondaryBuffViewer`, which re-asserts alpha and park and installs a `SetPoint` hook that re-parks whenever anything moves the viewer, the same protection the unclaimed pool frames already have.

The re-park is deferred by one frame on purpose: Blizzard re-anchors this viewer from inside its Edit Mode layout pass, which goes on to move protected systems, so re-parking inline would carry taint into the rest of that pass. The delay also coalesces `ClearAllPoints` + `SetPoint` bursts into a single park. A park blocked by combat lockdown is flushed on `PLAYER_REGEN_ENABLED` rather than dropped.

Two restore-path defects are fixed alongside, both of which a durable park would otherwise have made permanent:

- `RestoreBlizzardCDM` never restored the secondary viewer's alpha, so disabling the CDM module left Blizzard's tracked bars invisible.
- It restored `EnableMouse(true)` on a viewer whose Blizzard default is mouse-disabled, leaving an invisible click-catcher. `RestoreBlizzardBuffFrame` already guarded against exactly this.

### Diagnostics

Adds `/cdmbb`, which dumps the viewer's suppression state, alpha and anchor. While Tracked Buff Bars own the display it should read `suppressed=true`, `alpha=0.00` and a point at `(-10000, 10000)`.

### Testing

- `luac -p` clean.
- In-game: ran a normal dungeon with Tracked Buff Bars enabled, no duplicate Blizzard bar appeared.
- `/cdmbb` confirms the viewer stays parked and at alpha 0.

### Required Giphy
Funny Face Dog GIF

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICooldownManager.lua
+++ b/EllesmereUICooldownManager/EllesmereUICooldownManager.lua
@@ -3069,6 +3069,53 @@ end
 -------------------------------------------------------------------------------
 --  Hide / Restore Blizzard CDM
 -------------------------------------------------------------------------------
+
+-- Suppress the secondary buff-bar viewer (BuffBarCooldownViewer).
+--
+-- It is parked offscreen rather than Hidden because TBB mirrors min/max/value
+-- straight off Blizzard's Bar frames -- a hidden viewer stops updating them.
+-- The park, not the alpha, is what actually holds: Blizzard's hide-when-
+-- inactive fade animates the viewer's alpha back to 1 whenever a tracked buff
+-- goes active, through a path no hook can see (the same lesson as the
+-- unclaimed-frame park in CollectAndReanchor). So once anything re-anchors the
+-- viewer back to its Edit Mode position -- a layout apply, SaveLayouts, a
+-- zone-in -- the next buff proc in combat draws Blizzard's bars on top of ours.
+--
+-- The SetPoint hook makes the park self-healing, mirroring the one the
+-- unclaimed CD/utility pool frames already get.
+function ns.ParkSecondaryBuffViewer(frame)
+    if not frame then return end
+    local fc = FC(frame)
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
that gif LMFAOOOOO

Issue #717, opened 2026-07-13
[BUG] Micro Menu vanish


TryAndErrorBen:
every 2nd pet battle my micromenu is vanishing

PR #1009, opened 2026-07-26
refresh to show micro_menu after pet_battle

bjyitu:
fixed micro_menu hidden after pet_battle


EllesmereGaming:
Thanks! Looks like a bug, but i'm not sure this actually fixes it, was this tested in game? is it maybe missing part of the fix in the PR? The hider registers a constant state, so `_onstate-vis` fires once and never again. When the pet battle UI hides the micro menu, nothing re-shows it. Pushing the driver again is the right idea, but `RefreshMicroMenuHider` bails early when nothing changed (if want == mmLastApplied then return end). 'hide' comes purely from the profile, which a pet battle doesn't change so this should return before re-registering anything, and I can't see how this fix actually works.

Also, is it Blizzard's micro menu that goes missing, or the EUI micromenu block on your data bar?

bjyitu:
After adding this modification, I no longer lose Blizzard's micro menu after entering and exiting pet battles. I'm not sure if it was because of reloading the interface multiple times that it succeeded.

In addition, I also found other users submitted similar questions in Issues. https://github.com/EllesmereGaming/EllesmereUI/issues/717#issue-4877930532

Issues screenshot:

Before entering the pet battle:
before

After exiting pet battle:
after

EllesmereGaming:
ok fix in with next patch ty!

PR #1025, opened 2026-07-27
fix(cdm): match active-state rules on trinkets after a logout

dfrisone:
## Problem

A custom Active State set on a trinket stopped working after a logout, and stayed dead
until the setting was opened again. Reported as a setting that would not persist, which
is how it was chased for a long time -- but the stored value was never lost.

## Cause

An icon can be named by two different tokens:

- `-itemID` in the settings store. `ResolveCustomActiveKey` maps a slot frame to the
  equipped item so each item tracks its own settings.
- `-13` / `-14` on the trinket slot frame itself.

Which one a RULE ends up with depends on whether `GetInventoryItemID` was readable when
rules were armed. **At login it is not.** The equipped-trinket skip in `FakeActive_Rearm`
never fires, so the rule is keyed `-itemID`, while the frame that renders moments later
carries `-13`. `ApplyRule` compares `fc.spellID` by raw equality, so nothing ever matched.

Nothing reconciled the two afterwards, so the rule sat armed against an icon that did not
exist for the rest of the session. Any settings touch forced a rebuild, which re-armed
with equipment present, produced the slot key, and it started working again -- hence
"I have to choose it again".

Captured live on an affected client:

```
rule3 key=-249339 src=-249339 dur=20 trigger=cast casts=[1260633] frames=0 <- no icon
```

The trigger resolved correctly. The rule simply addressed a key no icon carried. `src`
equal to the item key confirms it came from the main loop rather than the slot pass,
which only happens when the equipment lookup returned nil.

## Fix

Compare through a slot-key -> equipped-item-key map instead of raw equality. Resolving at
MATCH time rather than arm time fixes every ordering variant at once, because frames
render long after equipment is readable.

The map is a plain table so the per-frame loops pay a lookup rather than an API call per
frame per rule. It is refreshed on equipment change, at re-arm, and lazily while empty
(which covers the login window).

## Known narrow edge

A profile carrying BOTH a slot-level Apply-to-Bar entry and a per-item entry can arm two
rules during the login window that now resolve to the same frame. The overlay is per
frame, so the later rule wins rather than stacking, and the two entries chain to
near-identical settings. Only a configured CD-ready sound could double there, since the
rules arm under separate `srcKeys`.

## Scope

Item-keyed active states only: trinkets, potions, custom items. Spell keys are positive
and were never exposed to this.

## Testing

Verified in game on an affected character, on a full logout (not a reload), without
opening any settings first:

- Before: `rule9 key=-251787 ... frames=0 <- no icon`, no glow.
- After: `rule9 key=-251787 ... frames=1`, glow runs for its configured 20s.

The second commit is the developer-only diagnostic that identified this; it extends the
existing `EllesmereUI.FakeActiveDebug()` and adds no user-facing surface.

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICdmFakeActive.lua
+++ b/EllesmereUICooldownManager/EllesmereUICdmFakeActive.lua
@@ -1427,5 +1427,35 @@ if EllesmereUI then
         ns._fakeActiveDebug = not ns._fakeActiveDebug
         print(("|cff0cd29fEUI FakeActive|r debug %s | armed=%s rules=%d"):format(
             ns._fakeActiveDebug and "ON" or "off", tostring(_armed), #_rules))
+
+        -- Per-rule dump. "armed with N rules" reads true in every broken report,
+        -- so on its own it separates nothing. What matters per rule is whether
+        -- its trigger resolved to a cast spell at all, and whether any live icon
+        -- carries its key -- the two independent ways a rule goes inert.
+        local FCt, icons = ns._ecmeFC, ns.cdmBarIcons
+        for i = 1, #_rules do
+            local rule = _rules[i]
+            local triggers = {}
+            for sp, list in pairs(_castMap) do
+                for j = 1, #list do
+                    if list[j] == rule then triggers[#triggers + 1] = sp end
+                end
+            end
+            local matched = 0
+            if FCt and icons then
+                for _, list in pairs(icons) do
+                    for j = 1, #list do
+                        local f = list[j]
+                        local fc = f and FCt[f]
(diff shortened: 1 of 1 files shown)


dfrisone:
Re-verified in game on the exact commit this PR ships (`f7149b87`), on a full logout with no settings opened first.

```
rule9 key=-251787 src=-251787 dur=20 trigger=cast casts=[1253115] frames=1
EUI FakeActive ON  sid=-251787
```

The rule now matches its icon (`frames=1`, previously `frames=0`) and the active window opens and applies on use.

`rule3 key=-249339 ... frames=0` in the same capture is a second trinket that is not currently equipped, so no icon exists for it to match. That is the intended dormant state, not a regression.

PR #1030, opened 2026-07-27
fix(qol): use guild funds that cover only part of a repair bill

LoChinAn:
## The bug

Auto-repair handed guild funds to `RepairAllItems` only when the guild bank could
cover the **entire** bill:

```lua
and cost <= GetGuildBankWithdrawMoney()
```

When the remaining guild funds fell short of the total, the addon silently gave up
on them and charged the player the whole amount.

Reproduced in game: a **411g 69s 39c** bill with **185g 13c** of guild funds left.
Blizzard's own guild repair button offers to split it — guild pays 185g 13c, the
player pays 226g 56s 39c. EUI's auto-repair charged the player all 411g 69s 39c.
The 185g of guild funds went unused.

## Root cause

The server already splits a repair. `RepairAllItems(true)` never fails for want of
guild funds: it draws whatever the guild's remaining funds hold and charges the
player the rest. Blizzard's guild repair button does nothing more than this —
from `MerchantFrame.xml`, `MerchantGuildBankRepairButton`:

```lua
if(CanGuildBankRepair()) then
  RepairAllItems(1);
  PlaySound("ITEM_REPAIR");
end
```

The affordability gate pre-empted that split and discarded it.

## Changes

**1. Drop the gate.** Guild funds are now used whenever the option is on, the
player is in a guild, and `CanGuildBankRepair()` passes — matching Blizzard's own
button. The split is left to the server.

**2. Report the payer that actually paid.** The `(guild bank)` suffix was
previously derived from the pre-flight guess, so it appeared whenever guild funds
were merely *attempted*. A repair the player funded in full was still labelled as
guild funded. The share is now deduced from the player's own ledger: money that
leaves the player's pocket over a short `PLAYER_MONEY` window is the player's
share, and the remainder of the bill is the guild's.

**3. Show the split.** A split bill prints the guild's portion after the suffix.

## Message format

| Case | Output |
| --- | --- |
| Guild funds cover it all | `Repaired all items for 15g (guild bank)` |
| Player pays it all | `Repaired all items for 223g 42s 45c` |
| Split | `Repaired all items for 238g 42s 45c (guild bank) 15g` |
| Guild paid part, player cannot cover the rest | the paid amount, then `Not enough gold to repair.` |

The first two are unchanged from current behaviour. **No new locale strings** — the
split reuses `"Repaired all items for %s"` and `" (guild bank)"`, so no locale file
needs updating.

## Why the payer is measured instead of computed

The guild's usable funds are readable — `min(GetGuildBankWithdrawMoney(),
GetGuildBankMoney())`, with `-1` meaning an unlimited rank — which is the
arithmetic behind Blizzard's tooltip. It is not usable as the *source* of the
split, because `GetGuildBankMoney()` reads 0 until the player opens a guild bank
in the current session, which is the common case. Computing the split from it
would report "player paid everything" on most sessions.

It is used instead as a **ceiling** on the deduced share. A stale reading is
either too high to bind or zero and skipped, so it can only ever tighten a wrong
answer, never introduce one.

## Cost

Nothing is registered outside the confirmation window. `EUI_RepairWatcher` is
created on the first guild-funded repair and reused (frames are never collected);
`PLAYER_MONEY` is registered only for the ~0.5s settle window, and every exit path
routes through `StopRepairWatch`, so the event cannot be left registered.

## Testing

- Guild funds cover the full bill → single line with the suffix, player's gold unchanged
- Split bill → both portions matched against the guild repair button's tooltip figures and against the actual gold delta
- No guild funds → full self-payment, no suffix
- Auto-sell junk active with a large junk load → sale proceeds arrive in their own `PLAYER_MONEY`, trailing the repair debit; the split was unaffected
- Watcher lifecycle → frame absent until the first guild repair, `PLAYER_MONEY` unregistered after every resolution

## Known limitations

- Should a junk sale and the repair debit ever share a single `PLAYER_MONEY`, the
  deduced guild share would be overstated. The funds ceiling bounds this, though
  the ceiling is skipped when guild bank data is not loaded. Not observed in
  testing — `SellAllJunkItems` sells one item per round trip, so its credits
  trail the repair debit.
- If the merchant window closes inside the settle window while the repair did not
  go through, the message can overstate what was repaired.

### DIFF
--- a/EllesmereUIQoL/EllesmereUIQoL.lua
+++ b/EllesmereUIQoL/EllesmereUIQoL.lua
@@ -919,10 +919,11 @@ qolFrame:SetScript("OnEvent", function(self)
     --     session, so the figure cannot be trusted on its own -- it serves as a
     --     ceiling on the deduced share, never as the source of it.
     --   * Hence the guild share is deduced from the player's ledger, and only
-    --     outgoing amounts are summed: junk sales from the same MERCHANT_SHOW
-    --     land in this window and a net figure would cancel against them. In
-    --     practice they never share a PLAYER_MONEY -- SellAllJunkItems sells one
-    --     item per round trip, so its credits trail the repair debit.
+    --     outgoing amounts are summed. The junk sweep above keeps firing passes
+    --     across this window, so a net before/after figure would cancel its
+    --     credits against the repair debit and invent a guild share. Credits
+    --     arrive in their own PLAYER_MONEY and are ignored; only a sale landing
+    --     in the same event as the debit could still net off.
     local repairWatcher, repairWatchLast, repairWatchOut
     local repairWatchGen = 0
 
--- a/EllesmereUIQoL/EllesmereUIQoL.lua
+++ b/EllesmereUIQoL/EllesmereUIQoL.lua
@@ -908,6 +908,116 @@ qolFrame:SetScript("OnEvent", function(self)
         end
     end
 
+    -- Auto-repair result watcher. Three things here are not inferable from the
+    -- code:
+    --   * RepairAllItems(true) never fails for want of guild funds -- the server
+    --     pays whatever the guild allowance still covers and silently charges
+    --     the player the rest, so "nothing left to repair" says nothing about
+    --     who paid.
+    --   * The guild's usable funds are readable (GuildRepairFunds below), but
+    --     GetGuildBankMoney() reads 0 until the player opens a guild bank this
+    --     session, so the figure cannot be trusted on its own -- it serves as a
+    --     ceiling on the deduced share, never as the source of it.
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
need a rebase on this please <3

LoChinAn:
Rebased — the conflict was with the Auto Sell Junk rewrite (4d98bafa,
e17218af), whose sweep block sits at the same insertion point as this PR's
repair watcher. Both kept, sweep untouched.

The last commit only refreshes a comment: it claimed junk credits trail the
repair debit, which held when SellAllJunkItems fired once but not with the
new multi-pass sweep. No code change — credits arrive in their own
PLAYER_MONEY and were never counted.

PR #1032, opened 2026-07-27
Add "Show Absorbs as Health" to Raid/Party absorbs

nulltyto:
## What

An optional rendering mode for Raid/Party absorbs: the part of a shield that fills the **empty** health bar is painted in the health bar's own texture and colour, so a shield reads as health until the bar is full. Only the surplus past 100% still renders as an absorb.

The goal is making "in trouble or not" readable at a glance, by removing the health/absorb split as visual noise.

- 80% health + 10% shield → bar reads 90%, all one colour.
- 80% health + 30% shield → bar reads full, with 10% shown as surplus.

## Options

Two toggles in the existing **Absorb Rendering** cog, next to Show Overshield. Both are registered in the `absorbs` section map, so raid and party both get them and per-section party overrides work.

| Setting | Default | Notes |
| --- | --- | --- |
| Absorbs as Health | off | Overlay placement only. The edge modes draw the whole absorb through the backfill and have no missing-health section to paint, so the toggle disables itself there. |
| Health Edge Line | on | Hairline on the true health edge, so the shield-painted section stays tellable apart. Greys out unless the above is on. Self-gates on "has shield" via a 1px bar fed the absorb, so it never reads the secret. |

## How it works

The geometry already existed — the forward absorb bar is clipped to the missing-health region — so this is a paint change rather than new math, and it stays secret-value safe. Three supporting pieces:

**Background.** Anchored to a clamp bar fed the absorb already capped to missing health, so it starts where the *shield* ends rather than where health ends. Otherwise it sits under the shield section and, at any fill alpha below 1 (Dark Mode's is 0.9), makes it read washed out next to real health.

**Overshield.** Right-aligned at the bar's end rather than growing leftward from the health edge, so every health level renders the way 100% health already does. Left as-is, it stranded a stripe mid-bar once the forward shield was painted as health. Position comes from a gate anchored to the backfill fill's left edge and fed the clamped absorb, landing at `maxHealth - overshield` with no arithmetic on secrets.

**Colour.** The bar colour and the fill texture's alpha travel on the same two channels the health bar uses, rather than being multiplied into one — the Classic and Custom Dynamic modes derive the colour from secret health, so combining them throws.

The options preview mirrors all of it.

## Secret-value constraints

Three worth flagging for review, all found the hard way:

- Colours out of `GetStatusBarColor` may be secret (Classic / Custom Dynamic). Pass them to setters; never compute with them.
- `clampBar`, `overGate` and `hairGate` are anchored to fills driven by `SetValue(secret)`, so **their own geometry queries return secrets**. Safe to position, and safe to anchor things to; never safe to read back.
- Their width is the value-to-pixel scale, so they are re-sized every update like the other absorb bars. The raid tier sizes and the party width resize the health bar long after creation, and a stale width there silently collapsed the background's rect.

## Testing

Verified in-game on party frames across all five health colour modes (Class, Custom, Classic, Custom Dynamic, Dark).

https://github.com/user-attachments/assets/632e83d5-b945-4e29-bd03-4a64807edddc

Not yet exercised:

- Blizzard Modern absorb style — a separate branch, covering the seam-spark suppression and the overshield-spark re-anchor.
- Toggling the feature back off while in Dark Mode — the fill-alpha reset path.
- Raid frames with group size changing (`ns._activeSizeW` resizing).

## Note on the locale check

`Locales/_keys.txt` is stale on `main` independently of this branch — regenerating it adds 24 keys from previously merged work, none from here. I left it untouched so this diff stays scoped, which means the locale check will fail until it is regenerated separately. The new strings here are runtime keys (`L(row.label)` and tooltip text), so they do not belong in the static list regardless.

### DIFF
--- a/EllesmereUIRaidFrames/EUI_RaidFrames_Options.lua
+++ b/EllesmereUIRaidFrames/EUI_RaidFrames_Options.lua
@@ -1387,6 +1387,22 @@ initFrame:SetScript("OnEvent", function(self)
                       tooltip="Show the part of an absorb that exceeds your empty health and backfills over your current health. When off, absorbs only fill the empty part of the health bar; on Default Blizz Frames the glow line stays pinned at the right edge.",
                       get=function() return SVal("showOvershield", true) end,
                       set=function(v) SSet("showOvershield", v) end },
+                    { type="toggle", label="Absorbs as Health",
+                      tooltip="Draw the part of a shield that fills the EMPTY health bar in the health bar's own texture and color, so a shield reads as health until the bar is full. Only the overshield past 100% still renders as an absorb. Overlay placement only.",
+                      disabled = function()
+                          local st = SVal("absorbStyle", "none")
+                          return st ~= "blizzardModern" and SVal("absorbEdgeMode", "overlay") ~= "overlay"
+                      end,
+                      disabledTooltip = "Requires Placement to be Overlay",
+                      rawTooltip = true,
+                      get=function() return SVal("absorbAsHealth", false) end,
+                      set=function(v) SSet("absorbAsHealth", v) end },
+                    { type="toggle", label="Health Edge Line",
+                      tooltip="Mark where your real health ends with a hairline, so a shield drawn as health is still tellable apart at a glance. Uses the absorb color; only drawn while a shield is up.",
+                      disabled = function() return not SVal("absorbAsHealth", false) end,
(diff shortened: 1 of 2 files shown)


nulltyto:
This was created after an on-stream comment by jdotb on how he wished you could set up absorbs.

Probably needs further testing other than me just jumping off things and shielding/healing myself.

Feel free to close this w/o adding if deemed not worth the addition.

PR #1041, opened 2026-07-27
Feat/QoL: new Raid Tools page — ready check, pull timer, raid markers

Absol3m:
## What does this PR do?

Adds a **Raid Tools** page to Quality of Life, next to Movement Alerts.
Off by default; nothing changes for anyone who does not switch it on.

Pure API — no Blizzard frame is hooked, reskinned or reparented. The panels
are our own frames drawn with the suite's primitives, so there is no taint
surface from borrowing Blizzard's raid UI.

## What it does

Four **independently positionable** panels, each with its own saved position,
its own scale and its own on/off switch. They stack by default so they read as
one panel until you move something; unlock mode's anchor-linking drags several
as a unit.

| Panel | Contents |
|---|---|
| Group | Ready check, role check, convert to raid/party, disband (confirmed) |
| Pull Timer | Three configurable durations + stop |
| Target Markers | The 8 raid symbols on your target + clear |
| World Markers | The 8 coloured ground flares + clear all |

image

Capture d&#39;écran 2026-07-27 171704

Splitting the categories is the deliberate difference from Method Raid Tool,
which ships one fixed bar: target markers can sit under the raid frames while
the group buttons live elsewhere.

**Access:** `/euiraid` or a keybind. **Options:** QoL → Raid Tools.

Capture d&#39;écran 2026-07-27 174706

## Combat model

Raid markers have to be placeable mid-pull, and that constraint shapes the
feature:

* Every marker button is a `SecureActionButtonTemplate` built **once** at load
  and never re-anchored, re-parented or resized afterwards.
* Panel visibility runs through a secure snippet driven by a state driver, not
  Lua `Show`/`Hide` — the client blocks hiding a frame that owns protected
  children during combat, which is exactly when a raid leader needs the markers.
* Geometry changes are deferred out of combat behind a dirty flag, so a settings
  change made mid-fight lands on `PLAYER_REGEN_ENABLED` instead of being lost.

Marker buttons are dimmed rather than disabled when the player lacks assist —
the game already no-ops the call. Solo counts as permitted, matching MRT, so the
panels are usable on a target dummy.

## MRT compatibility

MRT is the reference implementation, so the parts players already rely on follow
it rather than inventing an alternative:

* **Slash commands come from the `SLASH_*` globals**, never literals. `/tm` and
  `/cwm` are localized — a hardcoded literal silently breaks every non-English
  client.
* `marker` is a **string**, and set/clear split across `action1`/`action2`.
* Clear-all is a macro (`/cwm All`); the attribute form only clears one index.
* World markers are coloured discs using MRT's own `wm_color` values, so both
  addons agree on which flare is which colour.
* The pull timer hands the countdown to whichever boss mod is loaded
  (`SlashCmdList.pull` → `BIGWIGSPULL` → `DEADLYBOSSMODSPULL`) before running
  Blizzard's own, guarded by `InChatMessagingLockdown` — the same sequence as
  MRT's `Timers.lua`. Pressing our button and MRT's produces the same pull.

No addon comms are used, so there is no interaction with MRT's `EXRTADD` /
`MRTADDA`–`I` prefixes.

## Fits the existing suite

Visibility is the shared multi-select system, not a local rule:
`BuildVisibilityDriverString` compiles the secure driver, `EvalVisibilityModes`
settles the out-of-combat case a re-registered driver will not re-fire, and the
options row is `BuildVisibilityModeRow`. Mouseover is filtered out via the
existing `caps.noMouseover` — hover reveal needs the Lua poll the insecure
modules run, and an inert checkbox is worse than none.

Visuals go through `MakeStyledButton`, `MakeBorder`, `GetFontPath` and
`GetAccentColor`; confirmations through `ShowConfirmPopup`; messages through
`EllesmereUI.Print`. The settings are one slice of the shared `EllesmereUIQoLDB`
profile, read through the same per-call accessor BattleRes and MovementAlert use.

## How was it tested?

Tested solo and in a party: panel display, positioning, per-panel scale and
enable, target markers, the visibility axes, the options page.

## Follow-ups

Deliberately out of scope, each planned as its own PR chained on this one:

1. An addon-comms layer. It is the prerequisite for the durability and weapon
   enchant columns of a raid check — neither is readable on other players, so
   each client has to broadcast its own, which is how MRT does it — and later
   for shared raid notes and addon version checks.
2. A raid check window (roster grid: flask, food, rune, raid buffs), on top of
   that layer.

## Checklist


Absol3m:
This is the first step in developing the raid tool; I’m working on making it interoperable with MRT.
If you spot any features you’d like to see included, please feel free to let me know in the comments on this PR – I always read everything, even after it’s been merged.

hellsy55:
This looks great! Can we have an option to show or hide especific EUI raid groups just like the blizzard frame has?

Absol3m:
> This looks great! Can we have an option to show or hide especific EUI raid groups just like the blizzard frame has?

I’ll look into this with a view to future updates, if this part is approved by the team.

PR #1071, opened 2026-07-29
Add translation for 'Change Spec' in Korean

Crazyyoungs:
## What does this PR do?

- Players using the Korean client will now see `"특성 변경"` properly rendered in the UI.

## How was it tested?

- Tested on the live retail client with Korean language settings.

## Checklist


EllesmereGaming:
manually merged, in with next patch ty!

PR #1077, opened 2026-07-29
feat(item level): Add an option to color item level by track or not

Xelia-wow:
## What does this PR do?

Adds a "Color by Rarity" toggle for item level text, both in the Bags/Bankand on the Character/Inspect sheet. Item levels are normally colored by the item's upgrade track (Explorer/Veteran/Champion/Hero/Myth); with this option enabled, they're always colored by item rarity instead, ignoring the track.

- Bags: new toggle in the "Item Level Options" cog next to "Show Item Level" (applies to both bags and bank).
- Character Sheet / Inspect: new toggle in the cog next to the "Item Level" checkbox on the Character Sheet card.
- Both toggles default **off** (unchanged track-coloring behavior).
- Added French, German, Korean, Russian, Simplified and Traditional Chinese translations for the new label/tooltip.
- Included two one-time migrations to clear a pre-existing, never-shipped "custom item level color" flag (`itemlevelUseCustomColor` /  `charSheetItemLevelUseColor`) that could get stuck ON during dev testing and silently outrank both the track color and the new toggle.

## How was it tested?

Open bag/bank/character sheet with and without the option checked

## Screenshots

image
image

## Checklist

- [x] Zero cost while disabled: no events registered, no polling, no hooks  doing work, no frames built -- just one extra boolean check inside existing color-resolution code that already runs every refresh
- [x] Cheap while enabled: same code path, no polling/timers/per-frame allocations added
- [x] No writes onto Blizzard-owned frames -- only touches EUI's own bag/bank buttons and the existing themed character-sheet label frames, via the  same accessors the surrounding code already uses
- [X] Tested in-game, works on live retail; no load errors on the 12.1 PTR client


EllesmereGaming:

PR #1081, opened 2026-07-29
QoL: Raid Tools gains a subgroup visibility row

Absol3m:
Blizzard's raid frames used to expose a per-subgroup filter and the UI for it
is gone — the functions linger on `CompactRaidFrameManager` but nothing
exercises them, and it is protected besides. So a raid leader who wants to
watch only groups 1–4 has nowhere to say so.

This puts the row of eight numerals back, in Group & Pull.

### A row, not a panel

This PR previously proposed a third Raid Tools window and converted `showAs`
into a cumulative set of windows plus a combine flag. That is withdrawn.

Show As owns the window composition outright and does so deliberately — the
file says as much in its header. A fourth axis for one strip of buttons would
be paying for a whole window's worth of state (a shell, a saved position, an
unlock element, a title, a migration) to draw eight numerals. The row instead
rides whatever composition the user already chose, and **`showAs` is not
touched at all**.

The result is one file, +154/−1, and no migration.

image

### One setting, one owner

The toggles drive `EllesmereUIRaidFrames`' own `visibleGroups` directly rather
than keeping a copy, and they read it the same way the raid frames do — absent
means unfiltered. That addon's default is groups 1–6, which the row shows as-is;
a second default here is exactly the drift that reading theirs avoids.

With that addon disabled there is nothing to drive, so the row dims and its
label says why instead of lying. That is the whole fallback for someone running
the Blizzard raid frames or another addon's — driving Blizzard's own filter was
considered and rejected, since it is protected and there is no supported read
to paint from. Anyone who replaced our raid frames is using something that
carries its own group filter anyway.

### Rank does not gate it

These toggles change only what *this* client draws, so a raider with no assist
can still filter their own view. `MakeGroupButton` registers every button into
the permission sweep automatically, so the toggles carry a `localOnly` flag and
that sweep skips them — otherwise the two would fight over the same alpha, last
writer winning, and the row would go dead for anyone without assist.

### Combat

Toggling under lockdown sets the value and defers the header rebuild to
`PLAYER_REGEN_ENABLED`. Rebuilding secure group headers is forbidden in combat:
`_LayoutGroupsImpl` returns early there, and the raid frames' own post-combat
pass only re-lays out for roster and size-tier changes, so it would never pick
this up on its own.

The replay sits before the mode gate in the event handler, for the same reason
`applyPending` does — the setting has already changed, and swallowing the
re-render leaves the frames showing the old set.

### Cost

The repaint is memoized on a bitmask of everything it draws — the eight toggles
plus whether there is anything to drive — because it runs on
`GROUP_ROSTER_UPDATE`, which bursts through a raid night, and nothing it reads
changes on that event.

### Notes

- No new locale keys in `_keys.txt`: the two labels reach `L` as variables, the
  documented arrangement this file already uses for its section and marker-row
  labels. The in-game harvester picks them up.
- Rebased on v8.8.8. Merges cleanly, and no longer collides with the Raid Tools
  cog PR in `BuildAll()` — that conflict was a consequence of the withdrawn
  approach.

Tested in game.

### DIFF
--- a/EllesmereUIQoL/EllesmereUIQoL_RaidTools.lua
+++ b/EllesmereUIQoL/EllesmereUIQoL_RaidTools.lua
@@ -107,6 +107,7 @@ local ROW_GAP      = 4
 local MARKER_SZ    = 23
 local MARKER_LBL_H = 10
 local ICON_SZ      = 30
+local RAID_GROUPS  = 8               -- subgroups a raid can hold; fixed by the game
 local PULL_SLOTS   = 3
 local PULL_DEFAULTS = { 3, 5, 10 }   -- also seeded into DB_DEFAULTS below
 ns.PULL_DEFAULTS = PULL_DEFAULTS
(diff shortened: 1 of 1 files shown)


EllesmereGaming:

PR #1108, opened 2026-07-31
feat/QoL: arrange raid groups from a Raid Tools cog

Absol3m:
Blizzard's raid frames let a leader drag a member between subgroups. The suite
hides Blizzard's raid manager and replaces its frames, so it currently takes
that away — **this gives it back**. A cog on the Group & Pull title band opens a
window of eight groups and five slots each, and a member is moved by dragging it
onto another slot.

### Nothing here is secure, and that is not an oversight

`SetRaidSubgroup` and `SwapRaidSubgroup` carry no lockdown restriction, so
unlike the Raid Tools panels this needs no secure frames, no state driver and no
combat deferral. The cog is a plain Button on the shell — scripts on a protected
frame are unrestricted, only its attributes are combat-sensitive.

### Move versus swap is the game's rule

A subgroup holds five and the client will not make room. So dropping onto a
**player** swaps the two, and dropping onto an **empty slot** moves. One rule
covers both cases without asking the user to know which is which — a full group
simply has no empty slot to land on.

### Drag affordances

The origin fades because the member has been lifted out of it, the cursor
carries the name, every legal destination takes a faint wash, and the slot that
would receive the drop takes the accent frame and wash together. The drop then
uses exactly the slot that was outlined, so what the user saw is what happens.

The target is resolved from the drag's own `OnUpdate`: a frame under the cursor
does not fire `OnEnter` while a drag is in flight.

### Permission

The cog exists only for a raid leader or assistant — everyone else already sees
the composition on their raid frames, so an inert control would be clutter. It
rides `RefreshPermissions`, which already memoizes on exactly the rank inputs it
depends on, so it appears and vanishes as rank moves rather than waiting for a
settings pass.

### Scale and position

Both are the window's own, in its own profile slice. It stays out of
`EllesmereUI._popupFrames` deliberately: that list is panel-family membership,
walked by `IsPanelFamilyAnchor` to decide whether a tooltip rides the panel
scale, and a world window opened from a raid panel is not a member. It takes the
same notification through `_onScaleChanged` and folds the panel scale and its own
multiplier together.

### Note

`Locales/_keys.txt` also picks up ten keys from the Friends group feature that
were already missing on main — the gate regenerates the whole file, so any PR
touching a `.lua` carries them.


EllesmereGaming:

PR #1115, opened 2026-08-01
Minimap map icon size (blips) slider + free move magnet snapping

labrie75:
## What does this PR do?

Adds three minimap features plus a profile-switch fix:

**1. Map Icon Size (Blips) slider** (Display > Minimap, below Size; 0.5-3.0, default 1.0)

Tracking blips (herbs, ore, quest dots) are engine-drawn textures that cannot be resized individually, so at large map sizes they get hard to read. This scales the whole Minimap while dividing its frame size, keeping the on-screen map at exactly `mapSize` -- only the blips change size. Everything else is compensated:

- Clock / location / coords / FPS keep their own configured scales.
- Indicator buttons, button rows, addon buttons, and the Omnium Folio keep their sizes (addon buttons restore their own scale when grouped into the flyout).
- The map's screen position is preserved when the scale changes mid-session.
- While dragging the slider, only the map preview updates; the full relayout runs once after the drag settles, so surrounding icons never shimmer or flash-resize.

Also raises the map Size slider cap from 600 to 1000 -- on ultrawide monitors there is room to run the minimap much larger than 600, and blips were the readability blocker at large sizes.

**2. Free Move magnet snapping**

While Shift-dragging a button (Free Move Buttons), it now snaps flush to nearby free-moved buttons (stack above/below/beside, plus edge/center alignment) and to the minimap border -- inside or outside face on square shapes; on circle shapes the button snaps onto the rim along its current angle. Hold Ctrl together with Shift while dragging to disable snapping for pixel-precise placement. The Free Move Buttons tooltip is updated to mention this.

**3. Circle-shape polish**: with a circle map shape --

- the button row runs along the rim instead of a straight line (start corner and direction still follow Button Row Position; the free-move magnet also snaps onto the rim);
- the black button backgrounds render round, drawn slightly larger than the button so corners never poke out;
- square addon-button icons get a round mask so they read as circular icons (EllesmereUI's own atlas icons already have transparent art and are left alone; the mask is dropped when a button returns to the flyout, where icons render square).

Everything follows the Shape setting automatically; square shapes are unchanged. Note that switching between square and circle re-lays out the button row, so buttons the user had free-moved to custom positions may need repositioning after a shape switch. Corner-pinned buttons (Omnium Folio, corner Mail) are left as-is since they have their own position offsets.

At the default blip scale (1.0) and without dragging, every code path computes exactly what it did before -- no change for existing users.

**4. Profile switching re-applies minimap settings**: `RefreshAllAddons` only called the minimap's flyout refresher, so switching profiles (or presets) left the minimap on the outgoing profile's size/shape until a reload -- the new per-profile blip scale hit the same gap. Added the module's own `_EMM_ApplyMinimap` bridge call next to the existing flyout one, matching how the other modules are refreshed there (`_ERB_Apply`, `_EDB_Apply`, ...). One line in `EllesmereUI_Profiles.lua`.

Includes the koKR locale entries for the new option, its tooltip, and the updated Free Move tooltip. No `_keys.txt` update is needed -- this PR adds no explicit `L()` literals (the new option label/tooltip localize through the widget auto-translation path like the existing minimap options).

## How was it tested?

Tested in game on live retail: blip scales 0.5-3.0 at map sizes 140-1000, square and circle shapes, with clock/location/coords/FPS enabled, addon buttons grouped and ungrouped, Omnium Folio visible, and profile swaps. Verified while gathering that herb/ore nodes and other tracked blips stay at their correct map positions after scaling -- only their size changes. Profile switching verified with different blip scales and map sizes saved per profile (and via preset/profile keybind swaps): the minimap now follows the incoming profile immediately instead of needing a reload. Magnet verified against buttons and all four map edges (both faces), with the Ctrl+Shift bypass. Files parse clean, no load errors.

## Screenshots

image

Same map size, blip scale 1.0 / 1.5 / 2.0 -- the map footprint and all UI elements stay identical; only the blips grow:

blip

Circle shape: the button row follows the rim, and backgrounds/icons render round automatically:

blip2

## Checklist

- [x] New settings default **OFF** -- blipScale defaults to 1.0, where every code path computes exactly what it did before; magnet only acts during an explicit Shift-drag, Ctrl+Shift bypasses it
- [x] Zero cost while disabled -- no events, tickers, or OnUpdate added; all code runs inside ApplyMinimap, the options slider, or an active Shift-drag
- [x] Cheap while enabled -- one child-scale pass per settings apply; the per-frame magnet check allocates nothing (no closures/tables per tick), and the slider's drag preview only touches the minimap and its direct children
- [x] No writes onto Blizzard-owned frames -- scale/anchor calls only (SetScale/SetSize/SetPoint); per-child state lives in the module's frame-field data (weak tables)
- [x] Tested in-game, works on live retail; no load errors (not separately run on 12.1 PTR)

P.S. The motivation: while herb/ore farming I run a large minimap, but the tracking blips stay tiny regardless of map size, so nodes are easy to miss at a glance. This slider lets the blips grow to match how large you run the map.


EllesmereGaming:

EllesmereGaming:
actually looks like this one just needs a rebase and I can take a look if you want to do that

labrie75:
> actually looks like this one just needs a rebase and I can take a look if you want to do that

Yes. I had forgotten about it too, so I’d appreciate it if you could take a look.
I just made it as a separate addon for personal use and have been using it, but there are a few issues.

PR #1119, opened 2026-08-01
QoL: move and scale the Loss of Control frame in Unlock Mode

labrie75:
## What does this PR do?

Adds a Loss of Control mover to Unlock Mode, so the alert can be placed and sized like the other Blizzard windows Shifter already handles.

It reuses the existing hidden-window proxy path (`LOOT_WINDOWS`), with two small additions:

- `alwaysOn` -- the entry is not part of the Loot opt-in group. Whether the frame ever appears is Blizzard's own "Loss of Control Alerts" combat option, so a Shifter toggle of its own would only duplicate a Blizzard setting. The mover is simply always offered, and the label reads `Loss of Control (on)` / `(off)` from the `lossOfControl` CVar so it is obvious when the alert itself is switched off.
- `scalable` -- the frame is one icon + timer + label block, so resizing it by width or height would distort it. Both inputs drive the same SCALE, keeping the aspect ratio; the anchor offsets are divided back out so the frame lands exactly where the mover showed it. The mover box tracks the live frame's real size, floored to a size that fits its label.

Everything else follows the existing loot-window behaviour: hidden rect-only proxy, `ignoreFramePositionManager`, SetPoint/OnShow re-application, no mouse state touched, protected/forbidden frames skipped. Position and scale are stored where Shifter already keeps its window positions (account-wide in `EllesmereUIDB`, not per profile) -- deliberately the same as every other Shifter window rather than a one-off exception. Existing entries are unaffected: `alwaysOn` and `scalable` are nil for them, so their code paths compute exactly what they did before.

Includes the koKR locale entries for the new label.

## How was it tested?

Tested in game on live retail: dragged and scaled from Unlock Mode, verified the live alert appears at the saved position and scale (`/reload` and fresh login), with Blizzard's Loss of Control Alerts option both on and off, and with the Loot unlock toggle both on and off (the entry is independent of it).

Known limitation: the `(on)`/`(off)` label is resolved when Unlock Mode builds the mover, so flipping the Blizzard option mid-session needs a reload to refresh the text. Updating a live mover's label would need access to the unlock module's internal mover table, so I left it alone rather than widen the API for a cosmetic refresh -- happy to add that if you'd prefer it live.

## Screenshots

Unlock Mode mover at a small and a large scale (the frame keeps its proportions):

width 200
image

width 400
image

## Checklist

- [x] New settings default **OFF** -- N/A: no new setting. Nothing renders unless Blizzard's own alert option is on, and the frame stays at its Blizzard position until the user moves it
- [x] Zero cost while disabled -- no events, tickers, or OnUpdate added; the proxy is a hidden rect and the only hooks are the SetPoint/OnShow pair the loot windows already install
- [x] Cheap while enabled -- nothing runs per frame: position/scale apply on the frame's own Show, the mover box size is computed only when Unlock Mode builds or resizes it, the live frame size is read once and cached, and the CVar is read once per mover build
- [x] No writes onto Blizzard-owned frames -- SetPoint/SetScale and the sanctioned `ignoreFramePositionManager` opt-out only; state lives in EllesmereUIDB and the module's frame-field data
- [x] Tested in-game, works on live retail; no load errors (not separately run on 12.1 PTR)

### DIFF
--- a/EllesmereUIQoL/EllesmereUIQoL_Shifter.lua
+++ b/EllesmereUIQoL/EllesmereUIQoL_Shifter.lua
@@ -782,7 +782,48 @@ local LOOT_WINDOWS = {
     { name = "BonusRollFrame",     key = "EUI_BonusRoll",   label = "Bonus Roll",   order = 640, defW = 330, defH = 120, defY = 240 },
     { name = "GroupLootContainer", key = "EUI_GroupLoot",   label = "Group Loot",   order = 641, defW = 300, defH = 80,  defY = 340 },
     { name = "AlertFrame",         key = "EUI_AlertToasts", label = "Alert Toasts", order = 642, defW = 300, defH = 100, defY = 160, fixedSize = true },
+    -- alwaysOn: no Shifter toggle of its own -- Blizzard's "Loss of Control
+    -- Alerts" combat option decides whether the frame ever appears.
+    -- scalable: one icon + timer + label block, so both size inputs drive the
+    -- same SCALE and the aspect ratio is kept.
+    { name = "LossOfControlFrame", key = "EUI_LossOfControl", label = "Loss of Control", order = 643, defW = 220, defH = 90, defY = 400, alwaysOn = true, scalable = true },
 }
+
+-- Saved scale for scalable entries, keyed by frame name.
+local function GetLootScale(name)
+    local db = EllesmereUIDB
+    return tonumber(db and db.shifterLootScales and db.shifterLootScales[name]) or 1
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
would you mind rebasing this? sorry i fell so far behind on PRs

labrie75:
> would you mind rebasing this? sorry i fell so far behind on PRs

Understood.
Please close it for now, and I’ll fix it and upload it later.
This doesn’t cause any problems like the minimap Blip icon.

PR #1121, opened 2026-08-01
zhTW: translate 46 new keys and unify threshold wording

LoChinAn:
Adds Traditional Chinese for 46 keys and corrects 7 existing ones.

New keys come from two sources:

- **v8.7.1 additions** — friendly nameplate subtitle (player title / guild),
  charge count hiding on action buttons and the cooldown manager, Always Show
  Pet Frame, and the battle-res tracker's display style, font outline and
  color options.
- **Options text the key scanner could not classify** — strings returned from
  inside `disabledTooltip = function() ... end`, strings hoisted into a local
  before being used as a field value, and labels passed positionally to
  module-local helpers. All of them reach `EllesmereUI.L()` at render time, so
  they were always translatable; they just never showed up as candidates.

Also normalizes Threshold wording: 閾值 is replaced by 臨界值 or 門檻 so each
entry matches the term already used by the neighbouring options in its panel.

Rebased onto v8.7.1. Only `Locales/zhTW.lua` is touched.


LoChinAn:
Because it's significantly behind the main branch, to avoid conflicts, I will open a new PR after closing this one.

PR #1126, opened 2026-08-01
feat(cdm): replace cooldown slots with active buffs

mTx87:
## What does this PR do?

Adds an opt-in `Replace with Buff` action to cooldown and utility icon settings.

A cooldown can be mapped to a Blizzard Cooldown Manager tracked buff. While that buff is active, its native Blizzard buff frame is routed into the cooldown's existing slot, preserving the slot's size, position, border, keybind, ordering, and Rotation Assist identity while showing the buff's native icon, duration, and stacks. When inactive or unavailable, the original cooldown is shown normally.

The picker includes Blizzard's tracked-buff catalog, including currently untalented entries. Custom Spell IDs are accepted only when they resolve to an existing Blizzard buff-viewer entry, and `cooldownID` disambiguates catalog collisions. A tracked buff can replace only one cooldown slot.

Mappings are stored in the cooldown's existing per-spec spell settings. No new event registrations or polling are added when no mappings exist.

## How was it tested?

- Tested in-game on live retail as Windwalker Monk
- Verified Spinning Crane Kick can be replaced by active Dance of Chi-Ji
- Verified the native buff icon and duration display in the original cooldown slot
- Verified the cooldown returns when the buff is inactive
- Verified the generated test addon and `/mtx87cdm` diagnostics are not part of this branch
- `git diff --check` passes and all added code is ASCII-only
- PTR/12.1 client validation and broader edge-case coverage are still pending

## Screenshots

Before/after screenshots will be attached to the PR.

## Checklist

- [x] No writes onto Blizzard-owned frames (the feature intentionally reanchors the native Blizzard buff frame into the selected slot; this needs maintainer review for the 12.1 path)

## Maintenance update (2026-08-27)

- Rebased onto upstream/main v9.0.7; current upstream taint-safe provider, item-cache, gamepad, and visibility fixes are retained.
- Reassigning or removing replacements now prunes empty per-spell settings entries, avoiding stale appearance work.
- Lua syntax, locale extraction, and git diff checks pass.


mTx87:
PR new options
replace_cd_icon_with_buff_list
replace_cd_icon_with_buff

PR without buff up (spinning crane kick icon)
replace_cd_icon_with_buff_nobuffup

PR with buff up (dance of chiji replacing spinning crane kick icon)
replace_cd_icon_with_buff_buffup

EllesmereGaming:
this is being implemented with next patch but I went about it a different way to lighten the performance hit. thank you!!

mTx87:
> this is being implemented with next patch but I went about it a different way to lighten the performance hit. thank you!!

mhmm but you missed the "custom spell id" option for that. so for example windwalker conduit buff can't be used as it is not in CDM per default.

EllesmereGaming:
this was intentional temporarily until I find a more efficient way to handle custom ids, time is limited as i leave for blizzcon tomorrow so if I can get to it tonight I will otherwise will take a look when I return

PR #1133, opened 2026-08-02
fix: apply the chosen font to the options menu widgets

dfrisone:
### Problem

Reported against Global Settings: a font imported through shared media and set as the global font is not applied to the EUI menu. The reporter also noted that some dropdowns picked the font up while others did not, and could not find a pattern:

> I cant really see a pattern how each dropdown could be tied to a function. Some modules are applying more of the fonts in the dropdown, others less

### Cause

`MakeFont` in `EllesmereUI.lua` hardcoded the bundled face:

```lua
fs:SetFont(LOCALE_FONT_FALLBACK or EXPRESSWAY, size, flags or "")
```

Every options row, panel and popup in the suite builds its text through that helper -- roughly **380 call sites across 25 files** -- so the global font setting never reached any of them.

### Why it looked patternless

The inconsistency tracks **which builder a row uses**, not which module it belongs to. A handful of specialised dropdown builders resolve the font themselves:

- `BuildVisOptsCBDropdown` (Visibility Options)
- `BuildMultiApplyDropdown`
- `BuildReorderCBDropdown`

Those call `GetFontPath("options")` directly and picked the font up correctly. Everything built through the generic `BuildDropdownControl`, and every label and header alongside it, went through `MakeFont` and stayed on the bundled face.

That is why two dropdowns sitting inches apart on the same page disagreed: on the Action Bars page, Visibility ("ALWAYS") and Visibility Options ("NONE") rendered in the chosen font while Border Style ("Solid"), Border Size ("Thin") and Icon Order ("Default") did not.

### Fix

`MakeFont` now resolves `GetFontPath("options")`, which falls through to the global font when no per-module override is set.

That is also the locale-correct source. `GetFontPath` runs through `ResolveFontName`, which forces the system glyph font on CJK and Cyrillic clients, and that is exactly what the `LOCALE_FONT_FALLBACK` check here was approximating by hand. The result is memoised in the font cache, so the call is cheap at this volume.

The bundled face stays as a fallback for two cases:

1. Before the font system is initialised, where `GetFontPath` is not yet on the table.
2. When a user-imported face fails to load. `SetFont` returns false on failure, and without that retry a broken custom font would render every menu string **invisible** rather than merely unstyled -- worth guarding when the whole feature is importing arbitrary faces through shared media.

### Scope

The EUI menu, panels and popups. The reporter also mentioned unit frames and the XP bar showing a different font; those are separate surfaces with their own module font keys and are not addressed here rather than assumed to share a cause.

Ordering is safe: `GetFontPath` is defined later in the file than `MakeFont`, but there are no `MakeFont` calls between the two, and every call site runs when a panel is built rather than at load. If it were ever called earlier the guard degrades to the previous behaviour rather than erroring.

### Testing

`luac -p` clean. Verified in game: with a distinctive imported face set as the global font, the previously-inconsistent rows now match the rest of the panel.

oh no you didnt GIF


EllesmereGaming:
this is actually intentional, EUI options panel is supposed to remain expressway, I'm normalizing the few widgets that were accidentally following the global font

PR #1136, opened 2026-08-02
[Feature] VERTICAL HEALTH BARS!!!!

Kirihasio2:
## What does this PR do?
u know what it does.

## How was it tested?

kulia tested. 8.7.2 retail

## Screenshots

image
image
image

## Checklist

- [✔️ ] New settings default **OFF** (no behavior change without opt-in)
- [✔️ ] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built
- [✔️ ] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations)
- [✔️ ] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames
- [✔️ ] Tested in-game, works on live retail; no load errors on the 12.1 PTR client

### DIFF
--- a/EllesmereUIRaidFrames/EUI_RaidFrames_Options.lua
+++ b/EllesmereUIRaidFrames/EUI_RaidFrames_Options.lua
@@ -1,4 +1,4 @@
-﻿-------------------------------------------------------------------------------
+-------------------------------------------------------------------------------
 --  EUI_RaidFrames_Options.lua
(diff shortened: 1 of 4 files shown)


Kulia240:
unfortunately it passes ulduar and other textures look fine.

I hate this so much KEKW

PR #1143, opened 2026-08-02
Use health bar border size for cast bars

DoGi114:
Use health bar border size for cast bars

## What does this PR do?
Makes unit-frame cast bars use the configured health-bar border size instead of a hard-coded value.

The existing default of 1 is retained when no border size is configured.

## How was it tested?
Tested on my local branch with the stable version of the game ( _retail )

## Screenshots
image

## Checklist

**Hard to make it OFF by default since its just inheritance ( which in my opinion should work like that from the begining.**


DoGi114:
@EllesmereGaming 
Hi! When you have a chance, could you review this small cast-bar border-size change? It makes the cast bar use the unit frame’s configured `borderSize` instead of a hard-coded value. Thanks!
Hope to see it in the next release <3

PR #1149, opened 2026-08-02
fix(qol): Auto Open Containers silently cancels the player's casts

dfrisone:
Reported by @AngelDEA (12.1.0 PTR, RU client): "if I cast anything, it force cancels every 1 second". The video shows Stormwind, out of combat, standing still, casting a 1.5s mount, and getting red "Прервано" on every attempt. Disabling EUI stopped it.

## Cause

`C_Container.UseContainerItem` is an interrupting action: the engine cancels whatever the player is casting, silently, with no error and nothing on screen connecting it to an addon.

Auto Open's cycle gated on `InCombatLockdown`, an open merchant, a loot window and its own `_openBusy` flag, but never on whether the player was casting. Any open that landed mid-cast ate the cast.

It is the only call in the module that can break a cast (checked `CastSpellBy*`, `UseAction`, `UseItem*`, `UseInventoryItem`, `UseToy`, `UseContainerItem`, `RunMacro`, the shapeshift APIs and `FollowUnit`), which is what identified the feature before any in-game testing.

The once-a-second cadence comes from a container that keeps failing to open: a slot still locked at the 0.5s recheck is neither progress nor a genuine failure, so it never lands in `_failedItems` and is retried on the next cycle, with the `UseContainerItem` bag update feeding the next scan. Nothing here is specific to 12.1.

## Fix

Gate the cycle on the player casting, at both the entry point and each `step()`. Per-step matters because a cycle paces itself across several seconds of timers, so a cast can begin long after the entry gate passed.

A cast is transient, so this defers rather than blacklists: the deferred scan resumes on the player's `UNIT_SPELLCAST_SUCCEEDED`/`STOP`/`CHANNEL_STOP`, and only when `_missedScan` shows the gate actually dropped something. Those events fire for every instant cast, so an unconditional re-scan would put a full bag walk on every global cooldown.

## Testing

Confirmed in two stages by the reporter:

- Cause: with EUI enabled and only "Auto Open Containers" turned off, the casts stopped breaking.
- Fix: on a build of this branch with Auto Open turned back on, casts are no longer interrupted and containers still open normally when not casting.

Crazy Eyes Looks Fine GIF by MOODMAN


dfrisone:
Superseded by #1151, which folds this fix in alongside the mailbox pause. Both changed the same guards in the Auto Open cycle, so they would have conflicted on merge; shipping them as one change instead. The verification here still stands and is carried over into #1151.

PR #1154, opened 2026-08-02
More Raid tool options

RedAces:
## What does this PR do?

- Add the ability to disable each pull timer (set it to zero)
- Add checkboxes to disable the "role check", "convert to raid" and "disband" buttons

## Screenshot

Screenshot 2026-08-02 185132

## How was it tested?

Tested on 12.0.7 retail
NOT tested on 12.1 PTR

## Checklist

 - these new settings default to **ON** because thats the previous state


RedAces:
Added: hide the raid tools frame if youre in raid and are neither a Leader or Raid Assistant

RedAces:
Ive split this PR into 3:
* https://github.com/EllesmereGaming/EllesmereUI/pull/1222
* https://github.com/EllesmereGaming/EllesmereUI/pull/1221
* https://github.com/EllesmereGaming/EllesmereUI/pull/1220

PR #1169, opened 2026-08-03
UnitFrames: add Mouseover option for Blizzard Aura Frames

Trenchfoote:
## What does this PR do?

Adds a "Mouseover" toggle to Unit Frames -> Blizzard Aura Frames (Player Buffs &
Debuffs). When enabled, the buff and debuff frames stay hidden and each one
reveals only while the cursor is over its icons. Buffs and debuffs are
independent, so hovering one does not reveal the other.

Defaults off, applies live with no reload, and works whether or not the
styled-aura skin ("Enable Styled Buffs & Debuffs") is turned on -- which is why
the toggle sits outside that section gate.

## How was it tested?

Tested in game on both clients -- live retail (12.0) and the 12.1 PTR. Same addon
build on each.

Toggled on/off: frames hide, reveal per-frame on hover, tooltips still work while
faded, and both return to full alpha when the setting is turned off. No load
errors on either client.

## Screenshots

_Adding before/after shortly._

## Checklist

### Notes for review

Driven by the aura buttons' own `OnEnter`/`OnLeave` rather than a ticker: an
alpha-0 frame still receives mouse input, so the buttons keep their tooltips and
keep firing while invisible. The fade-out defers one frame and re-tests which
button is hovered, so sliding between icons does not flicker.

Hover is tested against the buttons, never the parent frame -- `BuffFrame`'s rect
extends well past the visible icons (Edit Mode region, expand button,
consolidation frame), so `frame:IsMouseOver()` stays true long after the cursor
has left the auras.

Nothing exists until first enable: no frame, no event registration, no hooks.
Buttons are hooked on a deferred sweep off player `UNIT_AURA`, skipping
already-hooked buttons via the `ffd` weak table.

Touches the player aura frames but adds no `IS_121` branch -- it uses the same
`frame.auraFrames` / `AuraContainer` accessors the existing skin path already
uses ungated. Flagging it in case you want that gated differently.

### DIFF
--- a/EllesmereUIUnitFrames/EUI_UnitFrames_Options.lua
+++ b/EllesmereUIUnitFrames/EUI_UnitFrames_Options.lua
@@ -15672,6 +15672,7 @@ initFrame:SetScript("OnEvent", function(self)
             db.profile.playerAuras[key] = v
             if ns.RefreshPlayerAuras then ns.RefreshPlayerAuras() end
             if ns.ApplyPlayerAuraScale then ns.ApplyPlayerAuraScale() end
+            if ns.ApplyPlayerAuraMouseover then ns.ApplyPlayerAuraMouseover() end
         end
 
         _, h = W:Spacer(parent, y, 20);  y = y - h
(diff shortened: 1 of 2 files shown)


Trenchfoote:
Discord_LaF4pdGDQ0

PR #1177, opened 2026-08-03
chore(ci): improve github workflow security

math280h:
## What does this PR do?

_This PR does not touch the addon itself_

This PR improves the security around the github workflows according to [GitHub Security Guidance around workflows](https://docs.github.com/en/actions/reference/security/secure-use)

This could be improved further by adding a `CODEOWNERS` file that limits who can approve changes to these files as "in theory" someone with write access could do a lot of "damage" through workflows. But I would need to work with you all to make sure the right people/teams are in it.

## How was it tested?

Tested locally as much as I could using [act ](https://github.com/nektos/act) (ofc without creds for release)

## Checklist

- [N/A] Tested in-game, works on live retail; no load errors on the 12.1 PTR client

### DIFF
--- a/.github/workflows/locale-check.yml
+++ b/.github/workflows/locale-check.yml
@@ -11,11 +11,16 @@ on:
       - 'Locales/_keys.txt'
       - '.tools/extract-locale-keys.sh'
 
+permissions:
+  contents: read
+
 jobs:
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
thank you!!

PR #1187, opened 2026-08-04
Action Palette: a hold-and-release action picker

nulltyto:
## What this is

A new child addon, `EllesmereUIActionPalette`. You hold a binding, steer to an action, and release to use it. The release resolves inside a secure snippet, so the palette works in combat.

Four steering models:
- **Grid** — a block of entries; the nearest entry to the cursor is selected.
- **Fan (cursor)** — a fixed strip steered by pointing.
- **Fan (scroll)** — a coverflow strip scrubbed with the mouse wheel; usable while the right mouse button steers the camera.
- **Arc** — entries on a ring or arc, selected by angle.

An entry can open another palette (one level of nesting). Grid nests come in three styles: a lane along the palette's perimeter, a halo around the entry, or a popout block beside the palette.

## Status: early draft, for eyes only

**Nested palettes are still broken in-game.** The current design gates a nest behind secure OnEnter/OnLeave frames, and the latest round of fixes has not yet been validated live. Everything else on the list below works in-game today:

- Open, steer, release, and cancel in all four steering models.
- Mount/toy/item/macro/spell entries, with a picker for adding them.
- Saved variables migrate from the old Radial Wheel name.

Known open items besides nesting: unlock-mode placement is stand-in X/Y sliders, and the override bindings are not yet scoped.

Feedback on the interaction model and the options layout is what we are after at this stage.


nulltyto:
Pushed the next round of nested-palette work (rebased onto current main, plus four new commits).

**Grid nests**
- A lane run may use the whole perimeter before it stacks a second row, so two sub-palettes in adjacent cells no longer collapse into clusters.
- Each nest's coverage now has every other nest's entry cell carved out of it, so you can glide from one sub-palette's entry straight onto its neighbor's and the nests hand off in place.

**Arc nests**
- Fixed children that showed but could not be selected: the disarm test had a dead ring between the parent's icon and the first ring of children, so every reach disarmed the nest the moment it left the icon. The ground is now a beam plus a wedge centered on the parent.
- Children spread along a single ring up to Max Nest Span (default a quarter circle) before they spill into a second ring.

**Options**
- Slot labels now ship off, and the center logo ships on. Profiles that never touched these pick up the new defaults; explicit choices are kept.
- Palettes can be named, and given an icon shown by the nested entry and the picker. Unset keeps the current look.

All of it is verified offline by the agreement harness (real secure-snippet plumbing, walks to every child, release resolution). Still needs in-game confirmation:
- Swap between the two adjacent bottom-row sub-palettes by direct glide.
- Reach arc children, including near a widened arc's far ends.
- Watch for adjacent arc nests flapping on small movements (the arc no longer darkens the other entries' gates while one nest is armed).
- The rename/icon inputs in Palette Setup.

Known limit, fix in progress: with three or more sub-palettes on one grid the region budget can overflow and drop coverage, which can break a reach or a swap. A follow-up commit will make the budget lossless.

nulltyto:
Update: markers, preset palettes, and add/delete palettes (`68d07b70`). The branch was also rebased onto current main.

**Markers in the action picker.** A new Markers category offers the eight target markers, the eight world markers, and a clear entry for each set, in the star-to-skull order the game's own marker menus use. They are two new slot kinds that resolve to `/tm`, `/wm`, and `/cwm` macrotext, so the secure snippet is untouched — it already writes one attribute per cell, and these slash commands do what the raid manager's buttons do. World marker slots store the icon position and map to the engine's marker numbers through Blizzard's `WORLD_RAID_MARKER_ORDER`.

**Preset palettes.** The Add Palette chooser offers presets beside the default empty palette: Target Markers, World Markers, Hearthstones, Teleports, Potions, and Druid Forms. Each preset builds from what the character knows or carries at the moment the chooser opens, and a preset with nothing to offer stays out of the list — which is also how the class-specific ones gate themselves.

**Add and delete instead of the count slider.** The Number of Palettes slider is gone. Add Palette opens the chooser above and points the editor at the new palette. Delete Palette (confirmed by a popup) really removes the palette rather than hiding it the way the slider's decrease did: the palettes above close ranks, nested entries repoint or leave with it, and each shifted palette's keybind moves down with it, both keys if two are bound. Deletion is blocked in combat because it touches bindings.

Verified offline (syntax, a fresh-context review pass, and the nest agreement harness still reports ALL AGREE) but not yet live-tested in-game. The first things to try live: marking while grouped (both sets, including the clears, in combat), the chooser anchoring under the Add Palette button, and deleting a middle palette with keys bound on either side of it.

nulltyto:
Status: a bug-hunt pass over the module found eight bugs. All are fixed in the nine commits ending at 12daf916.

**Fixes**

- **Stuck nests (the important one).** A claim that arms geometrically — the press pre-arm, or the re-arm after a leave — never gets Blizzard's `_wrapentered` flag, because only a motion OnEnter raises it, and the sandbox cannot set underscore attributes. Its parent gate's OnLeave pre-body therefore never ran: step off the cell through ground no region rect covers and the nest stays open, the block stays dim, and no other nest can arm for the rest of the hold. The disarm now hangs off an OnEnter instead, which needs no flag: a screen-wide, motion-only floor gate per palette sits below the real gates, shows while a claim is armed, and runs a claim-agnostic leave test on entry.
- **Two-key holds.** A second palette key pressed mid-hold re-laid the shared live view, so the first key's release mapped its chosen cell through the wrong palette — wrong pet, mount, or spec — and then closed the second palette mid-hold. The screen now belongs to the key that holds it: the second press opens nothing, and its release fires nothing insecure and closes nothing. Its own secure action still fires from its own pushed geometry; insecure code cannot stop that.
- A release exactly on the arc's outer boundary rounded up to index n+1, which with nests pushed is the first nested cell — it fired without its claim armed. The release snippet now bounds the index the same way HitTest does.
- A nested-palette slot with a missing or out-of-range palette number made EnsurePalette error on a nil compare, which broke every paint of the containing palette. It now degrades to the question-mark render.
- The editor's fan strip and the options preview measured the strip at the full count, but the fold draws only half of it — the preview shrank its icons to half of what fits.
- Switching to a profile with fewer palettes while the options page was open made the page silently create a phantom out-of-range palette in the saved variables. The editing index is now clamped in BuildPage and reset on module reset.
- The dim-unusable tooltip promised toy tinting the module deliberately does not do.
- A full 12-slot palette's preview reserved room for a "+" placeholder that is not drawn.

**Verification so far**

`luac -p` is clean on both files and the nest agreement harness reports ALL AGREE. The harness compiles the floor-gate snippets for real, but its focus model does not exercise the floor's behavior, so the first two fixes need a live pass:

- [ ] Cursor-mode 3x3 grid, middle nest cell: press, move off the cell through uncovered ground — the nest must disarm.
- [ ] Normal UI hover returns after every hold ends, including cancels and Escape (the floor must hide).
- [ ] ARC palette: an armed nest now disarms over a plain entry. This follows the palette's own ground test, but it is a behavior change — confirm it feels right.
- [ ] Hold key A, press key B mid-hold, release in both orders: A's palette stays, B fires nothing insecure and closes nothing.
- [ ] Switch to a profile with fewer palettes while the options page is open: no phantom palette in the saved variables.
- [ ] A 12-slot palette's preview fills its block.

**Known limits:** while a claim is armed, the floor takes hover (not clicks) from the rest of the screen for that hold. The drawn-versus-pushed divergence after an in-combat edit (a refused push) is not fixed; any cheap guard would also suppress correct in-combat fires.

nulltyto:
### f542d44a — world marker order, per-palette keys, cycling markers

**World markers placed the wrong marker.** `WORLD_MARKER_ENGINE` was Blizzard's `WORLD_RAID_MARKER_ORDER` copied whole. That constant lists its eight engine numbers **skull to star** — the order its own dropdown draws them in — so read star to skull it lands on the marker mirrored about the middle. Reported live: star placed the skull, the orange circle placed the red cross, the purple diamond placed the blue square, the green triangle placed the moon. The map is now `{5, 6, 3, 2, 7, 1, 4, 8}`, which is the engine's own run of blue, green, purple, red, yellow, orange, silver, white. Target markers were never affected — they take the icon position directly.

**Every palette can take a key.** `MAX_BOUND_PALETTES` is gone and `Bindings.xml` carries all sixteen entries, so the "Nested Only" state no longer exists: a palette added as number 7 binds like number 1. To keep that free, a secure button is built only for a palette that *has* a key. A profile that binds two of its sixteen pays for two buttons and two pushes; the rest are skipped by `PushPalette` and still reach the sandbox through whatever nests them.

**Cycling marker entries.** Two new kinds, `cycleraidtarget` and `cycleworldmarker`. Each press places the next marker, star to skull and back — the whole set in one slot, for a palette with no room for nine. The run is pushed as one macro text per step and the **secure snippet steps the counter itself**, because an insecure `SetAttribute` is refused in combat, which is the whole of when marking matters. `OnPostClick` mirrors the counter back onto the slot, so it survives a reload and the icon always draws the marker the next press places. Offered in the picker only, never in the presets — a preset that lays out all eight has nothing to cycle.

**Keybind overwrite now asks.** `SetBinding` takes a key from whatever holds it without a word, and the displaced binding is often something the user does not miss until they reach for it mid-fight. The dialog names both sides and offers Rebind / Keep. It only asks for a real theft: an unbound chord takes nothing, and rebinding a palette to the key it already holds takes it from itself. Combat is re-checked on confirm, since the dialog can sit open while a fight starts.

**Markers picker split.** `Markers` now opens `Target Markers` (11 rows) and `World Markers` (10) instead of one list of 21. A picker category can carry a `subs` list, and the Back row walks one level at a time — general, so any category that grows too long can be split the same way.

Also carries the settings page overhaul from the previous batch: the page is cut to the rows most users need, with the rest in four cog popups.

#### Verification

`.tools/palette-nest/nest_agree.lua` gained a check that presses ten times round each eight-step cycle and asserts the fired macro text, the wrap, and that the drawn name matches what then fires. Reverting the marker table to the old values makes it fail with exactly the reported symptoms, which independently confirms the diagnosis:

```
cycle world marker: press 1: fired /wm 8, wanted /wm 5; press 2: fired /wm 4, wanted /wm 6;
                    press 3: fired /wm 1, wanted /wm 3; press 4: fired /wm 7, wanted /wm 2
```

With the fix in place the harness reports ALL AGREE.

#### Still to confirm in game

- All eight world markers place the marker their icon shows.
- A cycle entry steps through all eight and wraps; its icon between presses shows what that press will place; it resumes after a reload; it still advances in combat.
- A palette at index 7+ binds, and its key opens it after a reload.
- Deleting a palette in the middle shifts its successors' keys down.
- The keybind dialog names both actions; Keep changes nothing, Rebind takes the key.
- Picker navigation: Back reads "« Markers" from a marker list and "« Categories" from the Markers list; Custom Macro... still reads "« Categories" after visiting Markers; no row sits under the Back row on any level.

The picker navigation has no offline coverage — the harness does not load the options file.

PR #1199, opened 2026-08-05
Feature/player aura bars

JuJuFX-dev:
## Player Aura Bars

Replaces `EllesmereUIUnitFrames_PlayerAuras.lua` (the old aura-style module) with **Player Aura Bars (PAB)**: a filter-driven buff/debuff bar system built on `EllesmereUI.AuraKit`, with a redesigned options UI and an embedded live preview.

### Features

- **Filter engine**: buffs/debuffs resolve through a class vocabulary (All Buffs, Has Duration, boss/role/priority aura, stealable, dispel-typed, Extra Spells, etc.) shared with `EUI_UnitFrames_AuraContainers.lua`'s `TOKEN_CLASSES`/`CANDIDATE_CLASSES` instead of hardcoded spell lists. Multiple filters combine per bar.
- **Three built-in bars**: Buffs, Debuffs, and External Defensives (migrated into PAB as a third built-in bar, previously its own thing).
- **Unlimited custom bars**: each independently configurable — filters, Sort Method/Direction, Growth Direction (horizontal Left/Right and now vertical Up/Down), Icon Effects Per-Filter, Row Spacing, Icons Per Row.
- **Icon Effects Per-Filter**: per-filter icon styling (glow/border) for both buffs and debuffs, applying correctly on the default Debuffs bar as well as custom bars.
- **Live preview**: embedded preview in the options pane (`EUI_PlayerAuraBars_ManagerPages.lua`) simulates Filters, Sort, Extra Spells and Icon Effects against a placeholder/real-spell-icon pool without requiring live auras in-game.
- **Sort Method/Direction**: e.g. Important-first, with the preview's simulation matching real in-game resolution order (not just re-shuffling which icons are picked).
- **Unlock Mode support** for custom bars: drag-to-position, Grow Left/Right/Up/Down, wired through `EUI_UnlockMode.lua`'s generic mover (see below).
- **EditMode/legacy seeding**: on first PAB use, default Buffs/Debuffs bars are seeded from the user's existing Blizzard EditMode aura frame settings and the old aura-style module's saved config, so upgrading users keep their prior setup instead of resetting to PAB defaults.
- **Bar tile subtitles** live-refresh based on selection mode (Filters vs Extra Spells) instead of a static resolved-count snapshot.

### Bug fixes bundled in this branch

- Stale `duration = 0` on pre-existing buffs after login/reload (2s post-login aura-container poke, with an earlier poke variant reverted as ineffective).
- Cooldown swipe overlay reappearing on aura updates / after zone changes (swipe alpha forced to 0 as a fallback); right-click-to-cancel added for buffs.
- Custom/default Debuff bars showing every debuff regardless of configured filters.
- Icon Effects Per-Filter not applying on the default Debuffs bar.
- Default-value mismatches between the options UI and actual rendering (default + custom bars).
- Grow Left/Right no-op for custom bars in Unlock Mode.
- Unbounded custom debuff bar table growth on delete; cog-tooltip fallback not resolving through `L()`.
- Reload lag and a forbidden-object crash in debuff Icon Effects glow layering (see AuraKit change below).

### Shared/global file changes — and why

This branch touches four files other modules depend on. None of them change existing behavior for non-PAB consumers except where noted:

- **`EllesmereUI_AuraKit.lua`**
  - Fixed `ApplyStyleToRegions`'s stack-count text color to read `c.r/c.g/c.b/c.a` instead of `c[1..4]` — `stackColor` is stored as an `{r,g,b,a}` table everywhere it's built, so the old indexed access returned `nil` and crashed. This is a real AuraKit bug affecting any consumer with a custom `stackColor`, not PAB-specific; PAB's Icon Effects work just surfaced it.
  - Shifted the button's internal frame-level ladder (dispel ring `+3 → +4`, stack/duration text carrier `+4 → +5`) to open a level for the DM per-filter glow (`ApplyDmFx`/`PAB_ApplyDmFx`), which now sits between the per-filter border override (`+2`) and the dispel ring. Without this, PAB's per-filter glow could render on top of the dispel ring/text, or (the forbidden-object crash) collide with a level Blizzard's engine already owns. `EUI_RaidFrames_AuraContainers.lua`'s `ApplyDmFx` frame levels are updated in lockstep (`+2 → +3`, carrier `+4 → +5`) to match this shared ladder.

- **`EllesmereUIUnitFrames/EUI_UnitFrames_AuraContainers.lua`**
  - Added a `dispeltyped` candidate class (`TYPED_DEBUFF_TYPES`) and threaded `candValue` through `BuildChain`/`DeclareElementGroup`/`ApplyGroupConfig` so a candidate class can carry a value set, not just a boolean — needed for PAB's own filter resolution to reuse the exact same class definitions.
  - Exposed `TOKEN_CLASSES`/`CANDIDATE_CLASSES` as `ns.UF_TokenClasses`/`ns.UF_CandidateClasses` so PAB shares one source of truth for the class vocabulary instead of maintaining a second copy that could silently drift.
  - Removed the `debuffAnchorBuffs` ("Anchor Buffs with Debuffs") merge path in `AnchorContainer`/`ApplyGroupConfig`. This was the old aura-style module's mechanism for stacking buffs onto the debuff container; PAB owns bar composition itself now (separate bars, separate anchoring), so the merge became dead weight and a second, conflicting anchoring path for buffs. Confirmed intentional, not incidental scope creep.
  - Added `SetTextColor` change-guarding via a `CK()` color fingerprint (`d.ufDurColor`/`d.ufStackColor`) and cached `style.fontPath` once per style rebuild instead of resolving it on every `ApplyUFText` call — this runs on every registered aura button on display updates, so redundant `SetTextColor`/`GetFontPath` calls are real, avoidable per-frame cost.

- **`EUI_UnlockMode.lua`**
  - `CreateMover` now recognizes the `PAB_` bar-key prefix: `canGrow` includes it, and the Grow-direction popup reads/writes direction via `EllesmereUIUnitFrames:GetGrowDirectionForBar`/`SetGrowDirectionForBar` instead of the ActionBars/CDM-specific paths those functions already used for other bar families.
  - The popup omits "Grow Centered" for `PAB_` bars: AuraKit's `SetFlowLayoutGrowthDirection` (`AnchorUtil.FlowDirection`) is a strict Left/Right/Up/Down axis with no centered concept, unlike the CDM/ActionBars bar renderer this popup was originally built for. Offering it would show a selectable option PAB's layout can't actually honor.

### Files touched

- `EllesmereUIUnitFrames/EllesmereUIUnitFrames_PlayerAuraBars.lua` (new, ~4.3k lines) — core PAB module
- `EllesmereUIUnitFrames/EUI_PlayerAuraBars_ManagerPages.lua` (new, ~2.5k lines) — options UI / manager pages
- `EllesmereUIUnitFrames/EllesmereUIUnitFrames_PlayerAuras.lua` — removed (superseded by PAB)
- `EllesmereUIUnitFrames/EUI_UnitFrames_Options.lua`, `EllesmereUIUnitFrames.lua`, `EllesmereUIUnitFrames.toc` — wiring/registration
- `EUI_UnlockMode.lua`, `EllesmereUIUnitFrames/EUI_UnitFrames_AuraContainers.lua`, `EllesmereUIRaidFrames/EUI_RaidFrames_AuraContainers.lua`, `EllesmereUI_AuraKit.lua` — shared-file changes, see above

### Testing

- Manually tested in-game: default Buffs/Debuffs/External Defensives bars, custom bar creation/deletion, filter combinations, sort methods, icon effects per-filter, Unlock Mode dragging/growth direction (horizontal and vertical), and EditMode/legacy seeding on both a fresh profile and an existing one.
- All Lua files checked with `luac5.1 -p`.


JuJuFX-dev:
manually merged

PR #1203, opened 2026-08-05
Modify the transparency adjustment range of the raid frames

ZhangXiaoD:


EllesmereGaming:
working on that file so manually merged, will be in with next update!

PR #1220, opened 2026-08-06
Raid tools hide buttons

RedAces:
## What does this PR do?

Adds the ability to hide each button.

## How was it tested?

Tested on 12.1 retail.

## Checklist


RedAces:
Rebased onto 12.1 main, not tested again yet.

RedAces:
Merge main into it and tested it on 12.1 live: everything works!

PR #1223, opened 2026-08-06
Fix addon compartment

RedAces:
## What does this PR do?

The addon compartement was not visible. Added it back!

## How was it tested?

Tested on 12.0.7 Retail.
NOT tested on 12.1 PTR

## Checklist


EllesmereGaming:
manually merging this, in the middle of a major file structure change for 12.1, will be in with next patch

PR #1224, opened 2026-08-06
Fix: player unit frame castbar shows no target

svart2521:
Bug: https://discord.com/channels/585577383847788554/1505218793851457727

Player unit frame castbar would not show target, while on target frame cast this was not the case.
Removed the  (and ownerUnit ~= "player") condition so the player frame's castbar goes through the same target-name lookup as target/focus.


EllesmereGaming:
fixed locally with a migration, will be in with next patch ty!

PR #1226, opened 2026-08-06
Stop toggled-off action bars from updating

Nnoggie:
## Summary

When you turn off a bar with the **Toggle Action Bar** keybind, the bar UI now fully switches off instead of continuing to update in the background. Your spell keybinds keep working normally.

The option is now called **Toggle Action Bar** instead of **Toggle Action Bar Visibility**, with matching tooltip text and translations.


EllesmereGaming:
manually merged, undergoing big file structure change for 12.1, ty king in with next patch

PR #1239, opened 2026-08-07
Action Palette: a hold-and-release action picker

nulltyto:
Supersedes #1187: same feature, rebuilt as three commits on current
main with the review feedback folded in. Notable differences from that
PR: the master switch now defaults **off**, the Popout grid nest style
was removed (it converged with Lane in behavior), and nested palettes
now contribute all 12 of their entries except where the nest is
bounded by its parent's own region (an arc's sector, a halo's ring of
eight), which keeps the 8-entry cap.

## What does this PR do?

Adds a new optional child addon, EllesmereUI Action Palette. The player
holds a keybind to open a palette of up to 12 actions, points or
scrolls to choose, and releases the key to fire. Palettes come in
grid, arc and fan layouts, can nest one level, and work in combat:
selection and firing run inside secure snippets, so no insecure code
touches protected state during lockdown.

Slots hold spells, items, toys, mounts, pets, macros, world markers
and equipment sets. A settings page with a live preview covers slot
assignment (drag to reorder, category picker, presets), per-palette
shape and appearance, and one keybinding per palette.

Everything is opt-in: the master switch defaults off, and with the
module disabled no events are registered, no override bindings exist,
and no secure buttons are built. Even when enabled, nothing exists
until a palette key is bound.

Also included:

- `.tools/palette-nest/`: an offline harness that executes the real
  WrapScript/snippet plumbing and checks the secure and insecure
  sides agree about nest arming and teardown. `.pkgmeta` already
  ignores `.tools`, so it never ships. Happy to drop it from the PR
  if you prefer no dev tooling in-tree.
- One small widget-layer change: `BuildCogPopup` rows accept a
  `noCapture` flag (same flag `DualRow` already honors) so a popup
  can mix capturing and non-capturing rows.

Two known deviations, called out rather than hidden:

- One frame (the palette live view) is created at file scope while
  the module is disabled. This is deliberate: a frame created inside
  OnEnable is billed by the CPU profiler to the loading addon, and
  the in-source comment explains it. Its handlers do nothing while
  disabled.
- The sixteen keybinding rows appear in the Key Bindings panel even
  while disabled. Gating them on the enable switch would make it
  impossible to bind a key before enabling.

## How was it tested?

- Live retail client (12.0.x): tested in dungeon and follower dungeon w/ dev mode
- The offline nest harness passes (`ALL AGREE`) across the arming,
  handoff and teardown paths, including icon-overlap sweeps of every
  layout/nest-style/scale combination.
- `luac5.1 -p` clean on both module files.
- PTR 12.1: NOT YET TESTED

## Checklist

- [x] New settings default **OFF** (master switch `enabled = false`;
      the palette cannot open without both enabling and binding a key)
- [x] Zero cost while disabled: no events registered, no polling, no
      hooks doing work, no secure buttons or bindings built (one inert
      file-scope frame, explained above)
- [x] Cheap while enabled: event-driven; the only OnUpdate runs while
      a palette is held open and stops with it
- [x] No writes onto Blizzard-owned frames; the module only touches
      its own frames (secure buttons, gates, catcher are all
      addon-created; Blizzard code is never SetScript'd)
- [ ] Tested in-game, works on live retail; no load errors on the
      12.1 PTR client [FILL IN after live pass]


EllesmereGaming:
you dropped this 

giphy

PR #1247, opened 2026-08-07
Fix: Default UI bag support for Shifter

svart2521:
Bug: https://discord.com/channels/585577383847788554/1523682056356495372
Added support to allow frame movement and zooming to Shifter module.


EllesmereGaming:
this one introduced some complexity so not doing this for now

PR #1264, opened 2026-08-08
Unit Frames: icon spacing for the Blizzard buff & debuff grid

labrie75:
## What does this PR do?

Adds icon spacing to the Blizzard Buffs & Debuffs grid, so the stock aura grid
can be tightened up (or spread out) the same way its icon size already can. 
At a glance:

| Area | Before | After |
|---|---|---|
| Grid gap | fixed 5px (Blizzard's layout) | Spacing X/Y sliders per grid, -30..30, default 5 = stock |
| Where | -- | the existing Icon Zoom cog, a Spacing X/Y pair under each Zoom slider |
| Default behavior | -- | 5 is Blizzard's own gap, so the default changes nothing; negative values overlap deliberately |

### Layout (`EllesmereUIUnitFrames_PlayerAuras.lua`)

- Buttons are re-anchored in a chain -- each icon relative to its neighbour, the first left on Blizzard's own anchor -- which reproduces the same grid at any gap
- Row/column direction comes from the container's own `iconStride` / `addIconsToRight` / `addIconsToTop`, so Edit Mode layouts are honored
- No aura data is read at any point (secret on Midnight); at the default the pass short-circuits
- Returning to the default (sliders reset, or a profile switch to one with no spacing saved) restores Blizzard's grid immediately: chaining once at the stock gap reproduces it exactly, tracked per grid so an untouched grid still costs nothing
- Rides the existing paths: the `UpdateGridLayout` hooks the skin already installs (using the aura list Blizzard just laid out, so an aura applied this pass is spaced on the same frame), the coalesced refresh pass, and `ReloadFrames` for profile swaps

### Options (`EUI_UnitFrames_Options.lua`, `EllesmereUIUnitFrames.lua`)

- Four sliders in the existing Icon Zoom cog, grouped buff-first / debuff-second so the Spacing X/Y pair under each Zoom slider reads as that group's spacing
- Every label reuses an existing locale key (Spacing X / Spacing Y), so no new strings and every locale renders it today
- Stored per profile in `playerAuras` alongside iconSize and the zoom values; a profile switch re-anchors immediately

### Compatibility

- Gated behind Styled Buffs & Debuffs like the rest of that page; the four values default to 5 (= stock), so existing profiles see no change
- No new events, tickers or OnUpdate -- anchor calls only, once per aura change

## How was it tested?

Tested in game on live retail with a koKR client:
positive, zero and negative spacing on both buffs and debuffs, with icon size and icon zoom changed
alongside

auras applied and expiring in combat (including grids that wrap to several rows)

Edit Mode grid direction changes and profile switches between
profiles with different spacing. No errors, no per-frame work added.

## Screenshots

Default (5 = Blizzard's own gap) and tightened:
image
image

image

## Checklist

- [x] New settings default **OFF** -- the four values default to 5, i.e. Blizzard's own spacing: the layout code short-circuits and the grid is untouched
- [x] Zero cost while disabled -- no new events, tickers or OnUpdate; at the default the hook compares two numbers and returns
- [x] Cheap while enabled -- anchor calls only, once per aura change (no aura reads, no per-frame work); the skin's own refresh pass is reused rather than a second one added
- [x] No writes onto Blizzard-owned frames -- SetPoint on the aura buttons only, guarded against re-entry
- [x] Tested in-game, works on live retail; no load errors

### DIFF
--- a/EllesmereUIUnitFrames/EUI_UnitFrames_Options.lua
+++ b/EllesmereUIUnitFrames/EUI_UnitFrames_Options.lua
@@ -15737,6 +15737,7 @@ initFrame:SetScript("OnEvent", function(self)
             db.profile.playerAuras[key] = v
             if ns.RefreshPlayerAuras then ns.RefreshPlayerAuras() end
             if ns.ApplyPlayerAuraScale then ns.ApplyPlayerAuraScale() end
+            if ns.ApplyPlayerAuraSpacing then ns.ApplyPlayerAuraSpacing() end
         end
 
         _, h = W:Spacer(parent, y, 20);  y = y - h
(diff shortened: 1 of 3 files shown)


EllesmereGaming:
this system is being upgraded to a new system in a few days for 12.1

PR #1274, opened 2026-08-09
[Feature] Action bars: Add an option to hide Blizzard's stock vehicle bar

Kirihasio2:
## What does this PR do?

New toggle under Menu, Bags & XP Bars > VEHICLE BAR, off by default.

## How was it tested?

Live 8.7.7

## Screenshots

image
image

## Checklist

- [ ✔] New settings default **OFF** (no behavior change without opt-in)
- [✔ ] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built
- [✔ ] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations)
- [✔ ] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames
- [✔ ] Tested in-game, works on live retail; no load errors on the 12.1 PTR client

### DIFF
--- a/EllesmereUIActionBars/EUI_ActionBars_Options.lua
+++ b/EllesmereUIActionBars/EUI_ActionBars_Options.lua
@@ -1334,6 +1334,22 @@ initFrame:SetScript("OnEvent", function(self)
         _, h = W:Spacer(parent, y, 12);  y = y - h
 
         -------------------------------------------------------------------
+        --  VEHICLE BAR
+        -------------------------------------------------------------------
+        _, h = W:SectionHeader(parent, "VEHICLE BAR", y);  y = y - h
+
+        _, h = W:DualRow(parent, y,
+            { type="toggle", text="Hide Blizzard's Vehicle Bar",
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
the top-level SetAlpha approach is good but can't merge mainly due to the recursive calls on enable mouse:

1. SetBranchMouse recursively calls EnableMouse on OverrideActionBar's descendants. Walking a Blizzard frame tree and force-enabling mouse on children can cause serious issues (a recursive re-enable exactly like this caused the v6.3.6 M+ cursor lockout catastrophe). Suppress the top-level frame only: SetAlpha(0) + EnableMouse(false) on the bar itself, and on restore only touch what you disabled.

2. The event frame + apply passes run for everyone even with the toggle off (6 events, plus the sweep on every vehicle transition). Zero cost while disabled means register the events only when the setting is on, and unregister when it's turned off.

3. The 0/0.3/0.8 retry ladder races Blizzard's dock. Hook the deterministic edge instead (e.g. the bar's OnShow via HookScript) and re-assert once.

PR #1278, opened 2026-08-09
fix(cooldownmanager): keep a charge spell correct at zero charges

nulltyto:
## What does this PR do?

Fixes a charge spell in the Cooldown Manager misbehaving once it runs **out** of charges. Reported by @Jesir against 8.7.7 on Retribution Judgment, with Hide Active State, Hide Charge Count at 0, and Cooldown State Effect (Hide Swipe Charges, Hide Duration when Charges >0).

Three symptoms, one cause:

- the icon regains full saturation shortly before the recharge ends, so it looks castable when it is not
- the recharge swipe breaks and is replaced by global cooldown geometry
- the duration text disappears for the final second of the recharge

**Cause.** Blizzard's `CooldownViewer` drops the recharge timer at zero charges -- `CheckCacheCooldownValuesFromCharges` gates on `currentCharges > 0` -- and drives the widget from `C_Spell.GetSpellCooldown` instead. That call returns the **global** as soon as the global outlasts what is left of the recharge, so pressing any ability in the last second flips `isOnGCD` true and `isOnActualCooldown = not isOnGCD and cooldownIsActive` short-circuits to false. `Blizzard_ActionBar` avoids this with a separate `chargeCooldown` widget driven from `chargeInfo.isActive`, and `Blizzard_CooldownBroadcaster` returns nil when `isOnGCD`; the viewer is the outlier.

EllesmereUI had independently reproduced the same inference in three places, which is why it surfaced as three separate symptoms. All three now ask one helper that never consults the global: `C_Spell.GetSpellCharges` for whether a recharge is running, and Blizzard's own `wasSetFromCharges` for whether any charge is in hand.

The helper answers only where it has standing to. Buff viewer items never set `wasSetFromCharges` -- `AddVisualDataSource_Charges` is reachable only from `CooldownViewerCooldownItemMixin`, while the buff mixins descend from `CooldownViewerItemMixin` and clear the field without re-adding it -- so a plain `false` there would read as "out of charges" for every charge spell on a buff bar. And when an aura owns the display, `CheckCacheCooldownValuesFromAura` has deliberately re-pointed the widget at the buff duration and lit the icon. Both cases return "unknown" and leave the caller's existing verdict untouched.

Existing settings keep their precedence: Keep Colored (On CD) still wins over the new greying, and Hide Recharge Swipe is deliberately left scoped to the "still have a charge" case.

**Secret safety.** `SpellChargeInfo.isActive` and `maxCharges` are `NeverSecret`, and `wasSetFromCharges` / `wasSetFromAura` are plain booleans that Blizzard computes in untainted code and caches -- so reading them is clean even though comparing `currentCharges` ourselves would be secret. The secret charge count is never read, stored, or compared.

## How was it tested?

Live retail 12.0, Retribution Paladin, Judgment (2 charges).

Reproduced the original report first, then confirmed each fix in game across several runs: spend both charges, build Holy Power, and spend it at 3s / 2s / 1s before the recharge completes. Verified the icon stays desaturated for the whole recharge, the recharge swipe survives the global, and the duration text appears immediately on dropping to zero and runs the full recharge.

Instrumented the window with a `SetDesaturated` write census to confirm which writer the icon actually wears, since `IsDesaturated()` stops answering once the texture has been written from a secret.

Not tested on the 12.1 PTR client.

## Screenshots

Not included -- the change has no new visual design, it restores the intended appearance of existing states. The reporter's videos of the broken behaviour are on the originating issue.

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) -- N/A, no new settings; this is a correctness fix to existing ones
- [ ] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built -- **partly.** There is no setting to disable, so the zero-charge pass runs unconditionally. Buff viewer icons leave on a single `type()` check; cooldown icons cost one `GetSpellCharges` plus one override lookup per desaturation write. The override lookup is deliberately not memoised: a proc can swap the override without the frame's spellID changing, so a memo keyed on that id would go stale mid-combat.
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations) -- the one `OnUpdate` added is a single coalesced settle pass that hides itself on its first tick, on a frame that only exists for players using Hide Duration when Charges >0
- [x] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames -- all state lives in the existing weak `hookFrameData`; the two `SetScript` calls are on an EllesmereUI-owned pooled frame
- [x] Tested in-game, works on live retail; no load errors on the 12.1 PTR client -- live retail only, see above

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
+++ b/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
@@ -1207,6 +1207,74 @@ local function CdmFrameIsActive(frame)
     return false
 end
 
+-- Recharge state of a charge spell, split into the two cases the cooldown API
+-- cannot tell apart on its own. Returns recharging, atZero.
+--
+-- C_Spell.GetSpellCooldown reports the GLOBAL cooldown whenever the global
+-- outlasts the spell's own remaining time, so at zero charges every read of it
+-- turns into "on GCD" for the last global of the recharge. Blizzard's Cooldown
+-- Manager inherits that: isOnActualCooldown is
+--     not self.isOnGCD and self.cooldownIsActive
+-- and isOnGCD is NeverSecret, so a true value short-circuits the and before the
+-- secret operand is ever evaluated and a clean false drops out. The icon
+-- un-greys mid-recharge and the widget is re-armed with global geometry. Action
+-- bars never show this: ActionButton_ApplyCooldown keeps a SEPARATE
+-- chargeCooldown widget driven from chargeInfo.isActive.
+--
+-- Two clean signals separate the states:
+--   ci.isActive          NeverSecret; true for ANY count below max.
+--   frame.wasSetFromCharges  Blizzard's own currentCharges > 0 answer. It is
+--       computed in their UNTAINTED code and cached as a plain bool, so reading
+--       the cached field here is clean even though comparing currentCharges
+--       ourselves would be secret. True means a charge is in hand.
+--
+-- Field-captured 2026-08-09 (/euidiag chargewatch 20271, Ret Judgment): the
+-- field read clean at every sample, stayed TRUE at one charge with a global
+-- running, and tracked the count alone. It does NOT follow the global -- the
+-- comments elsewhere in this file that say otherwise predate that capture.
+--
+-- atZero is nil, NOT false, when the oracle cannot be read: our own placeholder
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
ggs it was extremely close lol

PR #1279, opened 2026-08-09
fix(actionbars): "Alpha when on CD" randomly dims ready buttons on cast

SamJin98:
Fixes https://discord.com/channels/585577383847788554/1535743290534535349

## What does this PR do?

With **Alpha when on CD** below 100 (or **Desaturate on cooldown** on), casting any spell
could randomly dim/grey buttons whose spells were **not** on cooldown, snapping back
between casts. broken since 8.6.2

**Root cause:** every dim/desat decision reduces to "is this a real cooldown or just the
GCD", and that bit came from `cdInfo.isOnGCD` read at every repaint. `isOnGCD` is only trustworthy inside a real
cooldown-event dispatch. Since 8.6.1, repaints also run from synthesized contexts - the
rate-cap trailing flush, later the cast kick, press push, hot lane, and charge-tick
refetches. A stale `isOnGCD=false` read there mid-GCD classifies a ready button as on a
real cooldown and dims it for the rest of the GCD.

**Fix:** route every real-vs-GCD decision through one classifier, `CdRealFor`:

- A trust wrapper on the dispatcher stamps real `ACTIONBAR_UPDATE_COOLDOWN` /
  `SPELL_UPDATE_COOLDOWN` dispatches; self-dispatches (cast kick, cap flushes) are marked
  and never trusted. The wrapper consumes the mark at entry and pcall-guards the inner
  dispatch, so an error cannot strand either flag.
- `isOnGCD` is read and memoized per button only under that stamp; every other context
  reuses the last trusted bit. Inactive reads and the `OnCooldownDone` edge scrub the
  memo, so staleness is at worst one event cascade late, in the no-dim direction.
- `ApplyCDAlphaAll` drops its unguarded charge/item arm (dimmed banked charges from the
  options path); trusted walks re-apply the recharge-number occlusion on class flips.

Desaturate shares the same guards, so it's fixed by the same change. Swipe pushes, charge
swipes, counts, and usability/range coloring are untouched.

## How was it tested?

Live retail, alpha 20, chain-casting on dummies: no more false dims on ready buttons;
real cooldowns still dim and restore on the falling edge, including a cooldown that
completes mid-GCD; banked-charge spells stay bright through GCDs; desaturate behaves
identically. Loads on the 12.1 PTR client without errors. `luac -p` clean.

## Screenshots

Fixes https://discord.com/channels/585577383847788554/1535743290534535349

it no longer flashes

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) - N/A, pure bug fix with no new settings

### DIFF
--- a/EllesmereUIActionBars/EllesmereUIActionBars.lua
+++ b/EllesmereUIActionBars/EllesmereUIActionBars.lua
@@ -2021,8 +2021,12 @@ local function GetOrCreateButton(slot, parent, info, index, skipProtected)
         if btn.cooldown and not EFD(btn).cdDoneHooked then
             EFD(btn).cdDoneHooked = true
             btn.cooldown:HookScript("OnCooldownDone", function(cd)
+                local b = cd:GetParent()
+                -- Displayed schedule finished: scrub the trusted memo so a
+                -- mid-GCD refetch cannot keep dimming off a stale bit.
+                EFD(b).cdRealTrusted = false
                 if EAB._RefreshCooldownVisuals then
-                    EAB._RefreshCooldownVisuals(cd:GetParent())
+                    EAB._RefreshCooldownVisuals(b)
                 end
             end)
         end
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
fix was correct but another contributor had already submitted a fix around the same time and it had a slightly smaller performance impact due to not touching the hot path, thank you samjin as always you are amazing!

PR #1283, opened 2026-08-09
zhTW: translate 26 new keys from 8.7.8 and fix the Hash at label

LoChinAn:
Only `Locales/zhTW.lua` is touched: 26 lines added and 1 existing line reworded.

The 26 additions start with the strings that landed since #1212 (8.7.6),
covering Action Bars (the new housing XP bar), Minimap and Common (Blizzard's
addon compartment button), Cooldown Manager and Resource Bars (the Fill Up /
Smooth Bar Animation work from #1271 and #1272), and QoL (Raid Tools' new
leader/assist note, the low-durability warning, and the Artisan payout hold).

Wording follows the client wherever the client has a word for it. Housing
reuses Blizzard's own zhTW: `HOUSING_DASHBOARD_NEIGHBORHOOD_FAVOR_LABEL` is
房屋經驗值 and `HOUSING_DASHBOARD_HOUSEUPGRADE` is 房屋等級, which also keeps
the new keys consistent with the existing `L["House Favor Bar"]`. The two new
proper nouns come from the live zhTW client via Wowhead: Artisan's Consortium
is 工匠聯盟 and Shard of Dundun is 敦敦裂片. The addon compartment block
follows the zhTW client's use of 插件 for addon and matches the wording zhCN
already settled on. `Fill Up` is 填滿, staying inside the existing fill family
(填滿顏色, 反向填充, 消耗填滿效果) rather than introducing a new verb, and the
tooltip that cross-references it quotes 填滿 verbatim. The longer Raid Tools
description extends an existing key, so its first clause is reused word for
word.

Three of the additions are older Resource Bars strings found by eye while
testing this batch in game, on the Class Resource Bars detail panel: `Talent`,
`Multi-band Coloring` and `Stagger Full %`. They date from early July, so an
incremental pass that only harvests newly added lines never surfaced them.
`Stagger` uses the client's own `STAGGER` (醉仙緩勁), which is also what the
existing `L["Stagger Effect"]` uses, and `Multi-band Coloring` reuses the
多段著色 already quoted by the Single-threshold tooltip so the two rows agree.

The one reworded line is `L["Hash at %1$s"]`, on the same panel. It read
`"%1$s 處的刻線"`, which composed on screen as 層數 處的刻線 -- the trailing
space after the placeholder left the particle dangling. A space there is right
when the substitution is a character name in Latin script, which is why the
Death Recap keys keep it, but this one always substitutes a Chinese word
(層數 / 百分比 / 數值). It now reads 層數的刻度線, which also switches 刻線 to
刻度線, the term every other Hash key already uses -- including the
`Hash Line Style` cog sitting on that very row.

Compiles under Lua 5.1 and round-trips through the catalog parser.

### Checklist

- [x] N/A - no new settings
- [x] N/A - locale data only, no runtime cost
- [x] N/A - locale data only
- [x] N/A - no Blizzard frames touched
- [x] Loaded in-game on live (zhTW client); the new strings render correctly in the touched modules


LoChinAn:
Because it's significantly behind the main branch, to avoid conflicts, I will open a new PR after closing this one.

PR #1286, opened 2026-08-09
fix(chat): scroll-to-bottom chevron only worked on the first tab

dfrisone:
Reported by @Boomshakalek on 8.7.7. The sidebar scroll-to-bottom chevron works on the first tab only.

**Cause:** the chevron's `OnClick` hardcoded `ChatFrame1`, and Blizzard hides every docked window except the selected one. From any other tab the click was scrolling a hidden frame.

**Fix:** resolve the target at click time with `FCFDock_GetSelectedWindow`, already the idiom in five other places in this file, and call `ScrollToBottom` (what Blizzard's own button calls) instead of driving the scrollbar directly.

**Taint:** ChatFrames are `<ScrollingMessageFrame>` intrinsics whose `secureMixin` covers `ScrollToBottom`, so the field writes stay secure. Verified in game with `issecurevariable`: true on the new path, false on the old `ScrollBar:SetScrollPercentage` one.

**Left alone, both pre-existing:** the suppressed native per-frame button is deliberate (`FCF_UpdateScrollbarAnchors` anchors the ScrollBar to it), and undocked windows fall back to ChatFrame1.

**Test:** leave tab 1 scrolled up, switch to Combat Log, scroll up, click the chevron. It jumps to bottom and tab 1 keeps its position.

Tom And Jerry Cat GIF

### DIFF
--- a/EllesmereUIChat/EllesmereUIChat.lua
+++ b/EllesmereUIChat/EllesmereUIChat.lua
@@ -4270,12 +4270,19 @@ local function SkinChatFrame(cf)
         if settingsBtn then HookIconTooltip(settingsBtn, "Settings") end
         HookIconTooltip(scrollBtn, "Scroll to Bottom")
 
-        -- Scroll to bottom
+        -- Scroll to bottom. One sidebar serves the whole docked panel, so the
+        -- target has to be resolved at click time: docked windows other than the
+        -- selected one are Hide()n by FCFDock_UpdateTabs, and scrolling a hidden
+        -- ChatFrame1 from the Guild or Loot tab does nothing the user can see.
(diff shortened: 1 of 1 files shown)


dfrisone:
Closing as superseded. The scroll-to-bottom button on main already resolves the docked selection at click time, which is what this PR was adding:

```lua
scrollBtn:SetScript("OnClick", function()
    local selected = GENERAL_CHAT_DOCK and FCFDock_GetSelectedWindow
        and FCFDock_GetSelectedWindow(GENERAL_CHAT_DOCK)
    if ECHAT.EngineScrollToBottom then
        ECHAT.EngineScrollToBottom(selected or ChatFrame1)
    end
end)
```

It also routes through the engine rather than driving the scroll bar directly, so the fix landed in a better form than this branch had. Nothing here left to merge.

PR #1287, opened 2026-08-09
Fix: Resize filter close and refresh buttons to avoid click overlap

svart2521:
Bug: https://discord.com/channels/585577383847788554/1536031139167211611

Buttons were too close to one another and caused accidental clicks on wrong elements.
Resized X button and reducing hitbox for click input.
Shifted Refresh button more to the left and increased slightly.


EllesmereGaming:
i believe ptr has changed the format of this so this is not needed, please check back in after 12.1 if it's still an issue

PR #1292, opened 2026-08-09
fix(chat): RBG raid chat swallowed by ADDON_ACTION_BLOCKED on SendChatMessage

dfrisone:
## The bug

Reported by @Boomshakalek on 8.7.8. In Rated Battlegrounds every raid-chat message fires `[ADDON_ACTION_BLOCKED] EllesmereUIChat ... SendChatMessage` at `ChatFrameEditBox.lua:307 SendText`, and the message is silently swallowed. Reporter bisection proved the EUI chat module alone is necessary and sufficient.

## The RBG part is an amplifier, not the bug

`C_ChatInfo.SendChatMessage` is `HasRestrictions = true, RestrictedForMacroChatMessages = true` (ChatInfoDocumentation.lua:564-568). It is not unconditionally protected: it refuses only a tainted caller inside a restricted context (encounter, challenge mode, PvP match, or a communication-restricted map, per SecretPredicatesDocumentation.lua:52-56).

So the taint was present in every zone all along and simply had nothing to detonate on until a match started. That is why this could never be reproduced in the open world.

## Cause

EUI's `HookScript`s on `ChatFrame1-10EditBox` run inside Blizzard's own chat-state executions, which then write the `chatType` attribute tainted. Every later `SendText` reads it back at :286 before calling `SendChatMessage` at :307.

There are two symmetric detonators:

- `Deactivate` (ChatFrameEditBox.lua:541-559) calls `SetText("")`, dispatching OnTextChanged, and `ClearFocus()`, dispatching OnEditFocusLost, and only then calls ResetChatTypeToSticky / ResetChatType.
- `ChatFrameUtil.ActivateChat` calls `SetFocus()` at :458, which dispatches OnEditFocusGained **nested**, and only then calls `UpdateHeader()` at :467, which writes chatType. In a battleground that smart-switches a sticky RAID or PARTY to INSTANCE_CHAT.

A script dispatched by a Lua-called `SetFocus()` or `SetText()` runs nested inside the caller's execution, so a post-hook still precedes everything the caller does next. Focus-gained and focus-lost had to be treated as symmetric.

## Fix

Replace the edit-box hooks at these points with EventRegistry callbacks. `CallbackRegistryMixin:TriggerEvent` runs every registrant through `securecallfunction` (CallbackRegistry.lua:201) plus `secureexecuterange` (:204), so taint stops at our closure. Registering with the same owner string replaces rather than stacks (:128-130), so repeated skin passes stay safe.

Blizzard fires `ChatFrame.OnEditBoxFocusGained`, `.OnEditBoxFocusLost` and `.OnEditBoxPreSendText` for exactly these points. All EUI focus-gained registrations now go through `ECHAT._RegisterEditBoxFocusGained`.

`OnKeyDown` and `OnChar` are left as hooks deliberately: the engine dispatches those only for physical input, so they can never land inside `SendText` or `Deactivate`.

## Accepted narrowing, disclosed in-code

Plain Up/Down recall is chat lines only now. `ParseText` runs a non-secure slash command and `ClearChat` empties the box before `OnEditBoxPreSendText` fires, so `/dance` no longer enters EUI's history. Alt+Up/Down, which uses Blizzard's own `AddHistoryLine`, still has them.

## Testing

`luac -p` clean. Verified in game by the original reporter in a Rated Battleground: raid chat sends normally and the ADDON_ACTION_BLOCKED no longer fires.

Excited Lets Go GIF by Mind Pump Media

### DIFF
--- a/EllesmereUIChat/EllesmereUIChat.lua
+++ b/EllesmereUIChat/EllesmereUIChat.lua
@@ -3736,6 +3736,77 @@ local function StyleDockManager()
 end
 
 -------------------------------------------------------------------------------
+--  Chat-state callbacks: sent-line capture + recall-cursor reset.
+--
+--  Deliberately NOT widget HookScripts on a Blizzard edit box. See the long
+--  rationale block in SkinEditBox below: an EUI closure dispatched from inside
+--  one of Blizzard's own chat-state executions leaves that execution tainted
+--  after it returns, and ChatFrameEditBoxMixin:Deactivate writes the edit
+--  box's chatType ATTRIBUTE immediately after the two script dispatches it
+--  performs (SetText("") -> OnTextChanged, ClearFocus() -> OnEditFocusLost,
+--  then ResetChatTypeToSticky/ResetChatType). A tainted chatType attribute is
+--  read back by SendText on EVERY later message, which makes every send a
+--  tainted caller of C_ChatInfo.SendChatMessage.
+--
+--  EventRegistry is the sanctioned substitute: CallbackRegistryMixin
+--  :TriggerEvent runs every registrant through securecallfunction, so our
+--  taint stops at our own closure and never returns to Blizzard's execution.
+-------------------------------------------------------------------------------
+local _chatStateCallbacksDone = false
+
+-- Blizzard fires these callbacks for EVERY chat edit box, including the temp
+-- whisper windows (11+). Deriving EUI state from one of those poisons
+-- HistoryKeeper through the secret BN_WHISPER tellTarget, which is exactly
+-- what the "idx <= 10" gate in SkinEditBox exists to prevent, so the same
+-- filter has to be re-applied inside each callback body.
+local function PermanentEditBoxIndex(eb)
+    if not eb or type(eb.GetName) ~= "function" then return nil end
+    local n = eb:GetName()
+    local i = n and tonumber(n:match("^ChatFrame(%d+)EditBox$"))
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
i actually want to hold off on this for now until after 12.1 to see if it's still an issue after the full chat rewrite goes live so lets revisit after release

dfrisone:
Revisiting this as asked after 12.1, and closing: the fix is on main.

The edit-box hooks this PR was replacing are gone, and the EventRegistry approach it proposed is what the module now does, including this branch's own helper:

```lua
function ECHAT._RegisterEditBoxFocusGained(owner, fn)
    if not (EventRegistry and EventRegistry.RegisterCallback) then return end
    EventRegistry:RegisterCallback("ChatFrame.OnEditBoxFocusGained", function(_, eb)
        if not PermanentEditBox(eb) then return end
        fn(eb)
    end, owner)
end
```

`ChatFrame.OnEditBoxFocusGained`, `.OnEditBoxFocusLost` and `.OnEditBoxPreSendText` are all registered, and there is no longer a `HookScript` on `OnEditFocusGained`, `OnEditFocusLost` or `OnTextChanged` for the edit boxes, which were the two detonators described above. The permanent-frame gate survived too, as `PermanentEditBox`.

So there is nothing left to merge here. Worth a sanity check in an actual Rated Battleground at some point, since the original report was only reproducible in one, but that is a verification task rather than this branch.

PR #1305, opened 2026-08-10
feat(qol): attach FPS/latency readout to the Secondary Stats display

delasteve:
## What does this PR do?

It allows the player to merge the FPS and Latency counters into the secondary stat display. It appears as just another stat in the stat list. Also added was an option to "compact" the stat display. Currently, it displays 2 spaces between the stat and the stat value. With the switch active, it's only one space.

## How was it tested?

12.0.5 (Live), tested the new switches (attach to secondary display, compact stat mode)

## Checklist


delasteve:
Before

image

After (no compact option selected)

image

After (compact option selected)

image

PR #1314, opened 2026-08-10
fix(qol): stop a shared ability lockout firing the Movement Alert

dfrisone:
**Cause:** casting a charge-type ability briefly locks out its siblings through a start-recovery/cooldown category (Druid Skull Bash does this to Wild Charge for ~2s). The client reports that window through the same `C_Spell.GetSpellCooldown` fields a real cooldown uses, with `isOnGCD` false, so the GCD gate missed it and the alert flashed "No Wild Charge". A `d > 0` fallback branch also re-admitted everything the duration branch's `total > 1.5` floor had rejected.

**Fix:** classify by magnitude, the only thing separating a shared lockout from a real cooldown: ignore a window under 3s when the ability's own base cooldown is longer. Applied at all three read sites and at the base-duration cache, so a lockout cannot become the value it is classified against. Blizzard classifies the same way in `Blizzard_CooldownViewer` (`MIN_GLOBAL_RECOVERY_TIME`), just at GCD length.

**Test:** verified in game on Guardian and Feral druid. Skull Bash no longer fires the alert; a real Wild Charge cast still shows the full 15s countdown; charge-based specs still hide banked charges and alert at zero.

### DIFF
--- a/EllesmereUIQoL/EllesmereUIQoL_MovementAlert.lua
+++ b/EllesmereUIQoL/EllesmereUIQoL_MovementAlert.lua
@@ -154,6 +154,25 @@ local function GetKnownCategoryDuration(spellId)
     return 0
 end
 
+-- Shared lockout vs. the ability's own cooldown. Casting one charge-type
+-- ability briefly locks out its siblings through a start-recovery/cooldown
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
merged manually

PR #1321, opened 2026-08-11
UnitFrames: Fill Color modes with value-based health coloring

nulltyto:
## What does this PR do?

Adds health-by-value coloring to the player, target, and focus unit frames, as requested in the "health-by-value coloring for player/target unitframes" feature request.

The class/custom fill swatches on the Main Frames health section become a **Fill Color** dropdown with the same modes as the Raid Frames:

- **Class Color** -- unchanged current behavior (default)
- **Classic** -- the bar blends green / yellow / red with current health
- **Custom Color** -- one flat custom color (unchanged behavior, now an explicit mode)
- **Custom Dynamic Colors** -- the bar blends between three user-chosen stops (100% / 50% / 0%), shown as inline swatches next to the dropdown. Picking a dark 100% stop gives the requested dark-at-full style.

Dark Mode stays the module-wide master and overrides the row, exactly as it overrode the old swatches. The gradient applies only in the Class / Custom modes; the Gradient End Color picker moves into the gradient cog, and the cog rows disable in the value-based modes. The Bar Color sync icon copies the mode and the three stops.

Profiles that predate the dropdown resolve their mode from the legacy keys (`healthClassColored`, `customFillColor`), so existing setups load with identical colors and the dropdown shows the matching mode.

**Secret-value safety:** the two value-based modes feed a cached `C_CurveUtil` color curve to `UnitHealthPercent`, the same mechanism the Raid Frames Custom Dynamic Colors use. Lua never reads the health percent; the possibly-secret RGB goes only into `SetStatusBarColor` and never enters the gradient path or any comparison.

## How was it tested?

Live retail 12.0.7 (68974), in dungeon content (trash pulls and bosses):

- Custom Dynamic Colors on the target frame with dark/yellow/red stops: the color tracks damage in combat, including while health is a secret value.
- Mode switches between Class, Custom, and Custom Dynamic mid-session; options previews follow the dropdown.
- CPU profiled with `C_AddOnProfiler` recordings through the same boss in Custom Dynamic and Class modes: 0.075 vs 0.080 ms/frame in combat for the whole module -- within the same-session run spread, no measurable cost from the curve path.

Not yet run on the 12.1 PTR client.

## Screenshots

(To be added.)

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) -- the resolved default mode is Class Color, the current behavior
- [x] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built -- the default path adds two table reads per color update, nothing else
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations) -- the curve modes run only on health events; `UnitHealthPercent` returns one small ColorMixin per event, replacing the class-color lookup on the same path
- [x] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames -- no Blizzard frames touched
- [ ] Tested in-game, works on live retail; no load errors on the 12.1 PTR client -- tested on live retail; PTR run still pending

### DIFF
--- a/EllesmereUIUnitFrames/EUI_UnitFrames_Options.lua
+++ b/EllesmereUIUnitFrames/EUI_UnitFrames_Options.lua
@@ -1009,13 +1009,25 @@ initFrame:SetScript("OnEvent", function(self)
             -- Boss preview always renders as hostile-red since the real boss
             -- frame never class-colors (no player class).
             local cFill = settings.customFillColor
-            local isClassColored = settings.healthClassColored and unitKey ~= "boss"
-            if isClassColored then
+            local fillMode = ns.GetUnitHealthColorMode(settings)
+            -- The mini/pet frames have no mode dropdown and resolve "class" as
+            -- the fallback, so their class preview still follows the raw
+            -- healthClassColored key (pet defaults to the green fallback and
+            -- ToT/FoT to hostile red, matching their live frames).
+            local isSharedUnit = unitKey == "player" or unitKey == "target" or unitKey == "focus"
+            local isClassColored = fillMode == "class" and unitKey ~= "boss"
+                and (isSharedUnit or settings.healthClassColored)
+            if fillMode == "classic" or fillMode == "customDynamic" then
+                -- Value-based fill at the preview's fake health percent.
+                hR, hG, hB = ns.ResolveHealthByValueColor(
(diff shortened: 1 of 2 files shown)


Taum:
This has been also implemented by https://github.com/EllesmereGaming/EllesmereUI/pull/1513

PR #1330, opened 2026-08-11
Fix Blizzard-friendly nameplate hitboxes

Nnoggie:
- Restore Blizzard native plate geometry when EUI relinquishes friendly player nameplates.
- Explore per-nameplate hit-test ownership for EUI enemy and friendly plates, plus runtime size ownership handoff.
- WIP: sizing and mouseover behavior still require in-game investigation.

### DIFF
--- a/EllesmereUINameplates/EllesmereUINameplates.lua
+++ b/EllesmereUINameplates/EllesmereUINameplates.lua
@@ -3175,6 +3175,41 @@ function ns.RefreshStackingMotion()
     end
 end
 
+function ns.ApplyEnemyHitbox(plate)
+    if not plate then return end
+    local db = p or defaults
+    plate:SetSize(
+        GetHealthBarWidth() * ((db.hitboxScaleX or 100) / 100),
+        GetHealthBarHeight() * ((db.hitboxScaleY or 100) / 100))
+    local np = plate.nameplate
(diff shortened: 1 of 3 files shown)


Nnoggie:
## WIP investigation update

### Confirmed finding
EUI was applying enemy hitbox geometry through the global native nameplate size API even when Blizzard owned friendly player nameplates. Blizzard's friendly visuals could then extend beyond the native frame used for hit testing, leaving the visible plate unable to receive click or mouseover targeting. Restoring Blizzard's native geometry restored clickability in the tested live 8.7.8 configuration.

### Progress so far
- Added per-nameplate hit-test bindings so EUI enemy hitboxes can remain tied to EUI geometry instead of the shared native frame.
- Added an equivalent binding/reset path for EUI full friendly plates.
- Added runtime size ownership switching: EUI keeps its size while it owns friendly plates, while Blizzard size events and the prior Blizzard `nameplateSize` value are restored when EUI relinquishes them.

### Still unresolved
- EUI enemy nameplates and Blizzard friendly nameplates are not sizing consistently.
- Their geometry does not correctly follow the Blizzard Nameplate Size setting in every ownership/configuration path.
- Friendly mouseover behavior still needs investigation with the new per-plate bindings.

Keeping this PR in draft. More in-game validation is needed before review.

PR #1333, opened 2026-08-11
feat(unitframes): configurable weapon enchant placement (In Line / Above / Below)

dfrisone:
**Cause:** Weapon enchants always occupied the bar's first cells via a container shift. That placement is fine (it is the default UI's model), but it briefly looked broken because the ComputeGrid trailing-gap off-by-one (#1335) stacked on top of it, and there was no way to move the enchants for users who wanted uniform rows.

**Fix:** Per-bar three-way placement on the default Buffs bar. In Line (default) keeps the original model: first cells of the top row, container shift restored, in-combat cell-collision hide restored for this mode only. Above/Below put them on a dedicated cross-axis line with Offset X/Y in the cog (offsets are zeroed In Line, where the buttons sit in reserved cells). Below anchors from the vertically mirrored corner. The shift always re-seats, so mode flips and the last oil expiring reset it cleanly.

**Test:** Default (In Line): apply an oil, enchants take the top row's first spots and buff rows share an edge (needs #1335). Switch to Above/Below, confirm the dedicated line and working offsets, then flip back and confirm the shift resets. Drop an enchant mid-combat: its icon hides rather than overlapping the run, and regen repacks.

### DIFF
--- a/EllesmereUIOptions/EUI_PlayerAuraBars_ManagerPages.lua
+++ b/EllesmereUIOptions/EUI_PlayerAuraBars_ManagerPages.lua
@@ -68,6 +68,11 @@ local FONT_OUTLINE_VALUES = {
     outline = "Outline", thick = "Thick Outline",
 }
 local FONT_OUTLINE_ORDER = { "default", "none", "outline", "thick" }
+-- Cross-axis side for the weapon-enchant line. "Above"/"Below" read naturally for the
+-- horizontal growth directions; under Up/Down growth the same two values mean the side
+-- before/after the columns, mirrored the same way.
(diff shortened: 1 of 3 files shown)


dfrisone:
Closing: field-tested and rejected. Moving the enchants to their own line above the grid pushes them off the top of the screen, since buff bars normally sit at the top edge, and it removes the adjustability they had as in-grid cells. Reverting to the in-row model; the wrapped-row inset it was meant to fix needs a different approach.

dfrisone:
Reopening: field test confirms the buff rows DO align correctly with this change. The only real problem is placement -- the line renders above the grid, and buff bars typically sit at the top of the screen, so the enchant icons go off the top edge (only their duration text stays visible). Following up with position controls rather than reverting the alignment fix.

PR #1362, opened 2026-08-12
fix(quickdraw): classify toys dragged from the Toy Box as kind=toy

Barbiero:
## What does this PR do?

Blizzard's Collections/Toy Box frame hands GetCursorInfo() back with cursorType "item" instead of "toy" for dragged toys. SlotFromCursor stored these as kind="item", which runs SlotUsability through C_Item.IsUsableItem -- always "unusable" for a toy -- causing the icon to stay dimmed even though it is usable. Entries created via the "change action" search picker were unaffected since it always builds toy entries with kind="toy" directly.

Reclassify at cursor-read time via PlayerHasToy(), and extend the existing per-profile migration pass to fix already-saved entries too.

## How was it tested?

In-game, dragging and dropping a toy from the collections window into the quickdraw action menu.

If you do this on the live version, you´ll have an "item" entry:
image
and it will be dimmed(compare to the other entry that is a kind="toy" version):
image

After my change, this will now properly be a "toy", so it won't be dimmed (and broken entries are migrated properly as well, though not sure if there might be a side effect for doing this migration).

## Checklist

- [X] Tested in-game, works on live retail; no load errors on the 12.1 PTR client

### DIFF
--- a/EllesmereUIQuickdraw/EllesmereUIQuickdraw.lua
+++ b/EllesmereUIQuickdraw/EllesmereUIQuickdraw.lua
@@ -406,6 +406,15 @@ local function MigrateNames(p)
         -- carries none, so the copy is simply dropped.
         for _, slot in pairs(palette.slots or {}) do
             if slot.kind == "palette" then slot.name = nil end
+            -- Toys dragged in from the Collections/Toy Box frame before
+            -- SlotFromCursor reclassified them landed here as kind="item",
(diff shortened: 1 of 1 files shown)


Barbiero:
I didn´t  touch any locale keys so I'm suprised that the gh action failed there, but I have them re-generated in #1360

Barbiero:
rebased with latest main so now we don't have locale-keys conflicts :smile:

EllesmereGaming:
thank you!!

PR #1370, opened 2026-08-12
fix(movement alert): in-combat countdown, and the Skull Bash lockout when the window is secret

dfrisone:
### Cause
Two failures, both from a value being unreadable in combat.

`C_Spell.GetSpellCooldownDuration`'s second argument is `ignoreGCD` and it defaults to false, so while a global cooldown is running it describes the GCD rather than the spell's own cooldown. A GCD-length total is rejected downstream, leaving no duration object, and that object is the only thing that can draw a number the client keeps secret. Out of combat no GCD runs, which is why the timer appeared the moment the fight ended.

Separately, `IsLockoutWindow` classifies a cross-ability lockout by magnitude against the ability's own base cooldown, but it cannot run at all when the window is secret, so the ~2s Skull Bash lockout was alerted as a real Wild Charge cooldown.

### Fix
Ask for the real cooldown, and fall back to the other duration API so a spell cached with the wrong charge flag still gets an object. For the secret case, evaluate a `C_CurveUtil` step curve against the total: 0 below the threshold, 1 above, compared in C and fed to `SetAlpha`, never read. Same technique the charge visibility path already uses. Base resolution is shared between both classifiers so they cannot disagree.

### Test
Verified in game: Skull Bash as an opener and mid-dungeon no longer alerts, a real Wild Charge cooldown still shows, Death's Advance shows its in-combat countdown again, and decimals render with Show Decimal on.


EllesmereGaming:
manual fix

PR #1373, opened 2026-08-12
fix(raid frames): let an aura show duration text with no cooldown swipe

dfrisone:
### Cause
Two things blocked the text-only aura setup on Party/Raid frames.

`Cooldown:SetShown(false)` does not stick on an aura button: the engine calls `SetCooldown` on that frame whenever the slot's data refreshes, and the native call implicitly re-Shows it. So `hideSwipe` was written and then overruled, from both Hide Icons and Duration Swipe.

On top of that, Duration Swipe was greyed out whenever Hide Icons was on, so the one control that could have helped was unreachable in exactly the configuration that needed it.

### Fix
Use `SetDrawSwipe`, a CooldownStyle aspect the engine has no reason to clear on refresh (and `AllowedWhenTainted`, so the call is legal in restricted content). The cooldown stays registered via `SetDurationCooldown` because that registration is the button's duration SOURCE -- withholding it removes the duration text along with the swipe. Ungate the toggle; Opacity and Border keep their gating, since those describe icon art.

### Test
Verified in game on party frames: hidden icon with duration text, stacks and no swipe, and the swipe returns when re-enabled.

Greatest Of All Time Goat GIF by Nutrena Feed

### DIFF
--- a/EllesmereUIRaidFrames/EUI_RaidFrames_BuffManager.lua
+++ b/EllesmereUIRaidFrames/EUI_RaidFrames_BuffManager.lua
@@ -4166,8 +4166,13 @@ function ns.BM_BuildPage(pageName, parent, yOffset)
             end
 
             local durRow = SettingsRow(
+                -- NOT gated on Hide Icons, unlike Opacity and Border above. Those
+                -- describe the icon art and are meaningless without it; the swipe is a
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
merging this manually as i was doing similar work in that area

PR #1375, opened 2026-08-12
Fix Player Aura Bars filtered bars rendering every buff when the engine identity gate is degraded

dfrisone:
Reported: External Defensives (Externals filter) shows unrelated buffs at login until a /reload, after entering a vehicle, and after an unlock-mode round trip.

**Cause:** Blizzard_AuraContainerUtil.DoesAuraPassCandidateFilters only applies includeSpellIDs/excludeSpellIDs when CanApplyIdentityCandidateFilters passes, and for a helpful group that gate is `isHelpful and not UnitCanAssist("player", unit)`. Any moment the player is not assistable to itself (unit flags not streamed at login, control handed to a vehicle) parses the whole helpful set into a bar that asked for four spells, and group membership is cached per aura instance, so it sticks until a reload.

**Fix:** re-run the group config for every live container after those edges. SetAuraGroupCandidateFilters calls UpdateAllAuras, which marks FullAuraRebuild and re-evaluates membership with the gate settled. Edges: hooked container OnShow, container creation (login), PLAYER_ENTERING_WORLD, PLAYER_CONTROL_LOST/GAINED and vehicle enter/exit. Coalesced one-tick pass, no polling; the cinematic handler this replaces did the same thing for one edge.

**Test:** logged in on a druid, mounted and took a vehicle, entered and exited unlock mode: the Externals bar keeps only its filtered buffs.

### DIFF
--- a/EllesmereUIUnitFrames/EllesmereUIUnitFrames_PlayerAuraBars.lua
+++ b/EllesmereUIUnitFrames/EllesmereUIUnitFrames_PlayerAuraBars.lua
@@ -2150,6 +2150,7 @@ local function ApplyLiveConfig(isBuff)
             local _, spec = BuildContainerSpec(parent, cfg, grid)
             AK.RequestContainer(parent, "player", spec, function(newContainer)
                 buffsContainer = newContainer
+                WatchContainerVisibility(newContainer)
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
fix is in

PR #1376, opened 2026-08-13
fix(aurabuff): hide the pet reminder while mounted

Trenchfoote:
Mounting in the open world auto-dismisses the pet, so the "pet not summoned" reminder appeared for the whole ride and cleared itself on dismount when the pet returned. Skyriding already hid it, because the reminder pass suppresses everything while mounted and flying, which made the warning look inconsistent as well as wrong.

A pet cannot be summoned while mounted in any case, so the reminder is never actionable there. This matches the Pet on Passive reminder in the same block, which already skips while mounted for the same reason.

PLAYER_MOUNT_DISPLAY_CHANGED is already registered, so mounting and dismounting re-evaluate the reminder with no new events.

## Checklist


EllesmereGaming:
fix is live!

PR #1413, opened 2026-08-13
fix(chat): refuse clubTicket links in the link overlay

dfrisone:
**Cause:** the link takeover forwards clicks with a direct `SetItemRef`. Blizzard's clubTicket handler ends at `CommunitiesHyperlink.OnClickLink`, which lazily creates its `CLUB_TICKET_RECEIVED` listener and `SetScript`s the OnEvent. Reached from there that `SetScript` runs tainted, so the handler stays tainted for the session and its first statement, `C_Club.GetLastTicketResponse`, is protected.

**Effect:** one click on a community invite made every ticket arriving afterwards throw `ADDON_ACTION_FORBIDDEN` against EllesmereUIChat, including tickets Blizzard's own frames requested.

**Fix:** refuse `clubTicket` before the listener is created. Forwarding gained nothing, since the response is unreadable from an insecure path either way. `LinkBlockedMsg` now takes the string to show.

**Test:** with Shortened Channel Names on, clicking a community invite shows the message and throws nothing. Turn the setting off and invites work through Blizzard's zones as before.

### DIFF
--- a/EllesmereUIChat/EllesmereUIChat_Engine.lua
+++ b/EllesmereUIChat/EllesmereUIChat_Engine.lua
@@ -273,8 +273,10 @@ end
 --      FCF_OpenTemporaryWindow taint cascade). Link classes whose SetItemRef
 --      branch writes editbox/tell state are blocked in protected content
 --      with a user-visible message; secret links (lockdown lines) are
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
was fixed earlier by juju who forgot to mark it on the site MADGI

PR #1419, opened 2026-08-14
fix(raid frames): draw Health Bar Color tints with the health bar texture

### LINKED ISSUE #1339: Raid frames: health bar color indicators overwrite bar texture
"Health Bar Color" indicators on raid frames have flat (no) textures. They should use the health bar's configured texture.

KogasaPls:
Closes #1339.

## What does this PR do?

A "Health Bar Color" indicator painted a flat colour slab over the health fill, so a
tinted frame lost the bar's texture and read as a solid block beside untinted ones.
It now borrows the bar's own fill texture - whatever Health Bar Texture resolves to,
SharedMedia keys included - and recolours that, so a tinted frame keeps the same
shading as its neighbours.

The overlay is anchored to the fill texture, so it tracks the bar's geometry for
free: no polling, no timer, no hook, and no per-tick work.

It also inherits the bar's configured **Fill Opacity**, so a tinted frame is as
translucent as its untinted neighbours. Dark Mode is deliberately not inherited -
that dims the bar automatically, and a colour the user picked should not be
auto-dimmed with it.

Overlays register on the bar so a Health Bar Texture change re-anchors and repaints
them instead of leaving them on the old art: `ReanchorAbsorbToFill` for frames built
through `StyleButton`, `FB.ApplyStyle` for Focus/Boss buttons, which have no absorb
cluster and so never build one. Registrations are dropped when the container that
owns them goes - engine aura buttons are never freed - and rebuilt by the next paint.

Routed through it: the Buff Manager's healthcolor slots, the Debuff Manager's
healthcolor tile, and both options previews. Background Color deliberately keeps its
flat tint, since it covers the health bar's background, which is itself a flat fill.

## How was it tested?

Live retail (12.1). The overlay reports the same texture fileID as the bar's fill and
renders with its shading; a Health Bar Texture change repaints; vertical fill renders
correctly; Fill Opacity is respected and Dark Mode is not.

One behaviour was measured rather than assumed, because the design rests on it: these
fills clip by *resizing* the fill texture, not by rewriting its tex coords -
`GetTexCoord()` reads identical at full and at half fill. That is what makes
anchoring to the fill sufficient with nothing to update per tick. It is a statement
about these bars, whose fills are plain file textures, rather than about StatusBars
generally.

No IS_121-gated paths are touched.

## Screenshots

Control (textured health bar with no Health Bar Color indicator):

image

Before:
image

After:

image

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) - N/A, no new
      settings; this is a bug fix to an existing indicator's rendering
- [x] Zero cost while disabled: no events registered, no polling, no hooks doing
      work, no frames built - a bar with no healthcolor overlay never allocates a
      registry and the refresh returns immediately
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no
      per-frame allocations) - no OnUpdate, no timers, no per-call allocation; the
      registry is walked once per button on a reload or style pass, never per frame,
      per health tick or per aura update
- [x] No writes onto Blizzard-owned frames (weak-table pattern used);
      `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames -
      every frame and texture touched is ours; no hooks added
- [x] Tested in-game, works on live retail; no load errors on the 12.1 PTR client -
      verified on live retail (12.1); the PTR half no longer applies now that 12.1
      has launched

### DIFF
--- a/EllesmereUIOptions/EUI_RaidFrames_ManagerPages.lua
+++ b/EllesmereUIOptions/EUI_RaidFrames_ManagerPages.lua
@@ -2067,8 +2067,8 @@ function ns.DMP_RefreshPreview()
                         pv._dmHC = hc
                     end
                     local c = t.color or { r = 1, g = 0.25, b = 0.25 }
-                    hc:SetColorTexture(c.r or 1, c.g or 0.25, c.b or 0.25,
-                        (t.opacity or 45) / 100)
+                    ns.RF_TintOverBarFill(hc, health, c.r or 1, c.g or 0.25,
+                        c.b or 0.25, (t.opacity or 45) / 100)
                     hc:Show()
                 end
             elseif t.type == "glow" then
--- a/EllesmereUIRaidFrames/EUI_RaidFrames_AuraContainers.lua
+++ b/EllesmereUIRaidFrames/EUI_RaidFrames_AuraContainers.lua
@@ -1378,7 +1378,14 @@ local function BmApplyEffect(button, dd, style)
     if ind.type == "healthcolor" or ind.type == "bgcolor" then
         if dd.tex then
             local r, g, b = BmColor(ind.color, 0, 1, 0)
-            dd.tex:SetColorTexture(r, g, b, (ind.opacity or 100) / 100)
+            local a = (ind.opacity or 100) / 100
+            -- healthcolor rides the fill and borrows its texture; bgcolor covers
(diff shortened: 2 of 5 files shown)


EllesmereGaming:
second PR both very clean fixes, excellent work!

PR #1422, opened 2026-08-14
feat(macrofactory): add marker, ping and announce toggles to Set Focus

wanghws:
## What does this PR do?

Set Focus produced a bare `/focus` line and nothing else. Three optional lines now hang off
it, each behind its own toggle in the macro's right-click menu, and all three default OFF --
an untouched macro body stays byte-for-byte what it was.

| Toggle | Line added |
|---|---|
| Auto Mark Focus | `/tm [@focus] ~N` -- marks the focus with the raid marker you pick |
| Ping Focus | `/ping [@focus] onmyway` -- pings it for the group |
| Announce Focus | `/p Focus: {rtN} %f` -- announces it in party chat |

They share one leading `/stopmacro [@focus,noexists]`, so none of them act on the current
target when no focus was acquired. `~` is the 12.0.7 macro extension that declines to
overwrite a marker a group member already placed, and `%f` is the built-in chat substitution
for the focus unit's name. The announce line only names the marker while Auto Mark is on, so
it never advertises one nothing set.

The right-click menu gains a generic `extraToggles` / `extraChoices` pair any macro def can
reuse. Toggles copy the Show Tooltip row's shape but default off, storing `nil` rather than
`false` so an untouched setting leaves no key in SavedVariables. The marker picker is a
dropdown parented to the menu (so it closes with it) and stays usable whichever toggles are
on, since it only decides what Auto Mark writes.

## How was it tested?

a

## Screenshots
b

## Checklist

### DIFF
--- a/EUI_MacroFactory.lua
+++ b/EUI_MacroFactory.lua
@@ -139,6 +139,24 @@ end
 DYNAMIC_HEALTH_RECOVERY]]
 
 
+-------------------------------------------------------------------------------
+--  Set Focus extras -- auto raid marker, ping and group announce
+--
+--  All three are plain macro lines appended by the def's extraBody below, and
+--  all three default OFF. Nothing here registers an event or creates a frame:
+--  the game runs the macro, so a disabled toggle costs exactly nothing.
+-------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
this is awesome work thank you!

PR #1424, opened 2026-08-14
fix(editmode): keep the preset layouts in every C_EditMode.SaveLayouts payload

dfrisone:
**Cause:** `C_EditMode.SaveLayouts` replaces the character's entire layout set, and Blizzard always passes it `presets .. saved` with `activeLayout` an index into that merged list. `C_EditMode.GetLayouts` returns only the saved half, so `SyncEditModeIconCounts` wrote back a short list whose indices no longer matched the `activeLayout` travelling with it. It runs from `RegisterOnHide` on every options-window close.

**Fix:** rebuild the merged list the way `EditModeManagerFrame:UpdateLayoutInfo` does before saving, skip the write when the presets cannot be resolved (the CDM writer merged but never bailed), leave the read-only preset entries untouched, and never write while Blizzard's Edit Mode is open.

**Test:** with several custom Edit Mode layouts, open and close the EUI options window repeatedly, `/reload`, and confirm the layout list, the active layout and the CDM viewers are intact; then disable EUI and confirm the Essential and Utility viewers still render.


dfrisone:
Confirmed live on a 12.1.0 client:

```
/dump #C_EditMode.GetLayouts().layouts, C_EditMode.GetLayouts().activeLayout, #EditModePresetLayoutManager:GetCopyOfPresetLayouts()
[1]=3, [2]=5, [3]=2
```

2 presets + 3 saved layouts, `activeLayout` 5, and the getter returns a 3-entry list. The pre-fix payload was `SaveLayouts{layouts = <3 entries>, activeLayout = 5}`, an index two slots past the end of the array it ships with.

PR #1435, opened 2026-08-14
fix(aurakit): keep the dispel ring opaque above icon size 34

JuJuFX-dev:
## What does this PR do?

Fixes the dispel-type ring on aura icons rendering semi-transparent at larger icon sizes, which made the engine's dispel color visibly blend with the user's own border color instead of replacing it. Reported on Player Aura Bars debuffs, on both the default Debuffs bar and custom bars, at Icon Size 36 and above and most obvious around 50: a red 1px border plus a Magic debuff produced a muddy purple ring, with only the four corners showing the true dispel color.

The ring art is ours (`media/textures/square-ring.png`), registered with the engine purely as a tint target, because per-aura dispel data is secret and the show/hide and coloring decisions belong to Blizzard. To hold the ring at the user's physical-pixel border thickness instead of letting it scale with the icon, AuraKit crops the source window rather than stretching the texture. The band that survives that crop measures `C*t / (2*(s - 2t))` texels for canvas `C`, thickness `t` and drawn size `s`, and on the old 64px canvas it fell below a single texel once the drawn size passed `34 * t`. Under one texel the bilinear filter samples across the band's inner edge into the transparent interior, so the ring renders at alpha below 1 and the static border underneath bleeds through. The corners stayed correct because the band covers both axes there, which is exactly the artifact in the report.

The fix re-cuts `square-ring.png` at 256x256 with a 64 texel band. The band-to-canvas ratio stays 1/4, so the cropping math is completely untouched and the literal 4 in the formula still holds; the added texel density is the entire fix. The band now keeps at least 2.2 texels across the full 16 to 60 Icon Size range, so the ring stays fully opaque throughout. The only Lua changes are comments: they record the canvas resolution as a correctness constraint with the condition to re-check if the Icon Size slider is ever raised past 60, and drop the previous claim that the cropped band stayed solid at any icon size.

Since the ring lives in shared AuraKit, this also covers every other consumer that can hit the same sizes with dispel coloring enabled: Unit Frames, Raid Frames and Nameplates. Their rendering is unchanged below the old threshold, because the ratio and therefore the computed thickness are identical.

### Relationship to the other open PR

This is the minimal fix for the reported bug. PR #1436 solves the same bug a second way, by rebuilding the ring out of four solid strips so that texel sampling stops being involved at all, and the two cannot both be merged: they change the same block of `ApplyStyleToRegions`. They are offered side by side deliberately so the choice is yours rather than mine. This one is 18 changed lines and one re-cut asset, with no logic touched and nothing new to review in the restricted-execution path. It raises the failure threshold from icon size 34 to roughly 66 with comfortable margin, which clears the current 16 to 60 slider but is still a threshold rather than a removal. If the Icon Size maximum is ever raised past 66, this needs revisiting and the strip PR does not.

## How was it tested?

Live retail (Midnight 12.1). Reproduced on a Player Aura Bars debuff bar at Icon Size 50 with a 1px red border and a Magic debuff, on both the default Debuffs bar and a custom bar, then confirmed fixed with the new asset: the ring renders in the flat dispel color with no bleed-through from the border, and the ring thickness is unchanged at 1px. Sizes at and below 32, which never showed the artifact, look identical to before.

Note for anyone verifying this locally: a `/reload` is not enough, because the client keeps the loaded texture in memory. A full client restart is required to pick up the re-cut asset, otherwise the old 64px version keeps rendering.

## Screenshots

Before: Icon Size 50, 1px red border, Magic debuff. The ring reads as a blend of red and the Magic dispel color, with only single corner pixels showing the true color.

After: same settings, the ring renders in the flat Magic dispel color at the same 1px thickness.

## Checklist

  - N/A. No new settings. This is a bug fix to an existing, already opt-in display, which the criteria explicitly allow to change behavior without a new opt-in.
  - Nothing added. No new events, hooks, frames or code paths of any kind; the Lua diff is comments only.
  - Unchanged. The same single texture per button, the same change-guarded `SetTexCoord`, and the same crop math as before. The larger canvas costs texture memory only, 256 KB against 16 KB uncompressed for one shared asset.
  - The ring is a texture we create and own on our own holder frame, handed to the engine through its own `AddDispelTypeTexture` API. No Blizzard frame is written to and no script is replaced. Per-button state stays on AuraKit's existing side table.
  - Verified on live as described above. No version gates, no API changes at all; the existing 12.1 dispel-type texture path is untouched.


EllesmereGaming:
went with the refactor, good shit

PR #1438, opened 2026-08-14
Expand Player Aura Bars size limits

Nnoggie:
Raises Player Aura Bars base and per-filter icon size limits to 400, and duration and stack text size limits to 60.
Because of reasons.

image


EllesmereGaming:
400 u absolute freak

PR #1444, opened 2026-08-14
fix(qol): movement alert dies on an undefined ns

dfrisone:
**Cause:** `RepositionBuffAlertHost` stored its slot count on `ns`, a name EllesmereUIQoL_MovementAlert.lua never captures from the addon varargs, so the write threw every render pass once a buffActive spell (Burning Rush) built the host. The throw aborted `CheckMovementCooldown` before the show/refresh block, so no alert rendered and the host kept its provisional CENTER seat.

**Fix:** keep the count in a file-local alongside the other buff-alert handles.

**Test:** warlock with Burning Rush enabled in Movement Alerts, activate it in and out of combat: alert shows and the icon seats at the alert stack, no Lua error.

### DIFF
--- a/EllesmereUIQoL/EllesmereUIQoL_MovementAlert.lua
+++ b/EllesmereUIQoL/EllesmereUIQoL_MovementAlert.lua
@@ -1088,14 +1088,17 @@ end
 -- universal off-path -- can park the host; defined in the lane block below).
 local buffAlertHost, buffAlertBuilt, buffAlertRegenArm, buffAlertLastCount
 
-local function HideMovementDisplay()
+-- keepBuffLane: the cooldown display is going away but the buffActive lane is
+-- not. A buffActive spell takes no slot, so an empty cooldown stack is its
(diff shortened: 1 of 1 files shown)


dfrisone:
Awaiting the reporter to confirm this works in game before merge. Test build sent: 8.8.7-mafix1.

dfrisone:
Second cause found, pushed as cc5610ac. The `ns` throw was masking a real defect: a buffActive spell takes no display slot, so the cooldown stack is empty in its normal state, and the empty-stack branch called `HideMovementDisplay`, parking the buff-alert host that `EnsureBuffAlertLane` had shown seven lines earlier in the same pass. The alert could only ever appear while some other movement spell happened to be on cooldown. Before the first fix the throw aborted the pass before that line, which is the only reason the icon showed at all.

The empty-stack branch now keeps the lane's own verdict. Still awaiting reporter confirmation; test build 8.8.7-mafix2.

PR #1448, opened 2026-08-14
fix(player castbar) Hide border for the castbar when borderSize is 0

toanqc:
## What does this PR do?

Bug report: https://discord.com/channels/585577383847788554/1537251571593912371

Even when set the border size = 0 for the cast bar, when casting something, the border still shows. This PR hides the border when borderSize is 0

## How was it tested?

Tested in-game with multiple settings with border size is 0 or > 0.
Also enable the show Icon to make sure everything works fine without breaking.

## Screenshots

With border size = 0
after0

With border size > 0
after1

## Checklist


EllesmereGaming:
fixed this manually as there was an edge case regression in the PR, live with next patch thank you!

PR #1450, opened 2026-08-14
ptBR: Translate QOL, Aura Buff reminders, Resource Bars, etc 

Barbiero:
## What does this PR do?

A ton more strings translated, next I'll focus on global settings & cross-module strings, and leave Data Bars for last as that's a bit complex

## How was it tested?

in game with ptBR client

## Checklist


Barbiero:
Proud to announce that we now have every module and screen localized to ptBR  where possible! :brazil:  :tada: 

Of course, I might have missed some places. I'm only human, and telling the AI to do an actual full sweep of the entire addon is more than I can pay for. And also error prone(which is why I do manual checks and iterate on it).

PR #1453, opened 2026-08-14
Feature - Adjust Strata Level of Raid Frames

bchancel:
## What does this PR do?

Raid and party frames currently use the `LOW` frame strata, which means Blizzard spell auras can obscure them.

This PR adds a **Frame Strata** dropdown under **Raid Frames > Frames > Extras** with the following options:

- Background
- Low (Default)
- Medium
- High
- Dialog

The default remains `LOW`, preserving existing behavior. Party frames inherit the raid setting unless the Party Extras section is unsynced.

The change affects only raid and party frame containers. It adds no polling, timers, hooks, event registrations, or runtime frames.

## How was it tested?

Tested in-game on Retail/Midnight 12.1.

- Confirmed the dropdown functions under Raid Frames > Frames > Extras.
- Confirmed `Low (Default)` preserves the existing frame layering.
- Reproduced Blizzard spell auras rendering above frames at `LOW`.
- Confirmed changing the strata to `MEDIUM` places the frames above those auras.
- Verified stacked elements like debuffs render properly at different strata levels

### Before and after

| Low (Default) | Medium |
|:---:|:---:|
| Blizzard aura overlapping raid frames at Low strata | Raid frames above the Blizzard aura at Medium strata |

### Frame Strata dropdown

Frame Strata dropdown under Raid Frames Extras

## Checklist

- [x] New settings default **OFF** **N/A:** this is a non-boolean setting that defaults to the existing `LOW` behavior, so there is no behavior change without opt-in.
- [x] Zero cost while disabled: no new events, polling, hooks, or frames are added.
- [x] Cheap while enabled: changes are event-driven with no polling, timer-based logic, or per-frame allocations.
- [x] No writes onto Blizzard-owned frames; only EllesmereUI-owned containers are modified.

### DIFF
--- a/EllesmereUIOptions/EUI_RaidFrames_Options.lua
+++ b/EllesmereUIOptions/EUI_RaidFrames_Options.lua
@@ -35,6 +35,13 @@ initFrame:SetScript("OnEvent", function(self)
     local ReloadFrames = ns.ReloadFrames
     local floor = math.floor
 
+    -- Keep shared labels unchanged while marking this dropdown's default.
+    local rfStrataValues = {}
+    for key, label in pairs(EllesmereUI.FRAME_STRATA_LABELS) do
+        rfStrataValues[key] = label
+    end
+    rfStrataValues.LOW = rfStrataValues.LOW .. " (Default)"
+
     local function GetOutline()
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
good idea thank you!

PR #1456, opened 2026-08-14
Improve heal prediction logic

svart2521:
Bug: https://discord.com/channels/585577383847788554/1537628410749386823

Issue: With Absorb Style = none, a gate deep in UpdateAbsorb() did ab:Hide(); return before the function ever reached the heal-prediction paint code further down — so heal prediction silently never ran, no matter how it was configured

Fix: that return now only fires if heal prediction is also off; otherwise execution falls through so its paint block still runs. (Also folded in the two earlier fixes — the outer all-off guard, and the incoming-heals value being excluded from the repaint cache)


Barbiero:
@EllesmereGaming @svart2521 seems like this PR introduced a syntax error: L2584 is missing an "if" statement, not sure if the condition was intended to be removed(so the equivalent `end` needs to be removed too), or if it should be restored

EllesmereGaming:
yeah @Barbiero I fixed this one locally and then updated the full git with the new patch update. sorry for the temporary syntax issue it was only for users that were pulling directly from git

PR #1460, opened 2026-08-15
Fix combat item comparison tooltips

jixinliu666:
## What does this PR do?

Bug fix only. This PR fixes item comparison tooltips remaining visible during combat when EllesmereUI's existing `Show Tooltips` mode is set to `Out of Combat` or `Out of Boss Combat`.

This edge case is easy to encounter because active-use trinkets are commonly placed on action bars. When a player clicks or moves the mouse over an active trinket button during combat, Blizzard's action-button tooltip path can create item comparisons even though EllesmereUI is configured to show tooltips only outside combat. The issue is therefore most visible on action-bar trinkets and may not be noticed when hovering equipment in the character panel.

The fix:

- Clears shopping comparison tooltips when they are shown or anchored while the primary tooltip is currently suppressed.
- In `Out of Combat` mode, clears existing shopping comparison tooltips when combat starts; in `Out of Boss Combat` mode, clears them when a boss encounter starts.
- Defers cleanup outside protected tooltip and action-button call stacks.
- Restores comparison behavior after combat or an encounter ends and the item is hovered again.
- Installs comparison hooks lazily on first use of a non-default mode; registers transition events only for `Out of Combat` and `Out of Boss Combat`, and guards callbacks when comparison support is inactive.
- Queues no comparison cleanup while comparison support or the current mode suppression is inactive.

## Root cause

EllesmereUI suppresses the primary `GameTooltip` by parking it in a hidden host, but Blizzard renders item comparisons in separate `ShoppingTooltip` frames managed by `TooltipComparisonManager`.

The important ordering is:

1. Blizzard starts and populates `GameTooltip`.
2. Blizzard's item pipeline calls the comparison manager.
3. The comparison manager creates and shows `ShoppingTooltip1` and `ShoppingTooltip2`.

Therefore, the repo-compliant `OnShow` and `TooltipComparisonManager:AnchorShoppingTooltips` post-hooks can only clear the comparison after Blizzard has shown it. A brief first-frame flash can occur, but the comparison is then removed instead of remaining visible throughout combat.

This PR intentionally accepts that possible one-frame flash. Preventing the comparison from being created would require writing Blizzard's existing `suppressAutomaticCompareItem` control field before item finalization. That would be a write onto a Blizzard-owned frame and would conflict with the repository's zero-taint acceptance criterion. Post-show cleanup is the narrow, repo-compliant tradeoff for this bug fix.

## How was it tested?

- Reproduced with an active-use trinket placed on an action bar in the live WoW Midnight 12.1 client.
- Verified that a comparison tooltip already visible before combat disappears when combat starts.
- Verified that a comparison tooltip already visible when a boss encounter starts disappears immediately.
- Verified that re-hovering an item after combat or the boss encounter ends restores comparison behavior.
- Verified that `Show Tooltips = Always` continues to work.
- Verified that switching modes takes effect on subsequent tooltip or combat-state evaluation and reconciles the lazy event registration.
- No regression was observed in the tested tooltip flows.
- A brief flash while the comparison is first created in combat is expected with this repo-compliant implementation.
- The final inactive-mode guard was additionally static-checked and deployed after the live test.

## Screenshots

Before - the comparison tooltip remains visible during combat:

Before: comparison tooltip remains visible during combat

After - the comparison tooltip is cleared during combat:

After: comparison tooltip is cleared during combat

## Checklist

- [x] New settings default **OFF** - no setting was added; this is a genuine bug fix.
- [x] Zero cost while disabled - comparison state frames and transition events are created or registered lazily when a non-default mode is used. Comparison hooks are installed lazily on first use and return immediately when support is inactive; the default mode creates no comparison hooks, events, or comparison state frame.
- [x] Cheap while enabled - event-driven; no `OnUpdate` polling, wall-clock timing logic, or per-frame table allocations. `C_Timer.After(0)` is used only to defer cleanup outside a protected call stack.
- [x] Zero taint risk - no fields or addon state are written directly onto Blizzard-owned frames. Blizzard frames are observed with `HookScript`/`hooksecurefunc`, and `SetScript` is used only on addon-created frames; cleanup calls are protected with `pcall`.
- [x] Midnight only - tested in-game on live; the final inactive-mode guard was additionally static-checked and deployed, and no version gates or pre-Midnight APIs were added.

### DIFF
--- a/EllesmereUIBlizzardSkin/EllesmereUIBlizzardSkin.lua
+++ b/EllesmereUIBlizzardSkin/EllesmereUIBlizzardSkin.lua
@@ -2657,6 +2657,7 @@ do
     GameTooltip:HookScript("OnHide", function(tt)
         -- Only fires for unparked hides (a parked tooltip is never visible); clears the default-anchored flag promptly.
         _defaultAnchored = false
+        if not (_comparisonSupportActive or _comparisonStateWatcherRegistered) then return end
         if EllesmereUI._tooltipSuppressedByMode(tt) then
             QueueSuppressedComparisonClear()
         end
--- a/EllesmereUIBlizzardSkin/EllesmereUIBlizzardSkin.lua
+++ b/EllesmereUIBlizzardSkin/EllesmereUIBlizzardSkin.lua
@@ -2494,6 +2494,8 @@ do
     -- combat when alwaysCompareItems is enabled.  Keep the cleanup in this global
     -- tooltip controller rather than in ActionBars: bags, world items, and every
     -- other default-anchored item tooltip use the same comparison manager.
+    -- Use Blizzard's existing comparison-suppression flag; addon-owned state
+    -- stays in locals so no custom state is stored on Blizzard frames.
     local _comparisonTooltips = { ShoppingTooltip1, ShoppingTooltip2 }
     local _comparisonClearPending = false
     local _comparisonFlagOwned = false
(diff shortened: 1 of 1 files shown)


jixinliu666:
Maintainer note: if Blizzard's existing `GameTooltip.suppressAutomaticCompareItem` field is considered an acceptable and sanctioned control input for EllesmereUI, this fix could be tightened to arm it before Blizzard's item finalizer runs. That would prevent `ShoppingTooltip1`/`ShoppingTooltip2` from being created and eliminate the brief first-frame flash when hovering an action-bar item in combat.

The current PR intentionally avoids that field and uses repo-compliant after-show cleanup, so it does not require an exception. If the control field is allowed, I would be happy to update the implementation to use the no-flash path.

PR #1473, opened 2026-08-15
fix: borders covering text on skinned bars

Sanxy09:
## What does this PR do?
Fix borders on skinned bars (experience bar, reputation bar, etc.) covering text by lowering AddBorder drawLayer from OVERLAY to BORDER and removing the +1 frame level offset in PP.CreateBorder.

## How was it tested?
In-game: verified reputation bar labels and experience bar text are no
longer occluded by borders.

## Screenshots
image
image

## Checklist


EllesmereGaming:
adjusting this to be more surgical, same concept fix though. will be live with next patch

PR #1474, opened 2026-08-15
feat(pab): allow custom shapes and zoom for player aura icons like action bars

Barbiero:
## What does this PR do?

Adds custom **Icon Shape** support to Player Aura Bars (Buffs and Debuffs, default and custom bars), matching Action Bars' shape system:

- New **Icon Shape** dropdown (Square, Circle, Curved Square, Diamond, Hexagon, Portrait, Shield) with masked icons, a matching shaped border, and a shaped dispel-type ring for debuffs.
- **Border Size** changed from a 0-4 slider to a None/Thin/Normal/Heavy/Strong dropdown, mirroring Action Bars' `ns.BORDER_THICKNESS` — Thin/Normal/Heavy are disabled while a shape is active (shape borders only render two visually distinct states at aura-icon sizes), auto-snapping to Strong when entering shape mode from a disabled level.
- New **Icon Zoom** slider (was previously wired end-to-end but had no UI). For custom shapes, zoom is coupled to the mask-fill magnification the same way Action Bars does it, so the zoom slider actually reaches "unzoomed" instead of always forcing extra crop. Untouched for the "None" shape.
- **Apply to: All | Multiple** sync icons on Border Size and Icon Shape, covering every buff/debuff bar (default + custom), matching the Action Bars UX pattern.

## How was it tested?

Tested in-game on live across all Icon Shape options, Border Size levels, and Icon Zoom range, on both the default and a custom buff/debuff bar, in the Manager Page preview and in real combat (buffs and dispellable debuffs). Verified:
- Shape/border/zoom changes apply live without `/reload`.
- Cycling between shapes (including back to a previously-used one) and back to None repeatedly.
- Preview box matches the live bar's rendering at every Border Size level.
- No Lua errors (BugSack) during any of the above.

## Screenshots
Circular buff, Hex debuff (stockades is dark so buff are hard to see, sorry):
image

Diamond buff without border:
image

Shield, colored border, zoomed:
image

Options, I couldn't figure out a good way to place the "Multiple" link without overlap:
image

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) — Icon Shape defaults to None, Icon Zoom defaults to the prior hardcoded value; existing bars render identically until touched.
- [x] Cheap while enabled: event-driven, change-guarded (shape/border texture updates skip redundant `Add/RemoveMaskTexture` work when nothing changed)

### DIFF
--- a/EllesmereUI.lua
+++ b/EllesmereUI.lua
@@ -2739,6 +2739,62 @@ do
         if bd then bd.container:Show() end
     end
 
+    -- Variable-thickness border ring drawn from a single pre-shaped texture (natively
+    -- 7px thick), scaled by (7 - rawBorderSize) and clipped by `mask`. Called with `:`
+    -- so self.Point resolves to whichever PP module owns host's frame hierarchy
+    -- (world PP vs PanelPP) -- calling with `.` shifts every argument left by one.
+    function PP:ApplyMaskedShapeBorder(host, mask, texPath, rawBorderSize, r, g, b, a)
+        if not host then return end
+        rawBorderSize = rawBorderSize or 0
+        if rawBorderSize <= 0 or not texPath then
+            if host._shapeBorderShown then
+                host._shapeBorderTex:Hide()
+                host._shapeBorderShown = nil
+            end
+            return
+        end
+        r, g, b, a = r or 0, g or 0, b or 0, a or 1
+        -- Every field guarded: callers invoke this on every restyle pass regardless
+        -- of whether the border changed, and Remove/AddMaskTexture aren't cheap.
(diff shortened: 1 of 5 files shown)


Barbiero:
To note: I did not include the "cropped" version on PAB since that would be a lot of additional work and I wanted the other shapes to work

Also, the masks don't really work with the 0-4 border size that PAB was using much like how it is with Action Bars. I decided to mirror Action Bar settings as much as possible, including the "Only None or Strong" behavior when using custom shapes.

I think it'd be cool if these settings could also be synchronized across modules but in the end I decided to keep them PAB only

feedback is welcome

PR #1477, opened 2026-08-15
feat(quickdraw): add preset for ping system

tyyrenn:
## What does this PR do?

This PR adds a preset entry in the Quickdraw options that sets up a menu for using pings from the ping system. In doing so, it also adds support for rendering texture atlas textures as quickdraw icons. It does not, however, add any option-side support for configuring texture atlas textures on player defined menus. It also does not prevent such support from being added later.

## How was it tested?

I tested on 12.1.0 (Build 69299; current retail build as of this PR). I confirmed that the preset menu item was present and its icon rendered correctly and at the correct scale. I confirmed that choosing the preset creates a new action menu entry that the user can further edit. I tested that the icons are resizable via the existing size controls. I tested that the quickdraw menu renders correctly in-game using the arc, fan, and grid layouts, and that the pings fire when chosen from the menu.

### Preset Menu Entry
image

### Pings Action Menu Options
image

### In-world Menu Activation
image <br/>

image<br/>

image

## Checklist

### DIFF
--- a/EllesmereUIOptions/EUI_Quickdraw_Options.lua
+++ b/EllesmereUIOptions/EUI_Quickdraw_Options.lua
@@ -1613,6 +1613,41 @@ initFrame:SetScript("OnEvent", function(self)
         return out
     end
 
+    local function PingSlots()
+        return {
+            {
+                kind = "macrotext",
+                name = "Look",
+                macrotext = "/ping look",
+                icon = { atlas = "Ping_Marker_Icon_NonThreat" },
+            },
+            {
+                kind = "macrotext",
+                name = "Assist",
+                macrotext = "/ping assist",
+                icon = { atlas = "Ping_Marker_Icon_Assist" },
+            },
+            {
+                kind = "macrotext",
+                name = "Attack",
+                macrotext = "/ping attack",
+                icon = { atlas = "Ping_Marker_Icon_Attack" },
+            },
+            {
+                kind = "macrotext",
+                name = "Warning",
+                macrotext = "/ping warning",
+                icon = { atlas = "Ping_Marker_Icon_Warning" },
+            },
+            {
+                kind = "macrotext",
+                name = "On My Way",
+                macrotext = "/ping onmyway",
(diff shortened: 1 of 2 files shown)


Laxxzz:
Would be very cool to have the option to fold in the environment/unit toggle cVar to a separate list of ping options, allowing you to have 1 keybind for normal pings as you've created here, and a possible modifier+bind/completely separate bind for non-environment pings - toggle the cVar again after the ping sends and the normal ping bind operates normally again. 

Curious what your thoughts are on that?

tyyrenn:
> Would be very cool to have the option to fold in the environment/unit toggle cVar to a separate list of ping options, allowing you to have 1 keybind for normal pings as you've created here, and a possible modifier+bind/completely separate bind for non-environment pings - toggle the cVar again after the ping sends and the normal ping bind operates normally again.
> 
> Curious what your thoughts are on that?

I'm not sure I understand what you're asking for, but ultimately it's just running a macro when you pick an icon from the menu. Turning this into something more complicated than that is outside the scope of what I'm willing to take on right now.

If you can express what you're looking for as a macro, I'd consider testing it and updating the PR.

EllesmereGaming:
great PR thank you!

Laxxzz:
> > Would be very cool to have the option to fold in the environment/unit toggle cVar to a separate list of ping options, allowing you to have 1 keybind for normal pings as you've created here, and a possible modifier+bind/completely separate bind for non-environment pings - toggle the cVar again after the ping sends and the normal ping bind operates normally again.
> > Curious what your thoughts are on that?
> 
> I'm not sure I understand what you're asking for, but ultimately it's just running a macro when you pick an icon from the menu. Turning this into something more complicated than that is outside the scope of what I'm willing to take on right now.
> 
> If you can express what you're looking for as a macro, I'd consider testing it and updating the PR.

It's all good I can handle it as a separate PR. New functionality was added in 12.1 adding a cVar that allows you to toggle pings between their normal functionality, and only pinging the environment.

PR #1479, opened 2026-08-15
Tracking bars show wrong text

svart2521:
Bug: https://discord.com/channels/585577383847788554/1538120843597455481

The bug

Tracking Bars showed the wrong spell name (icon was fine, only the label text was wrong), and only in combat.

Along the way I fixed a few real but ultimately secondary issues: a spell-identity resolver that could guess the wrong spell when multiple related spells shared data, a frame-matching fallback that could bind ambiguously, and a stale-binding check for reused frame objects. 

The actual cause

The name text had a fallback that read Blizzard's own label directly off their frame (blizzNameFS:GetText()), trusting it completely. In combat, Blizzard's internal frame pool gets reused fast, and their name label can lag a tick behind their icon — so we were faithfully copying a leftover name from whatever that frame displayed before, even though the icon had already updated correctly.

The fix

Stopped reading Blizzard's label text for naming at all. The name is now derived from the bar's own configured spellID (deterministic, can't lag), with live aura data only as a fallback for when spell data isn't loaded yet — and that fallback is now validated against the config before being trusted.

Also left in a /euitbbdebug toggle that traces which code path set a name and what data it used, in case anything like this shows up again.

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICdmBuffBars.lua
+++ b/EllesmereUICooldownManager/EllesmereUICdmBuffBars.lua
@@ -2668,11 +2668,14 @@ local _findChildCache = {}
 -- Sticky cfg->frame bindings for the one-to-one assignment pass below (cfg table -> last
 -- paired Blizzard frame). Separate table for the same serialization reason; dropped on cache invalidation.
 local _tbbStickyFrame = {}
+-- cooldownID a sticky frame had when bound, to catch object reuse for a different slot.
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
manually merging as you had an error in the code that would've failed to parse and  broken the cdm:
luac: line 317: <eof> expected near 'end'

otherwise great PR, will be live with next patch ty!

Issue #1461, opened 2026-08-15
[Bug]The tick markers on the cast bar intermittently disappear and reappear during channeling.


ngblank:
I am experiencing an issue where the tick indicators on the cast bar frequently vanish while channeling Arcane Missiles. This problem persists even with all other addons disabled, and it appeared following the most recent EUI update(v8.8.7,v8.8,8).

ngblank:
https://github.com/user-attachments/assets/7d266d23-0d23-417a-b067-d5fb1fe2da9e

PR #1480, opened 2026-08-15
fix(castbar): accurate arcane missiles cast bar

uNBEx:
## What does this PR do?

Fixes cast bar channel tick marks for Arcane Missiles:

- The first and last missiles land at the bar's start and end, so a base channel now draws 3 dividers instead of 4.
- Accounts for extra missiles from Amplification (+2) and the tier set 2pc (+1): 3/4/5/6 dividers depending on which are active.
- Re-casting AM while a previous channel is still running keeps the old channel's missile schedule; the bar now draws the extra divider at the carried-over offset instead of a static fresh-cast grid. Re-casting immediately after a missile (which loses the bonus missile in-game) falls back to the fresh layout.

Also fixes a related bug for all tick-marked channels: a cast bar rebuild landing mid-channel (spec-override/profile refresh, observed live when gaining Bloodlust during a channel) hid the tick marks for the rest of the cast; the rebuild now redraws them.

## How was it tested?

12.1, target dummy, with/without Amplification talent, chaining and not, bloodlust during cast applied by party member.

## Screenshots

chaining_with_fix-ezgif com-optimize

## Checklist

- [x] New settings default **OFF** - N/A, no new settings; bug fix to the existing tick marks, still gated behind the Channel Ticks setting
- [x] Zero cost while disabled: no new events, hooks, or frames; all new code runs inside the existing setting-gated ShowChannelTicks / BuildCastBar paths
- [x] Cheap while enabled: runs once per channel start/update/rebuild, no polling, no per-frame allocations
- [x] No writes onto Blizzard-owned frames - N/A, touches only the addon's own cast bar
- [x] Tested in-game on live

### DIFF
--- a/EllesmereUIResourceBars/EllesmereUIResourceBars.lua
+++ b/EllesmereUIResourceBars/EllesmereUIResourceBars.lua
@@ -167,6 +167,11 @@ local PT = {
 --  Channel tick data: spellID -> { ticks, [modSpell, modTicks] } or { tickInterval }
 --  ticks = fixed tick count (haste speeds ticks up, count unchanged).
 --  tickInterval = fixed seconds/tick (haste extends duration, adds ticks).
+--  missiles = fencepost count: first missile AT channel start, last AT the
+--  channel end, so M missiles span M-1 intervals (interior marks at
+--  i/(M-1)); addMissiles = { [spellID] = extra } raises M while the player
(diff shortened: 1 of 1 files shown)


uNBEx:
this is also likely to fix https://github.com/EllesmereGaming/EllesmereUI/issues/1461

PR #1482, opened 2026-08-15
Ebon Might hijacking the potion

svart2521:
Bug: https://discord.com/channels/585577383847788554/1537991047035097169

Issue: A rule meant only for Blizzard's native cooldown bars had no check keeping it off custom bars. If a custom bar's icon frame ever carried a leftover cached spell ID from a reused frame, that rule could match it anyway — pasting the wrong icon/cooldown onto your potion slot. This got triggered by adjustments (icon size, glow, etc.) that force frames to get reused/reindexed.

Fix: Added an explicit check so built-in rules can only match icons on the three native bars (cooldowns, utility, buffs). Custom bars are now excluded outright, so a stale cached ID on a reused frame can no longer cause this cross-bar bleed.

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICdmFakeActive.lua
+++ b/EllesmereUICooldownManager/EllesmereUICdmFakeActive.lua
@@ -376,8 +376,12 @@ ApplyToFrame = function(iconFrame, rule, win)
     end
 end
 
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
this was actually a staleness bug, not the rule scope. fixed locally will be live with next patch

PR #1510, opened 2026-08-16
Add opt-in Party Target frames to party frames

apainter2:
Adds a minimal, opt-in Party Target feature: a small clickable secure frame next to each party member showing that member's current target (party1target..party4target). Left-click targets the unit via secure attributes so it works in combat with no taint.

The feature is Party-only and defaults OFF. It builds lazily on first enable (no frames, events, or work while disabled), is event-driven (UNIT_TARGET/UNIT_NAME_UPDATE/GROUP_ROSTER_UPDATE) with no OnUpdate or polling, and uses RegisterUnitWatch for visibility. Raid frames are untouched.

## What does this PR do?

Adds a minimal, opt-in Party Target feature: a small clickable frame next to each party member showing that member's current target (party1target..party4target). Left-click targets/selects that unit. Similar to a feather offered by Shadow Unit Frames and ElvUI.

It is a native extension of the existing party frame architecture in EllesmereUIRaidFrames.lua (the module that owns both party and raid frames). The feature is Party-only and defaults OFF.

- Zero cost while disabled: nothing is created or registered until enabled. OnEnable calls PT_SetEnabled(db.profile.partyShowTargets) (default false), which returns immediately when off.
- Lazy build on first enable: creates 5 secure child frames once, then registers events and unit watches.
- Taint-free secure click: each frame is a SecureUnitButton with useparent-unit=true + unitsuffix="target" + *type1="target", so the unit is resolved by Blizzard's secure handler at click time and works in combat with no protected mutation. Enable/disable is deferred out of combat.
- Cheap when enabled: event-driven (UNIT_TARGET filtered to owner units, UNIT_NAME_UPDATE, GROUP_ROSTER_UPDATE, PLAYER_ENTERING_WORLD); no OnUpdate, no timers, no per-frame allocations. Visibility owned by RegisterUnitWatch (no polling).
- Raid frames untouched: target frames attach only to party header buttons; no raidNtarget frames, no changes to raid creation/sorting/layout/events.

Configuration: Options -> Raid Frames -> Party tab -> FRAMES -> "Enable Party Targets".

Files changed:
- EllesmereUIRaidFrames/EllesmereUIRaidFrames.lua
- EllesmereUIOptions/EUI_RaidFrames_Options.lua

## How was it tested?

Live retail client. Verified: feature off = no frames/events; enabling shows target frames next to party members; name appears/updates on target change; left-click selects the unit (in and out of combat); frame hides when the member clears their target; disabling hides frames and stops work. Raid frames verified unchanged. No Lua errors.

## Screenshots

Before:
WoWScrnShot_081626_171706

After:
WoWScrnShot_081626_171716

## Checklist

- [x] New settings default OFF (no behavior change without opt-in)
- [x] No writes onto Blizzard-owned frames; HookScript/hooksecurefunc only, never SetScript on Blizzard frames

### DIFF
--- a/EllesmereUIOptions/EUI_RaidFrames_Options.lua
+++ b/EllesmereUIOptions/EUI_RaidFrames_Options.lua
@@ -5462,7 +5462,13 @@ initFrame:SetScript("OnEvent", function(self)
                   else db.profile.partyFlipGrowth = false end
                   PartyReloadAndUpdate()
               end },
-            { type="label", text="" });  y = y - h
+            { type="toggle", text="Enable Party Targets",
+              tooltip="Show a small clickable frame next to each party member with that member's current target. Left-click a target to select it.",
+              getValue=function() return db.profile.partyShowTargets or false end,
+              setValue=function(v)
+                  db.profile.partyShowTargets = v
+                  if ns.PT_SetEnabled then ns.PT_SetEnabled(v) end
+              end });  y = y - h
 
         -------------------------------------------------------------------
         --  ALL VISUAL SECTIONS
--- a/EllesmereUIRaidFrames/EllesmereUIRaidFrames.lua
+++ b/EllesmereUIRaidFrames/EllesmereUIRaidFrames.lua
@@ -702,6 +702,11 @@ local defaults = {
         partyUnlockPos    = nil,
         -- Party mirror of autoResizeTrackedBuffs ("Auto Resize Icons", Party tab); nil = on.
         partyAutoResizeTrackedBuffs = true,
+        -- Party Targets: a minimal clickable frame next to each party member
+        -- showing that member's current target (party1target..party4target).
+        -- OFF by default; the feature builds lazily on first enable (see
+        -- ns.PT_SetEnabled / the Party Target section below).
+        partyShowTargets = false,
     }
 }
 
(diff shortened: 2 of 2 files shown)


EllesmereGaming:
hey! thanks this looks promising. can i get a picture of the options? i think the targets should be much smaller relative to the actual party frames, similar to target unit frame vs target of target unit frame.

apainter2:
> hey! thanks this looks promising. can i get a picture of the options? i think the targets should be much smaller relative to the actual party frames, similar to target unit frame vs target of target unit frame.

image

apainter2:
Change to party target frame size as suggested by @EllesmereGaming.

image

apainter2:
Updated PR and my changes to match the 8.9.3 codebase.

EllesmereGaming:
we do have a feature halt currently but I am looking at potentially bringing this in next patch!

apainter2:
> we do have a feature halt currently but I am looking at potentially bringing this in next patch!

Hi there, I was wondering if this is still in the plans to be brought into the main release or to remain in a holding pattern?

apainter2 (commented):
Reviewed changes with several versions of EllemereUI code and no issues.

apainter2:
@EllesmereGaming PR updated to 9.2.1

PR #1512, opened 2026-08-16
Add SameBaseSpell function to compare ability IDs

svart2521:
Bug: https://discord.com/channels/585577383847788554/1538355647706431601

The issue

Tracking Bars could show data from the wrong spell — wrong name, wrong stack count, and the bar could even appear "active" for a spell that hadn't actually triggered. Root cause across all three: the code trusted whatever Blizzard's own frame was currently bound to (name, applications, active state) without checking that it was actually still bound to this bar's spell. Blizzard's frame pool reuses objects fast, especially in combat, so a bar could end up silently reading — and faithfully displaying — a completely different spell's data.

A related bug lived in the decimal-precision timer: it widened its "which aura is mine" filter using Blizzard's linkedSpellIDs, which groups mechanically-related-but-different spells together (e.g. an ability that requires a separate buff), not just true form-variants of the same spell.

The fix

Every place that reads identity-bearing data off a Blizzard frame now validates it against the bar's own configured spell before trusting it:

Name: derived from the bar's own spellID directly, not copied from Blizzard's label text.
Stacks: validated against the bar's spellID; if a read is proven to belong to a different spell, the raw Blizzard text fallback is skipped too (same wrong data, just unstructured) and the bar shows blank instead of a wrong number.
Active/shown state: same validation — a frame proven to belong to a different spell can no longer make the bar light up.
Decimal timer filter: only widens to a linked spell if it's genuinely a form-variant of the bar's own spell (same base spell), not just grouped in the same cooldown family.

Consistent principle throughout: an unvalidated guess is worse than a blank bar, so every fallback now fails safe rather than showing possibly-wrong data.

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICdmBuffBars.lua
+++ b/EllesmereUICooldownManager/EllesmereUICdmBuffBars.lua
@@ -2809,6 +2809,7 @@ local _tbbStickyFrame = {}
 -- only trust a binding while the pooled frame still serves the same cooldown
 -- slot (Blizzard reuses frame objects fast in combat). cooldownID stays a
 -- plain readable number in secret windows (the clean-sid cache keys on it).
+-- cooldownID a sticky frame had when bound, to catch object reuse for a different slot.
 local _tbbStickyCdID = {}
 
 function ns.InvalidateTBBFrameCache()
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
this has been adjusted, if not fixed with next update please lmk!

PR #1526, opened 2026-08-17
Nameplates: gate the dispel glow on what the player can dispel

nulltyto:
## What does this PR do?

Fixes a bug report (@nCarbon, v8.8.7): the nameplate Dispel Glow lights on enemy
buffs the player cannot dispel, and stays dark on ones they can.

The glow signalled ROW MEMBERSHIP, not capability. The buff row is narrowed with
`AuraUtil.AuraFilters.Dispellable`, which Blizzard documents as dispellable
"regardless of whether the player's raid can dispel them", and the glow then lit
every member of that row. A Mistweaver saw Captain's Bulwark (Magic) glow when
they only have Soothe; a Druid saw an enrage they CAN Soothe stay dark.

The addon already had capability-aware code with zero callers, and it could not
be revived as written: it classified each aura through `GetAuraDispelTypeColor`,
and in restricted content a tainted addon's aura reads are denied outright
rather than merely classified.

So the dispel type becomes a property of the aura GROUP, which the engine
resolves in its own Lua:

* The buff row splits into group `np` (`includeDispelTypes` Magic) and group
  `np2` (`excludeDispelTypes` Magic). The two are exact complements -- an enrage
  has no `dispelName`, so include rejects it and exclude keeps it.
* Only the group matching the player's capability glows. **Nothing is hidden**:
  every dispellable buff still renders, exactly as before.
* The split is taken only when the halves must differ -- one dispel type known
  and not the other, or both known with per-type colors on. A player who removes
  both kinds runs a single group, as today.
* Group 2 is declared lazily and parks at 0 frames when the split ends, so a
  character with no offensive dispel pays nothing.

Capability detection is rebuilt on the same pass. It called `IsPlayerSpell` /
`IsSpellKnown`, which live only in `Blizzard_DeprecatedSpellBook` behind the
`loadDeprecationFallbacks` CVar and are removed next expansion -- with that CVar
off, the talent branch never fired at all. It now uses `C_SpellBook.IsSpellKnown`
for talents (`IsSpellKnownOrInSpellBook` answers no for a passive) and registers
`TRAIT_CONFIG_UPDATED` and `PLAYER_ENTERING_WORLD`, so a talent swap no longer
needs a /reload. Mass Dispel was missing from the Priest entries.

Revives `dispelGlowUseTypeColor`, which had a default and no control: blue on
Magic, red on enrages. Off by default.

### Cost

No new per-frame work and no new allocation on a chatty event. One extra engine
group (4 buttons) per plate, only while split, declared lazily and parked at 0
afterwards. Two new registrations, both rare: `TRAIT_CONFIG_UPDATED` and
`PLAYER_ENTERING_WORLD`, each a 12-entry spellbook scan that no-ops unless
capability actually changed.

### Named trade-offs

1. While split, the row can reach 8 icons (4 per group) where it capped at 4.
   Dispellable enemy buffs are rare enough that this should not materialise in
   practice, but it is a real change.
2. Groups cluster rather than interleave: Magic icons sort ahead of enrages
   regardless of importance within the row.
3. Fail direction is "wrong, never late". A dispellable non-Magic enemy buff
   that is not an enrage would glow for a Soothe-capable player. I know of no
   such aura, but the deduction is ours, not the API's.

## How was it tested?

Live, 12.1.0 (build 69299). All four capability quadrants:

| Player | Magic buff | Enrage |
|---|---|---|
| Druid, Soothe talented | - | glows |
| Druid, untalented in place | - | dark (no /reload) |
| Shaman (Purge) | glows | dark |
| Hunter (Tranquilizing Shot) | glows | glows |
| Warlock, Felhunter out / dismissed | glows / stops | - |

Also verified: per-type colors on (blue/red) and off (single custom color), and
Show All Enemy Buffs suppressing the glow entirely.

**Known gap, stated plainly:** every test above was in unrestricted content. The
split reads `auraData.dispelName` through a candidate filter, and this repo
already documents a field incident where the `isStealable` candidate filter
matched nothing in instanced PvP because enemy aura data was redacted there (see
the comment above `BuffFilter`). If `dispelName` is redacted the same way, every
buff falls into the exclude-Magic group: a Magic-only dispeller would lose the
glow, and an enrage-only dispeller would see all of it glow. Untested. Happy to
gate the split on `AK.AurasRestricted()` as a fallback if reviewers prefer that
over waiting on a restricted-content capture.

Not yet tested: first plates after an in-combat /reload.

## Screenshots

screenshot-2026-08-16_18-59-00
screenshot-2026-08-16_18-58-47

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) -- the new
      `dispelGlowUseTypeColor` defaults off. The glow gate itself does change
      behavior without opt-in, because that is the bug being fixed.
- [x] Tested in-game on live; no version gates or pre-Midnight APIs added -- this
      REMOVES deprecated API calls rather than adding any

### DIFF
--- a/EllesmereUINameplates/EUI_Nameplates_AuraContainers.lua
+++ b/EllesmereUINameplates/EUI_Nameplates_AuraContainers.lua
@@ -224,11 +224,58 @@ end
 -- displayed buff is purgeable by construction and no per-aura signal is needed. Style and
 -- color come from the Dispel Glow options.
 
+-- Engine token for "has a dispel type". Blizzard documents it as dispellable
+-- REGARDLESS of whether the player's raid can dispel it, so it narrows the row
+-- but cannot answer "can I remove this" -- that is the capability gate's job.
+local DISPELLABLE = AuraUtil and AuraUtil.AuraFilters and AuraUtil.AuraFilters.Dispellable
+
 -- Effective glow state: with Show All Enemy Buffs on, the row is no
 -- longer dispellable-only, so the glow is suppressed entirely (the style
 -- dropdown and swatch are disabled in the options while that toggle is on).
+-- A character with no offensive dispel gets no glow either -- the option
+-- promises "you can dispel this", and the engine's DISPELLABLE token is
+-- documented as dispellable "regardless of whether the player's raid can
+-- dispel them", so it cannot answer that on its own.
 local function PurgeGlowActive()
-    return not not (ns.GetDispelGlow and ns.GetDispelGlow() and not PVal("showAllEnemyBuffs"))
+    if not (ns.GetDispelGlow and ns.GetDispelGlow()) then return false end
+    if PVal("showAllEnemyBuffs") then return false end
+    -- Without the DISPELLABLE token the row is not dispellable-only, so nothing
+    -- in it can be claimed to be dispellable by anyone.
+    if not DISPELLABLE then return false end
+    local magic, enrage = false, false
+    if ns.GetOffensiveDispelTypes then magic, enrage = ns.GetOffensiveDispelTypes() end
+    return magic or enrage
+end
+
+-- Buff row split. DISPELLABLE admits Magic buffs AND enrages, and the player
+-- rarely removes both, so the row is cut into two groups whose candidate
+-- filters are exact complements: group "np" carries Magic, group "np2" carries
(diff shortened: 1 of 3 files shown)


EllesmereGaming:
manually merged, derek mvp as per usual

PR #1537, opened 2026-08-17
 Bugfix/filtersmustfollowspec

cutekaylii:
## What does this PR do?

fix the bug where debuff filters are not saving per spec, current implementation was using the saving to same tile for all filters in debuff

## How was it tested?

in game testing, created different scenarios for spec healer /non healer

## Screenshots

image
image

## Checklist

- [y] New settings default **OFF** (no behavior change without opt-in)
- [y] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built
- [ y] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations)
- [y] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames
- [y] Tested in-game on live; no version gates or pre-Midnight APIs added


EllesmereGaming:
this one was actually a bit complex and half intended but not working the way it should've. I fixed it in a different way to reflect the way the system is used via buff manager so they match. excellent work though and highlighted the real problem area for me, much appreciated

PR #1549, opened 2026-08-17
Bags: add compact Armory slot groups

jacpub:
## What does this PR do?

Adds an opt-in **Compact Slot Groups** cog option beside **Group Armory by Slot**:

- Packs smaller Armory slot groups beside each other
- Keeps larger and multi-row slot groups on full rows
- Fills unused row space with empty-slot blocks
- Applies to The Armory and the Weapons / Trinkets, Armor, and Item Set Gear category views
- Automatically refreshes the open bag window size when either grouping option changes

The new setting defaults OFF. No runtime events, hooks, polling, or timers were added.

## How was it tested?

In-game test checklist completed:

- Default-off behavior remains unchanged.
- The compact option appears in the cog beside Group Armory by Slot.
- Smaller slot groups share rows correctly.
- Larger slot groups continue to use full rows.
- Empty-slot blocks fill unused row space correctly.
- Toggling either grouping option refreshes the open bag window size immediately.
- No load errors were observed during live retail testing.

Before/after comparisons:

1. Standard Armory slot groups
Screenshot 2026-08-17 133729

2. Compact Armory slot groups
Screenshot 2026-08-17 133625

3. Compact Slot Groups cog option
Before
Screenshot 2026-08-17 135407

After
Screenshot 2026-08-17 140922

## Checklist


EllesmereUIOptions/EUI_Bags_Options.lua
```diff
@@ -849,11 +859,68 @@ initFrame:SetScript("OnEvent", function(self)
...
+                          set=function(v)
+                              db.profile.bagCompactArmorySlotGroups = v and true or false
+                              ResetAndRefreshBagLayout()
+                          end },
+                    },
+                })
+                local leftRgn = armoryRow._leftRegion
+                local armoryCog = CreateFrame("Button", nil, leftRgn)
+                armoryCog:SetSize(26, 26)
+                armoryCog:SetPoint("RIGHT", leftRgn._control, "LEFT", -8, 0)
```
dfrisone, line 880: This cog never sets `leftRgn._lastInline` after anchoring, unlike every other inline cog in the codebase (e.g. `EUI_UnitFrames_Options.lua:4163-4164`). `QueueLabelClamp`/`_FlushLabelClamps` bound the row label against `region._lastInline or region._control`, so without this the label clamps against the toggle switch instead of the cog. Not visible with the current short label, but a longer translated string would run under the cog icon. Worth adding `leftRgn._lastInline = armoryCog` here.

EllesmereUIBags/EllesmereUIBags.lua
```diff
@@ -5840,6 +5974,13 @@ function EUI_Bags:RefreshInventory()
                 if #buckets > 0 then
                     local showAssign = assignCatIdx and EUI_CategoryManager
                         and EUI_CategoryManager:CanAssignToCategory(assignCatIdx)
+                    if RenderCompactSlotBuckets then
```
dfrisone: The `if RenderCompactSlotBuckets then ... else ... end` dispatch is written out three times across the call sites here, at ~6300, and at ~6535. The fallback loop bodies at the latter two are already textually identical apart from one variable name, and all three sites already reduce their category context to a single `assignCatKey` before calling `RenderCompactSlotBuckets`, so a shared `RenderSlotBuckets(buckets, subHeaderIdx, assignCatKey)` local would drop in cleanly at all three without extra parameters.

EllesmereUIOptions/EUI_Bags_Options.lua
```diff
@@ -32,6 +32,15 @@ initFrame:SetScript("OnEvent", function(self)
             local y = yOffset
             local h, _
 
+            local function ResetAndRefreshBagLayout()
```
dfrisone, line 35: `ResetAndRefreshBagLayout` duplicates the `_asCols`/`_asMaxGridW`/`_asMaxH` reset that's already inline in the pre-existing "Auto-Size to Fit" toggle about 70 lines below in this same function. Since this PR is already touching that area, folding the existing inline copy into this new helper would avoid a third place to update the next time an auto-size field is added.

EllesmereUIOptions/EUI_Bags_Options.lua
```diff
@@ -849,11 +859,68 @@ initFrame:SetScript("OnEvent", function(self)
...
+                })
+                local leftRgn = armoryRow._leftRegion
+                local armoryCog = CreateFrame("Button", nil, leftRgn)
+                armoryCog:SetSize(26, 26)
+                armoryCog:SetPoint("RIGHT", leftRgn._control, "LEFT", -8, 0)
+                armoryCog:SetFrameLevel(leftRgn:GetFrameLevel() + 5)
+                local armoryCogTex = armoryCog:CreateTexture(nil, "OVERLAY")
+                armoryCogTex:SetAllPoints()
+                armoryCogTex:SetTexture(EllesmereUI.COGS_ICON)
+                local function ArmoryCogOff()
```
dfrisone: `ArmoryCogOff()` and `ArmoryCogRequirement()` each independently fetch `db.profile.bagDisabledCategories` and re-test `dc["Armor"] == true`. Minor, but could fold into one `off, reason = ArmoryCogState()` function to match the file's other cogs, which only need a single predicate.

EllesmereUIBags/EllesmereUIBags.lua
```diff
@@ -5437,6 +5437,140 @@ function EUI_Bags:RefreshInventory()
         curY = curY - (blockRows * (SLOT_SIZE + SPACING))
     end
 
+    -- Compact Armory layout. Reuses the existing slot, sub-header, empty-pad,
+    -- and assignment pools; only the coordinates differ from the full-row path.
+    -- Build the closure only while both required settings are active.
```
dfrisone: This comment just restates the condition on the next line (`bagCompactArmorySlotGroups and ArmorySlotGroupingEnabled()` already reads as "both required settings are active"). The two lines above it explaining pool reuse are useful context, this one line could go.

dfrisone (commented):
Reviewed the compact Armory slot grouping logic in detail (coordinate math, upvalue threading, cross-file call sites, removed-behavior checks). No correctness bugs or regressions found, the new geometry and pool reuse all check out. A few minor cleanup items below, none of them blocking.

EllesmereUIOptions/EUI_Bags_Options.lua
jacpub: Good catch, added `_lastInline` so longer translated labels clamp correctly around the cog

EllesmereUIBags/EllesmereUIBags.lua
jacpub: Done. consolidated the three rendering paths into the shared RenderSlotBuckets helper

EllesmereUIOptions/EUI_Bags_Options.lua
jacpub: Done, the existing Auto-Size toggle now uses `ResetAndRefreshBagLayout` too

EllesmereUIOptions/EUI_Bags_Options.lua
jacpub: Done, combined the disabled state and reason into one `ArmoryCogState` function

EllesmereUIBags/EllesmereUIBags.lua
jacpub: Agreed, removed the redundant comment

PR #1551, opened 2026-08-17
Scope conditional-override spec precedence to the applying spec

dfrisone:
## Cause
A spec override outranked a conditional for any fkey ANY group had captured,
because the check reads a store-wide fkey index with no spec dimension. One
healer-only group therefore killed a Raid conditional on a Guardian belonging to
no group at all, and the value applied inside an editing session then vanished on
exit, permanently, with nothing in the UI to say why.

## Fix
SpecClaimsFKey scopes the check to fkeys the applying spec actually holds a value
for, applied symmetrically at all five sites: Cond.WriteValues, both branches of
Cond.Harvest, Cond.RestoreEntryDefaults, and the effective-value resolver. The
symmetry is load-bearing, not tidiness: scoping only the apply makes the
conditional write a value that nothing banks, silently discarding the user's edit.
Harvest also gains a conditional-held retain branch on the spec side, mirroring
the existing MatchOwnedFKey one, so the conditional's live value never banks as a
spec override. EntryOwning stays store-wide at its capture and UI ownership sites.

## Test
Needs the round trip that exercises banking: with a spec group owning a setting,
confirm the raid conditional applies with the panel CLOSED, edit it in raid, full
logout, and confirm the edit persisted and the shared default did not move.


dfrisone:
Added a third commit: the apply-side counterpart to the two already here.

Scoping the conditional's write to the applying spec was only half the
precedence model. A spec override entry tracks its fkey for every spec, but
only the specs in its group supply a value; the rest fall back to the baseline
recorded when the override was captured, and `WriteSpecValues` wrote that
baseline unconditionally. So an entry belonging to one group re-pinned its
pre-capture value on every apply for unrelated specs, and a conditional could
never own that key: its value was written and immediately overwritten.

Field case that found it: a Healing UI group holding Debuff Size pinned 34 on a
Blood DK, so the Raid conditional's 20 never applied. What made it hard to
diagnose is that everything looks correct except the frames. The conditional
stores 20, and opening it shows 20, because edit sessions bypass spec
precedence by design so you can see what you are editing. Only live rendering
shows the spec baseline.

The commit skips the baseline write when the spec does not claim the key itself
and the applied conditional supplies it. A spec that claims the key still wins.
Harvest already retains conditional-held fkeys, so apply and bank stay
symmetric. With no conditional applied `CondHeldFKey` returns false immediately,
so behaviour is unchanged for anyone not using conditional overrides.

Tested in game: the Raid conditional now applies on a spec the holding group
does not cover, without deleting the spec override.

dfrisone:
In-game verification of the precedence rule, both directions, same profile:

- Blood DK (spec 250). The holding group does not claim Debuff Size for it, so
  the Raid conditional now applies: live reads 20 in raid, back to the baseline
  out of raid. Before the fix it was pinned at 34 permanently.
- Restoration Druid (spec 105). Claims Debuff Size explicitly at 24. Stored and
  live both read 24, so an explicit per-spec value still outranks the
  conditional, unchanged.

That is the documented rule in both directions: a spec that claims the key wins,
a spec that does not lets the conditional own it.

dfrisone:
Banking round trip verified in game, which was the piece with no coverage:

- Edited Debuff Size 20 -> 26 with the Raid conditional applied.
- Left the raid: live returned to the baseline 34, and the applied pointer
  cleared. Nothing leaked into the default layer.
- Conditional store afterwards reads default 34, raid value 26. The edit banked
  into the conditional it was made under, and the baseline is untouched.

So all three sides hold: apply (conditional owns keys no spec claims), spec
precedence (an explicit per-spec value still wins), and bank (edits made under a
conditional persist to that conditional, not to the baseline).

PR #1552, opened 2026-08-17
Inherit baseline indicator sets for specs a Buff Manager fork never configured

dfrisone:
## Cause
_ERF_BM2ApplyLayer replaced the live per-spec indicator sets wholesale, so
activating a conditional Buff Manager silently emptied every spec/bucket the
fork hadn't captured. Confirmed against a reporter profile: a Death Knight's
Anti-Magic Shell had nowhere to render in raid because the fork's relevant
bucket was empty.

A first attempt merged the baseline UNDER the fork by presence (fork key exists
-> fork wins). That was defeated immediately: BM2_SpecInds seeds an empty
{inds={}} skeleton into a bucket on its FIRST READ (SeedSpec) and marks it
seeded, and BM2_SpecIndicators() reads every active bucket every frame -- so a
bucket ends up present-but-empty for reasons unrelated to what the user built,
and presence-based merge shadowed the baseline's real content the instant
anything looked at the bucket once. Worse for non-healer specs, whose active
bucket is the SHARED "nonhealer" key: an empty entry there ate the default
Defensives/Utility indicator for every non-healer spec at once.

## Fix
The merge is now by CONTENT: a fork bucket wins only when it actually holds
indicators (#v.inds > 0); an empty one falls through to the baseline instead of
shadowing it. Verified in isolation against four cases (empty-fork-inherits,
real-fork-wins, both-empty-stays-empty, fork-omits-key-inherits) with a
standalone Lua harness before touching the live file. The Debuff Manager needs
no equivalent: dmDebuff is a flat config, not a per-spec/bucket map.

## Test
Needs verification in game: activate a conditional Buff Manager, then check a
spec/bucket that was never deliberately customized in it while the condition
holds -- it should now show the baseline's indicators instead of nothing.
Separately confirm a bucket the fork WAS deliberately customized for still
shows the fork's own content, not the baseline's (no regression).


EllesmereGaming:
this behavior is actually intended but I have adjusted the way creating override buff managers works to be more intuitive/flexible. now on initial creation it will give you options to copy from existing ones or start blank or start with eui defaults

PR #1594, opened 2026-08-19
Fix Tracking Bar identity mismatches and secret-value crashes

svart2521:
Bug: https://discord.com/channels/585577383847788554/1539427695195066499
https://discord.com/channels/585577383847788554/1538933121049825374
https://discord.com/channels/585577383847788554/1539344874489389158
https://discord.com/channels/585577383847788554/1538355647706431601

The issue

Tracking Bars could display data belonging to the wrong spell — wrong names, wrong stack counts, and bars could show as "active" for a spell that wasn't actually up. Root cause: the code trusted whatever Blizzard's own frame happened to be bound to (name text, aura applications, active state) without verifying that frame still belonged to that specific bar's spell. Blizzard's frame pool reuses objects, especially during combat or when spells are cast together (e.g. a macro triggering Ghost Wolf and Spirit Walk simultaneously), creating brief windows where a bar could inherit a completely different spell's data — most visibly, Spirit Walk showing Ghost Wolf's name, stacks, or active state, since the two are mechanically linked in Blizzard's own data.

A related but separate issue: Death Advance's charge count showed a "0" that was either wrong (should've shown "1") or shown when it shouldn't have been displayed at all.

The fix
Identity validation everywhere data crosses a frame boundary: name, icon, stack count, and active-state all now verify the frame's aura actually belongs to the bar's configured spell before trusting it, instead of assuming any bound frame is correct.
MatchesSID no longer treats "linked" as "same spell": Ghost Wolf and Spirit Walk are grouped in Blizzard's data (Spirit Walk requires Ghost Wolf form) but aren't the same ability — matching now requires genuine form-variance (SameBaseSpell), not just group membership. Applied consistently in both the frame-matching logic and the decimal-timer filter.
Hardened against transient misreads: a config's captured base-spell ID now requires two consistent readings before committing (a single stale tick during frame reuse could otherwise permanently corrupt it), and the stacks display waits one tick after a frame rebind before trusting what it reads.
Fixed several unguarded secret-value comparisons (a newer WoW API restriction that blocks comparing certain combat-only data) that could crash mid-combat.

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICdmBuffBars.lua
+++ b/EllesmereUICooldownManager/EllesmereUICdmBuffBars.lua
@@ -2780,12 +2780,31 @@ ns.ApplyTBBBarSettings = ApplyTrackedBuffBarSettings
 --  Matches by cooldownID first (cached on cfg), then by spell ID variants from
 --  cooldownInfo. No external caches, no stale data in combat.
 -------------------------------------------------------------------------------
+-- cfg -> {sid=...} pending base-spell capture, awaiting a second confirming read.
+-- Separate table for the same serialization reason as sticky bindings.
+local _tbbBaseCapturePending = {}
+
+-- True if two ids are forms of the same ability, not just linked in the same group.
+local function SameBaseSpell(a, b)
+    if type(a) ~= "number" or type(b) ~= "number" then return false end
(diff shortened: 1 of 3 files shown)


EllesmereGaming:
excellent work but was a bit more complex so had to go a bit different of a route, can check the fallback system in the new update! confirmed with testers fixed

PR #1602, opened 2026-08-20
[Feature] Colour Class Resource Bar by if spell is usable check

RoakStatic:
## What does this PR do?
Adds a "Spenders" option to the Class Resource Bar's Threshold & Hash Lines settings. This option will recolour the bar while the tracked spell is currently castable - uses C_Spell.IsSpellUsable - so players can see at a glance when their spender ability is available. Same list/priority/UI shape as the existing Buff Colours option, but checked at lower priority. This option is designed to never override an active Buff Colours match.

Buff Colours own popup was refactored to be shared by the two features, so I didn't duplicate ~300 lines of identical code - hopefully this lessens the amount of work I need to do for later features.

Buff Colours own behaviour is unchanged, this refactor is most of the difference in size, not the feature itself.

## How was it tested?

WoW Client Build: 12.1.0 (69382)

- Spammed the toggle, confirming the bar colour applies and restores as expected.
- Ran Multiple Heroics via Dungeon finder
- Ran Delves
- World Questing throughout different zones

- Discovered an edge case with Spell Override abilities (Void Meta -> Collapsing Star) where IsSpellUsable returns a static false positive if not using the base/original spell ID - so using Collapsing Star as the SpellID instead of Void Meta. The correct method of inputting the SpellID in Spell Override cases has been outline in the Spenders tooltip so players are (hopefully) aware without needing to go to the Discord for help.
- Confirmed existing Buff Colours behaviour is unchanged after the popup refactor

## Screenshots

Feature Enabled:
Feature_Enabled

Spenders Options:
Spender_Colours

Buffs Options (Unchanged):
Buff_Colours

## Checklist


RoakStatic:
Cancelling this PR as it will be made redundant by my next one.

PR #1607, opened 2026-08-20
Fix(unitframes): doubled 1px border at castbar icon/bar seam

JuJuFX-dev:
Bugreport: https://discord.com/channels/585577383847788554/1526330850633977939

## What does this PR do?

Fixes a doubled 1px border where the cast bar icon meets the bar on Unit Frames castbars (player, target, focus, boss). The icon and the bar are separate frames, each carrying its own full 4-edge PP border, and since they sit flush against each other the shared edge drew twice, rendering as 2px instead of 1px. This suppresses the facing edge on each side using the same `_hideLeft`/`_hideRight` border-fusion pattern already used for the attached health/power seam elsewhere in this file, so only one clean 1px line remains at the seam. When the icon has a configured offset (no longer flush against the bar) both edges are shown normally instead, so a real gap never ends up with no border on either side.

## How was it tested?

Tested in-game on live.

## Screenshots

Before:
grafik

After:
grafik

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) -- N/A, no new setting
- [x] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built -- N/A, not a toggle


EllesmereGaming:
manually merged due to accidental stale tree

PR #1630, opened 2026-08-21
Fix: right clicking a target which is in the group but not in the zone would result in the pet frame menu

RedAces:
## What does this PR do?

Right-clicking the target frame on a group member could open the PET menu instead of the player one -- the same togglemenu misclassification v8.9.7 fixed for raid frames, which target/focus/targettarget/boss also sit on. Instead of passing the useCompactMenu flag at each site, the proxy factory now defaults to CompactUnitFrame_OpenMenu and the flag is dropped: the click-cast restore path and user-bound menu clicks fetch the proxy flagless and would otherwise have rebuilt a togglemenu proxy behind the fix. Safe for the non-group frames too -- the opener takes (frame, unit, button, isKeyPress), exactly what the secure "menu" action passes, and falls back to TARGET for NPCs.

## How was it tested?

Tested on 12.1 retail

## Checklist


RedAces:
Ive got this error if I right click a party member out of zone infight:
```
1x [ADDON_ACTION_BLOCKED] AddOn 'EllesmereUI' tried to call the protected function 'CheckInteractDistance()'.
[!BugGrabber/BugGrabber.lua]:554: in function '?'
[!BugGrabber/BugGrabber.lua]:538: in function <!BugGrabber/BugGrabber.lua:538>
[C]: in function 'CheckInteractDistance'
[Blizzard_UnitPopupShared/UnitPopupSharedUtils.lua]:203: in function <...ns/Blizzard_UnitPopupShared/UnitPopupSharedUtils.lua:190>
[tail call]: ?
[tail call]: ?
[Blizzard_Menu/MenuTemplates.lua]:112: in function <Blizzard_Menu/MenuTemplates.lua:109>
[tail call]: ?
[C]: ?
[Blizzard_SharedXMLBase/SecureTypes.lua]:385: in function 'CallFunctionIfSet'
[Blizzard_Menu/Menu.lua]:1065: in function <Blizzard_Menu/Menu.lua:1016>
[Blizzard_Menu/Menu.lua]:1126: in function <Blizzard_Menu/Menu.lua:1117>
...[C]: in function 'ExecuteAttribute'
[Blizzard_FrameXML/SecureTemplates.lua]:264: in function 'handler'
[Blizzard_FrameXML/SecureTemplates.lua]:755: in function <Blizzard_FrameXML/SecureTemplates.lua:733>
[Blizzard_FrameXML/SecureTemplates.lua]:769: in function <Blizzard_FrameXML/SecureTemplates.lua:762>
[Blizzard_FrameXML/SecureTemplates.lua]:821: in function <Blizzard_FrameXML/SecureTemplates.lua:805>
[C]: in function 'Click'
[Blizzard_FrameXML/SecureTemplates.lua]:565: in function 'handler'
[Blizzard_FrameXML/SecureTemplates.lua]:755: in function <Blizzard_FrameXML/SecureTemplates.lua:733>
[Blizzard_FrameXML/SecureTemplates.lua]:769: in function <Blizzard_FrameXML/SecureTemplates.lua:762>
[Blizzard_FrameXML/SecureTemplates.lua]:888: in function <Blizzard_FrameXML/SecureTemplates.lua:844>


Locals:
self=<table>
event="ADDON_ACTION_BLOCKED"
addonName="EllesmereUI"
addonFunc="CheckInteractDistance()"
name="EllesmereUI"
badAddons=<table>{
 EllesmereUI=true
}
L=<table>{
 NO_DISPLAY_2="|cffffff00The standard display is called BugSack, and can probably be found on the same site where you found !BugGrabber.|r"
 ERROR_DETECTED="%s |cffffff00captured, click the link for more information.|r"
 BUGGRABBER_STOPPED="|cffffff00There are too many errors in your UI. As a result, your game experience may be degraded. Disable or update the failing addons if you don't want to see this message again.|r"
 USAGE="|cffffff00Usage: /buggrabber <1-%d>.|r"
 STOP_NAG="|cffffff00!BugGrabber will not nag about missing a display addon again until next patch.|r"
 NO_DISPLAY_STOP="|cffffff00If you don't want to be reminded about this again, run /stopnag.|r"
 NO_DISPLAY_1="|cffffff00You seem to be running !BugGrabber with no display addon to go along with it. Although a slash command is provided for accessing error reports, a display can help you manage these errors in a more convenient way.|r"
 ERROR_UNABLE="|cffffff00!BugGrabber is unable to retrieve errors from other players by itself. Please install BugSack or a similar display addon that might give you this functionality.|r"
 ADDON_CALL_PROTECTED="[%s] AddOn '%s' tried to call the protected function '%s'."
}
```
Im not sure if this is related to this bugfix though

RedAces:
Unit Frames: only taint the menus that need it

4d77bd86 gave every menu proxy Blizzard's CompactUnitFrame_OpenMenu. That
opener is reached through ExecuteAttribute("menu-function", ...) -- an
attribute we wrote -- so our taint rides into the menu build and entries
with an interact distance block on CheckInteractDistance while in combat.
Most frames never needed it: togglemenu resolves party/boss/focus/arena
tokens through an explicit token case, and player/pet/vehicle on its first
UnitIsUnit checks. Only target, targettarget, focustarget and raidN fall
through to UnitIsOtherPlayersPet/UnitIsOtherPlayersBattlePet, which answer
true for a unit whose data has not streamed. Those four keep the compact
opener and its taint; everything else goes back to togglemenu and produces
none.

Routing the click into Blizzard's own hidden TargetFrame instead was tried
and does not work: its menu-function is not CompactUnitFrame_OpenMenu and
misclassifies the same way, so the default UI shows the pet menu too.

In order to get correct menus the taint is the only option IMHO. The taint does not block anything so its just a lua error. Afaik blizzard has to fix this upstream.

RedAces:
As far as Im aware is the PR https://github.com/EllesmereGaming/EllesmereUI/pull/1634 taintless, but with an incorrect right-click menu. As soon as blizzard fixes this bug upstream #1634 is the correct one.

PR #1635, opened 2026-08-21
feat(qol): add standalone movable gear durability display

apainter2:
Add an optional Quality of Life feature that shows a compact, independently movable "[icon] 87%" style readout of the lowest durability percent across equipped repairable gear, making it obvious when to repair.

- Event driven only (UPDATE_INVENTORY_DURABILITY, PLAYER_ENTERING_WORLD); no OnUpdate, ticker or polling.
- Zero cost unless enabled: default off, builds no frame and registers no events until first enabled; getFrame never creates so login-time ApplySavedPositions stays free while disabled.
- Reuses house systems: Lite.NewDB (own gearDurability slice on EllesmereUIQoLDB, other QoL data untouched), GetFontPath/SlugFlag fonts, MakeUnlockElement/Unlock Mode for movement and persistence, ShowWidgetTooltip for a per-slot breakdown, DualRow/DependentSetValue options widgets.
- Durability maths and white->red warning gradient mirror the DataBars durability block; reuses the DataBars forge glyph.

## What does this PR do?

Adds an optional Quality of Life feature: a compact, independently movable gear durability display that shows the **lowest durability percent across equipped repairable gear**, so it's obvious at a glance when to repair.

- Renders as `[forge icon] 87%` (icon and percentage each toggleable).
- Lives under **Quality of Life -> Keys, Logs & Brez**, in a new **GEAR DURABILITY**
  section alongside the BattleRes / Bloodlust trackers.
- Movable via the existing **Unlock Mode**; position persists per profile.
- Hover shows a per-slot breakdown through the house widget tooltip.
- Optional dynamic white -> red warning colour (mirrors the DataBars durability
  block), a static colour when dynamic is off, and an optional "hide at full".

It reuses the existing durability maths and white->red gradient from the DataBars durability block (so the two readouts always agree) and reuses the DataBars forge glyph. It does **not** modify or replace the DataBars block; the differentiator is that this is standalone and independently movable without requiring DataBars.

## How was it tested?

Tested in-game on live (Retail/Midnight 12.1), no Lua errors:
- Enable/disable via the section dropdown; sub-rows appear only when enabled.
- Toggled Show Icon / Show Percentage / Dynamic Colour / Hide at Full; adjusted Font Size.
- Dragged in Unlock Mode; `/reload` restores position; survives profile switch.
- Verified updates are event-driven (damage at a dummy / vendor repair) with no OnUpdate or polling.
- Confirmed nothing is built and no events are registered while the feature is off.

Also validated all touched Lua files with lua-language-server `--check` (zero diagnostics); ASCII-only, Lua 5.1, no `goto`.

## Screenshots

Before addon activation:
before-activation

After activation:
after-activation

Unlock Mode:
unlock mode

Durability Addon displayed:
durability

Detailed - When hovered over:
hover-detailed

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) -- `visibility` defaults to `NEVER`
- [x] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built -- `Apply()` builds nothing while off, and `getFrame` never creates (so login-time `ApplySavedPositions` stays free)
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations) -- only `UPDATE_INVENTORY_DURABILITY` + `PLAYER_ENTERING_WORLD`
- [x] No writes onto Blizzard-owned frames -- the feature only creates its own `UIParent` child; no `SetScript` on Blizzard frames

### Changelog

- **12e5cced** feat(qol): add standalone movable gear durability display
  - New Quality of Life feature: compact, independently movable `[icon] 87%`
    readout of the lowest durability across equipped repairable gear.
  - Options under QoL -> Keys, Logs & Brez (Enable, Show Icon, Show Percentage,
    Dynamic Colour, static Text Colour, Hide at Full, Font Size).
  - Movable via Unlock Mode with per-profile position; per-slot hover tooltip.
  - Reuses the DataBars durability maths + white->red gradient and forge glyph.
  - Default OFF; zero cost while disabled.

- **fdd93c19** fix(qol): refresh on all durability-change edges
  - On-screen number was stale after damage until /reload; now re-samples on
    UPDATE_INVENTORY_DURABILITY, PLAYER_REGEN_ENABLED, PLAYER_UNGHOST,
    PLAYER_EQUIPMENT_CHANGED, MERCHANT_CLOSED and PLAYER_ENTERING_WORLD (still
    event-driven, no polling).

Quick note on timing: I know CONTRIBUTING.md currently pauses feature requests until a few weeks after the 12.1 launch, with only bug fixes accepted in the meantime. I'm opening this now mainly for visibility and early feedback on the approach rather than to jump the queue.

I'm completely happy to:
- leave it sitting until the freeze lifts,
- close and re-open it later if you'd prefer, or
- adjust the design first if a standalone display isn't the direction you want.

For context, I deliberately kept it self-contained and reused the existing DataBars durability maths, the white->red colour model, the forge glyph, Unlock Mode, the house tooltip, and the DualRow options widgets, so it should sit naturally within the existing conventions and stay zero-cost while disabled. If you'd rather see this as a small extension of the DataBars durability block instead of a separate QoL element, I'm glad to rework it that way.

No rush at all on review, and thanks for the great addon.

### DIFF
--- a/EllesmereUIOptions/EUI_QoL_AutoLogging_Options.lua
+++ b/EllesmereUIOptions/EUI_QoL_AutoLogging_Options.lua
@@ -265,6 +265,14 @@ local function BuildAutoLoggingPage(pageName, parent, yOffset)
         y = y - lustH
     end
 
+    ---------------------------------------------------------------------------
+    --  GEAR DURABILITY
+    ---------------------------------------------------------------------------
+    if _G._EUI_BuildGearDurabilitySection then
+        local durH = _G._EUI_BuildGearDurabilitySection(parent, y, W, EllesmereUI.PP)
+        y = y - durH
+    end
+
     return math.abs(y - yOffset)
 end
 
--- a/EllesmereUIOptions/EUI_QoL_GearDurability_Options.lua
+++ b/EllesmereUIOptions/EUI_QoL_GearDurability_Options.lua
@@ -1,0 +1,115 @@
+if EUI_CLIENT_BLOCKED then return end -- pre-12.1 client failsafe (EllesmereUI_ClientGate.lua)
+-------------------------------------------------------------------------------
+--  EUI_QoL_GearDurability_Options.lua
+--  Options section for the standalone Gear Durability display (registered under
+--  EllesmereUIQoL). Embedded on the same page as the BattleRes / Bloodlust
+--  trackers via _G._EUI_BuildGearDurabilitySection.
+-------------------------------------------------------------------------------
+
+if not EllesmereUI._ModuleNS["EllesmereUIQoL"] then return end  -- module disabled: no options page
+
+local function DB()
+    local fn = _G._EUI_GearDurability_DB
+    return fn and fn() or nil
+end
+
+local function P()
+    local d = DB()
+    return d and d.profile and d.profile.gearDurability
+end
+
+local function Cfg(key, fallback)
+    local p = P()
+    if not p then return fallback end
+    if p[key] == nil then return fallback end
+    return p[key]
+end
+
+local function Set(key, v)
+    local p = P()
+    if p then p[key] = v end
+end
+
+local function Refresh()
+    if _G._EUI_GearDurability_Apply then _G._EUI_GearDurability_Apply() end
+    -- Content-sized element: keep the live mover in sync after size changes.
+    if EllesmereUI._unlockActive and EllesmereUI.RepositionBarToMover then
+        EllesmereUI.RepositionBarToMover("EUI_GearDurability")
+    end
+end
+
+local VIS_VALUES = { ALWAYS = "Always", NEVER = "Never" }
+local VIS_ORDER  = { "ALWAYS", "NEVER" }
+
+_G._EUI_BuildGearDurabilitySection = function(parent, yOffset, W, PP)
+    local y = yOffset
+    local _, h, row
+
+    _, h = W:SectionHeader(parent, "GEAR DURABILITY", y); y = y - h
+
+    row, h = W:DualRow(parent, y,
+        { type="dropdown", text="Enable Gear Durability",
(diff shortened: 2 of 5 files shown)


apainter2:
Pushed a follow-up fix (fdd93c19): the on-screen readout wasn't refreshing after
gear took damage -- it stayed at the last value until a /reload.

Root cause: it only re-sampled on `UPDATE_INVENTORY_DURABILITY` and
`PLAYER_ENTERING_WORLD`. On its own that first event is unreliable for a
standalone readout -- durability is lost in combat / on death, where it can fire
before the value has settled -- and unlike the DataBars durability block there's
no co-located time-driven block driving a shared 1s ticker to paper over it. (The
hover tooltip always looked correct because it samples live on each hover, which
is what made the staleness obvious.)

Fix: re-sample on the concrete durability-change edges instead, still fully
event-driven with no polling or OnUpdate:

- `UPDATE_INVENTORY_DURABILITY` - repair / damage edges
- `PLAYER_REGEN_ENABLED` - leaving combat (catches durability lost in combat / on death once settled)
- `PLAYER_UNGHOST` - durability applied after a corpse resurrect
- `PLAYER_EQUIPMENT_CHANGED` - gear swaps changing the lowest item
- `MERCHANT_CLOSED` - after a vendor repair
- `PLAYER_ENTERING_WORLD` - login / reload / zone

Retested in-game: the number now updates without a manual reload (most visibly
right as you leave combat), and it stays zero-cost while the feature is disabled.

apainter2:
Hi there,

Is there any news or updates on when QoL feature requests such as the above may or may not be included in the main release? Just wondering if I should convert this into a QoL Addon depending on timescales.

apainter2:
@EllesmereGaming PR updated to 9.2.1

PR #1645, opened 2026-08-21
feat(qol): add Compact Band layout to Raid Tools

0x963D:
## What does this PR do?

Adds an opt-in **Compact Band** choice to Raid Tools' existing **Show as** setting. **One Window remains the default**, and the existing layouts keep their current behavior.

Compact Band is one resizable row containing:

- Blizzard's eight raid-marker icons plus the existing clear icon
- **RC** for Ready Check, with optional Role Check on right-click
- **PT** for the three configured pull timers and stop action

Marker buttons share target and world actions through mouse buttons and Shift modifiers. The band registers as its own Unlock Mode element, so the standard cog provides width, height, anchoring, and size matching without adding a special mover API.

This rewrite deliberately removes the previous Chat integration, bundled Lucide assets, and Quick Fire bindings. RC and PT are plain text so the PR does not introduce opinionated third-party iconography. Project-owned icons matching the existing utility strip can replace them later if desired.

## How was it tested?

- Parsed both changed Lua files with `luaparse`.
- Ran `git diff --check`.
- Scanned additions for non-ASCII text.
- Verified that the locale key list is unchanged.
- Ran the UI consistency detector with no findings.
- Verified the final PR is one commit changing only the Raid Tools runtime and options files.

A fresh in-game pass and screenshots are still pending after this rewrite.

## Screenshots

Pending a fresh local in-game capture. The previous screenshots were removed because they showed the discarded Chat-matched layout and Lucide controls.

## Checklist

- [x] New layout defaults **OFF**: `showAs` remains `one`.
- [x] Zero Compact Band cost unless selected: its frames and `RAID_TARGET_UPDATE` listener are created/registered only for Compact Band.
- [x] Enabled behavior is event-driven; no polling or timer loop was added.
- [x] No Chat or Blizzard-owned frame is modified.
- [x] Existing Raid Tools layouts remain available.
- [ ] Fresh in-game verification after the rewrite.

### DIFF
--- a/EllesmereUIOptions/EUI_QoL_RaidTools_Options.lua
+++ b/EllesmereUIOptions/EUI_QoL_RaidTools_Options.lua
@@ -54,10 +54,24 @@ initFrame:SetScript("OnEvent", function(self)
         return (Cfg("mode") or "never") == "never"
     end
 
-    -- The runtime owns the normalize (unknown values read as "one").
+    -- The runtime owns the normalize (unknown values retain One Window).
     local function ShowAsVal()
         if ns.ShowAs then return ns.ShowAs() end
         return Cfg("showAs") or "one"
+    end
+
+    local function LegacyDisabled()
+        return Disabled() or ShowAsVal() == "compact"
+    end
+    local function RoleCheckDisabled()
+        return Disabled() or ShowAsVal() == "markers"
+    end
+    local function FullPanelButtonDisabled()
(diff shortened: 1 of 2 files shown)


0x963D:
Superseded by #1755. The replacement is rebased onto v9.0.6 and contains only the opt-in Compact Band layout. Quick Fire and all Chat integration were removed from this change.

PR #1651, opened 2026-08-21
Chat font selection ignored for the client's native CJK alphabet

svart2521:
Bug: https://discord.com/channels/585577383847788554/1540235406463860776

Issue: ECHAT.EngineFontFamily() built a per-alphabet font family for chat text, but hardcoded all three CJK alphabets (korean/simplifiedchinese/traditionalchinese) to fixed stock font files via CJK_FILES, ignoring the user's font selection. On a zhTW client, nearly all chat renders through the traditionalchinese member, so the font option had almost no visible effect — only the rarely-used roman/russian members respected it.

Fix: When an alphabet matches the client's own locale (CJK_CLIENT_ALPHABET), use the user-selected font instead of the stock file. The other two (non-native) CJK alphabets still fall back to the stock files, preserving the original tofu-prevention behavior for foreign-script text that occasionally shows up in the log.


EllesmereGaming:
you have a syntax error so can't merge but applied locally, in with next patch good work!

PR #1657, opened 2026-08-22
removed old unused strings, Split important UI strings into separate sections, improved translations for some strings, added EUI 9.1.8 localizations.

DlargeX:
## Checklist


EllesmereGaming:
ty dlarge!! could you please look at reworking this as an additive PR, the massive changes to the rest of the file cause regressions in several areas. just drop the deletions and resectioning for now we can revisit if truly necessary down the road but for now, we can't keep rewriting 90% of the file every update

DlargeX:
Oh, sorry, I didn't mean to give you extra work, but I'm finished now. :)

EllesmereGaming:
this still shows 229 still in use keys being removed so I can't merge as is yet

PR #1659, opened 2026-08-22
fix(abr): clear combat state after resets

Nnoggie:
## Summary
- clear stale AuraBuffReminders combat state when an encounter reset does not produce a player combat transition
- run the existing out-of-combat reminder cleanup once lockdown is gone


Nnoggie:
this one is tested and should be good to merge, there was one residual case missed when encounter started but player never got into combat themself (hunter reset etc)

PR #1661, opened 2026-08-22
Fix xp/rep databar showing 0% for friendship reputations

solocommand:
For friendship factions (Captain Tokka, etc.) the watched reputation does not display correctly, 0% with 8400/8401.

image

## What does this PR do?
Before rendering, check if the tracked reputation is a friendship rep, and use the standing/friendship rank thresholds if true.

## How was it tested?
Screenshots below, tested by switching tracked factions

## Screenshots

image
image

## Checklist

Pure bugfix, one additional API call added to identify the reputation type, and one to retrieve the standing/thresholds if needed.

- N/A: New settings default **OFF** (no behavior change without opt-in)
- N/A: Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built
- N/A: Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations)
- N/A: No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames

### DIFF
--- a/EllesmereUIDataBars/EllesmereUIDataBars_Blocks.lua
+++ b/EllesmereUIDataBars/EllesmereUIDataBars_Blocks.lua
@@ -2216,8 +2216,29 @@ ns.BlockFactories.xprep = function(blockCfg, slot, content, barCtx)
                 r = 0.5, g = 0.5, b = 0.5, rested = 0,
             }
         end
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
excellent fix thank you!

PR #1667, opened 2026-08-22
Update MailBox Interaction Check

zanthor:
Fixed QOL Auto-Open to suspend when mailbox is being interacted with instead of checking for the frame visibility.  This widens support to any mail addon instead of just native.

## What does this PR do?

Suspend Auto-Open when users are interacting with mailbox but not using stock UI.

## How was it tested?

Tested off v8.9.8 codebase and tested in game with Stock UI and TSM.

## Checklist

- [X] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations)
- [X] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames

### DIFF
--- a/EllesmereUIQoL/EllesmereUIQoL.lua
+++ b/EllesmereUIQoL/EllesmereUIQoL.lua
@@ -229,7 +229,10 @@ qolFrame:SetScript("OnEvent", function(self)
             return (MailFrame and MailFrame:IsShown()) and true or false
         end
         local function BankOpen()
-            return (BankFrame and BankFrame:IsShown()) and true or false
+            -- return (BankFrame and BankFrame:IsShown()) and true or false
+            -- Changed to checking interaction state instead of framestate for wider
+            -- compatability with other addons.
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
diagnosis was good here ty! but the fix was a bit wrong and would've broken some other things so I've applied a separate patch locally, will be live with next patch and should fix the issue, thank you!

zanthor:
For my learning can I bother you to point me at the fix you implemented?

EllesmereGaming:
Your diagnosis was right, the fix just needed to land in MailOpen() rather than BankOpen(). This is the new fix:

MailOpen() → C_PlayerInteractionManager.IsInteractingWithNpcOfType(Enum.PlayerInteractionType.MailInfo)
BankOpen() → same idea with its own types: Banker or AccountBanker

Your original patch put the MailInfo check into BankOpen(). It still suspended at the mailbox (which is why your testing passed), but it silently removed the bank suspension.

zanthor:
Damn it... that was sloppy of me.  It was literally right there...

PR #1668, opened 2026-08-22
Fix: ADDON_ACTION_BLOCKED on LFG auto-signup (Quick Signup taint issue)

svart2521:
Bug: https://discord.com/channels/585577383847788554/1540140610260963428

Issue: The "Quick Signup" feature auto-clicks the group-finder sign-up button from an addon OnShow hook. Because that click originates from addon code rather than a real user click, WoW's taint system blocks the resulting call into the protected ApplyToGroup() API, throwing ADDON_ACTION_BLOCKED.

Fix: Route the click through a SecureHandlerBaseTemplate frame. Insecure code only sets attributes (button reference, setting state, shift-key state); the actual :Click() runs inside a SecureHandlerWrapScript snippet, which executes without inheriting the addon's taint. Applied the same fix to the classic Dungeon Finder role-check accept button, which had the identical exposure.

### DIFF
--- a/EllesmereUIQoL/EllesmereUIQoL.lua
+++ b/EllesmereUIQoL/EllesmereUIQoL.lua
@@ -1332,16 +1332,33 @@ qolFrame:SetScript("OnEvent", function(self)
             end)
 
             -- Auto-accept role check; Holding Shift skips it so the dialog stays open (e.g. to type a signup note).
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
problem is deeper rooted then this I will have to do more research but will keep you in the loop

PR #1684, opened 2026-08-23
Add opt-in Combat Only visibility toggle for cursor, GCD, and cast circles

torda385:
## What does this PR do?

Adds an opt-in **Combat Only** toggle to the cursor circle, the GCD circle, and
the cast bar circle. When enabled, that element is shown only while the player
is in combat. Defaults off for all three, so nothing changes for existing users
until they turn it on.

Implementation follows the existing `instanceOnly` pattern rather than
introducing a new one:

* New `combatOnly` boolean in each of the three defaults tables, all `false`.
* Cursor circle: gated in `UpdateVisibility` alongside the existing
  `instanceOnly` check.
* GCD and cast circles: checked in their `UNIT_SPELLCAST_*` handlers, in
  `ApplyGCDCircle` / `ApplyCastCircle`, and folded into the existing hide
  branches in `UpdateVisibility`.
* Combat state is read via `InCombatLockdown()`, a native API, so there is no
  dependency on another module's load or event order.

`PLAYER_REGEN_DISABLED` and `PLAYER_REGEN_ENABLED` are registered so the cursor
circle re-evaluates on combat transitions. The GCD and cast circles are already
event-driven off `UNIT_SPELLCAST_*` and needed no new registrations.

In the options page, the GCD and cast toggles fill the empty right slot that
already existed on their "Only Show in Instances" rows, so neither section gains
a mid-section gap. The cursor section gets a new final row with an empty right
slot.

**One thing to flag on criterion 1:** the two `PLAYER_REGEN_*` registrations are
currently unconditional, so a user who never enables Combat Only still has them
registered. `UpdateVisibility` is cheap (no allocations, no loops) and the module
already registers `PLAYER_ENTERING_WORLD` and `ZONE_CHANGED_NEW_AREA` the same
way, so I judged this consistent with the existing pattern rather than a new
cost. If you would rather have them registered lazily on first enable and
unregistered when all three toggles are off, say the word and I will add that.

## How was it tested?

Live client, Midnight 12.1. Tested each of the three toggles independently and in
combination: entering and leaving combat, in open world and in instanced content,
with the toggle flipped both while in and out of combat. Verified the existing
"Only Show in Instances" and "Only Show When Hidden" behavior is unchanged when
Combat Only is off, and that the two options interact correctly when both are on.
Also confirmed both files compile clean under `luac5.1 -p` and that all added
lines are ASCII only.

## Screenshots

image

## Checklist

  <br>Mostly yes: no frames, no hooks, no polling, no work in any handler while
  disabled. Not fully: two `PLAYER_REGEN_*` events are registered
  unconditionally. See the note above, happy to gate them lazily if preferred.
- [x] N/A. No Blizzard frames are touched. No `SetScript` on Blizzard frames, no hooks added.

### DIFF
--- a/EllesmereUIOptions/EUI_QoL_Cursor_Options.lua
+++ b/EllesmereUIOptions/EUI_QoL_Cursor_Options.lua
@@ -297,6 +297,19 @@ initFrame:SetScript("OnEvent", function(self)
                 if _G._ECL_ApplyOnlyWhenHidden then _G._ECL_ApplyOnlyWhenHidden() end
               end }
         );  y = y - h
+                
+        -- Combat Only
+        _, h = W:DualRow(parent, y,
+            { type="toggle", text="Combat Only",
+              tooltip="Only shows the cursor circle while you are in combat.",
+              getValue=function() local p = DB(); return p and p.combatOnly or false end,
+              setValue=function(v)
+                local p = DB(); if not p then return end
+                p.combatOnly = v
+                if _G._ECL_UpdateVisibility then _G._ECL_UpdateVisibility() end
+              end },
+            { type="label", text="" }
+        );  y = y - h
         end   -- close Cursor Circle hidden-while-disabled gate
 
         _, h = W:Spacer(parent, y, 20);  y = y - h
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
great work ty! I did end up gating the events, nice catch :D in with next patch

PR #1689, opened 2026-08-23
Right-click pet capture blocked by "Disable Right Click" enemy-targeting override

svart2521:
Bug: https://discord.com/channels/585577383847788554/1540227593155117107

Summary: With "Disable Right Click" enabled, right-click was rebound to mouselook whenever the mouseover unit matched WoW's harm macro condition. That condition is true for any attackable unit, including neutral wild pets — not just hostile mobs — so it was intercepting right-click on capturable pets and blocking the capture interaction.

Fix: Split the logic by combat state. Out of combat (where pet capture happens), the binding is now computed directly, explicitly excluding wild pets and pet-battle companions, using a plain override-binding call rather than the macro-based driver. In combat, the original macro-driven secure driver is still used (only it), since that's required to change bindings safely mid-fight. The two paths no longer run at the same time, so they can't conflict.

### DIFF
--- a/EllesmereUIQoL/EllesmereUIQoL.lua
+++ b/EllesmereUIQoL/EllesmereUIQoL.lua
@@ -2805,6 +2805,53 @@ do
     end)
 
     local stateFrame = CreateFrame("Frame", "EUI_NoRightClickState", UIParent, "SecureHandlerStateTemplate")
+    local combatDriverActive = false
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
went with a different more performance efficient approach, but excellent diagnosis!

PR #1690, opened 2026-08-23
Fix: Class power ignore group members' spec events & optimization

RedAces:
## What does this PR do?

PLAYER_SPECIALIZATION_CHANGED carries a unit and is delivered for group members too, which two class power handlers did not account for.

Unit Frames: the container's own event driver registered the event without a unit filter and only tore the bar down, leaving the rebuild to RefreshAllAddons -- which never runs on a shared profile. Any teammate's spec change killed the bar until the next /reload, and the strip reserved by ResizeFrameForClassPower stayed behind as an empty, tickless bar. cpSpecWatcher already owns this event correctly (outside the container, unit-filtered, rebuilding after teardown), so drop the duplicate handler. Affects custom-resource specs only (Arms, Fury, Enhancement, Survival, Frost Mage, Vengeance/Devourer DH, Brewmaster).

Nameplates: same missing filter, but harmless -- ApplyClassPowerSetting rebuilds synchronously. Filter anyway so a raid full of spec swaps stops rebuilding the watcher and every pip on the personal plate for nothing.

## How was it tested?

Tested on an arms warrior in a +8 Delve (12.1 retail)

## Checklist

### DIFF
--- a/EllesmereUINameplates/EllesmereUINameplates.lua
+++ b/EllesmereUINameplates/EllesmereUINameplates.lua
@@ -4411,6 +4411,11 @@ local function EnableClassPowerWatcher()
         end
         classPowerWatcher:SetScript("OnEvent", function(_, event, ...)
             if event == "PLAYER_SPECIALIZATION_CHANGED" then
+                -- The event carries a unit and is delivered for GROUP MEMBERS too, so
+                -- without this filter a raid full of spec swaps rebuilds the watcher
+                -- (and every pip on the personal plate) over and over for nothing.
+                local specUnit = ...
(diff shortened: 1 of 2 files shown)


RedAces:
The bug happens for these specs:
- Arms Warrior
- Fury Warrior
- Enhancement Shaman
- Survival Hunter 
- Frost-Mage
- Vengeance-DH
- Devourer-DH 
- Brewmasster Monk

RedAces:
Test it:
- have any of the mentioned specs
- go into a group with another player
- have this other player respec -> your class bar (topmost bar in your player unit frame - if enabled) is empty

PR #1713, opened 2026-08-24
Bags: Bag and Bank Columns sliders, and fix Window Scale not rescaling an open bank

### LINKED ISSUE #1175: [Feature Request] Make bank column count configurable (currently a hardcoded local)
### **What**

A `bankColumns` profile key with a slider in the Bags options panel, so the bank grid width can be changed like the other settings that already apply to both windows.

Bags are covered separately by #321 (part 2), which asks for a column option for the player bag. This is only about the bank, which #321 does not mention. The two are different problems: the bag value at least exists in the profile, the bank value does not exist at all.

### **Why**

The bank grid is fixed by a file-level constant with no profile backing, at `EllesmereUIBags/EllesmereUIBags_Bank.lua:31`:

```lua
local COLUMNS = 14
```

It is consumed throughout the layout code (`gridW` at :1626 and :2380, col/row math at :1677, :1735 and :1808, `rows` at :1685, :1743 and :1816). Because it is a local rather than a profile key, changing it means editing the addon, and the edit is lost on the next update.

At the default slot size this leaves the bank much narrower than a full Warband bank needs, forcing vertical scrolling on tabs that would fit fine at a wider grid.

### **Precedent**

The bag half of the same module already reads its count from the profile, at `EllesmereUIBags.lua:132`:

```lua
return BP().bagColumns or 12
```

So the two halves of one module are inconsistent today. Outside EUI, other bag replacements derive the count from window width rather than fixing it, and land near 27 columns for a combined bank against about 13 for bags, so a single hardcoded 14 is a poor fit for the bank specifically.

### **Notes**

Passing along what turned up while tracking this down, in case any of it saves time. This looks like it fits existing patterns rather than needing new machinery:

- `BAGS_DEFAULTS` in `EUI_Bags_Options.lua` already holds bank-side keys (`bankSidebarCollapsed` at :26), so there is a home for it.
- The panel already drives the bank live through `_G.EUI_BankFrame` and its refresh entry points (`RefreshBank`, `RefreshIconZoom`, `RefreshTextSizes` at :181-182, :245-246, :261-262, :504-505, :513-514), so a slider could reuse `RefreshBank()` to reflow the grid without a reload.
- Replacing the local with a `GetBankColumns()` reader mirrors `GetColumns()` on the bag side.

Defaulting `bankColumns` to 14 would preserve current behavior exactly for anyone who never touches it, and this adds no events, hooks, frames, or OnUpdate work: it is a read of an existing profile value inside layout code that already runs.

LiQiuDGG:
## What does this PR do?

Two changes in the Bags module.

**Bag Columns / Bank Columns sliders.** Adds one options row exposing the grid width of the bag and bank windows. `bagColumns` already existed in the profile at `12` but was not reachable from the options page. `bankColumns` is new and defaults to `14`, the value `COLUMNS` has always been hardcoded to in `EllesmereUIBags/EllesmereUIBags_Bank.lua`. Both windows therefore look exactly as they did until a slider is moved. The warband bank follows Bank Columns. `bankColumns` is added to the migration profile key list in `EllesmereUI_Migration.lua` so it carries across profiles and presets.

**Window Scale on an open bank.** The Window Scale setter guarded on `_G.EUI_Bank`, which is always `nil`: `EUI_Bank` is a file local, and the frame's global name is `EUI_BankFrame` (`EllesmereUIBags/EllesmereUIBags_Bank.lua:209`). The guard never passed, so dragging Window Scale while the bank was open did nothing. The bank already applies the same `bagScale` value on open (`EllesmereUIBags/EllesmereUIBags_Bank.lua:2374`), so only the live update was affected; closing and reopening the bank picked the new scale up. The fix uses the real global.

Closes #1175.

## How was it tested?

Live on `12.1.0` (client `120100`), running a package of this branch built from `v9.0.0`.

- Bags, bank and warband bank at default and raised column counts
- Window Scale dragged with the bank already open, before and after the fix
- `/console scriptErrors 1` for the whole session, no errors
- Auto-Size to Fit on: Bag Columns behaves as the minimum its tooltip describes, with Auto-Size adding columns past it and never dropping below it
- Settings survive `/reload`
- `bankColumns` carries across a profile switch
- Bank Columns changed while the bank is closed applies on the next open

## Screenshots

Before, on stock `v9.0.0`:

**1.** Bank open, Window Scale `100`.

before, Window Scale 100

**2.** Window Scale `130`. The inventory rescales; the bank, already open, does not.

before, Window Scale 130, bank unchanged

After, on this branch. Each step changes one thing:

**3.** Defaults, Bag Columns `12` and Bank Columns `14`.

after, columns 12 and 14

**4.** Bag Columns `16` and Bank Columns `18`, Window Scale unchanged.

after, columns 16 and 18

**5.** Window Scale `125`, columns unchanged. The bank now rescales.

after, Window Scale 125, bank rescales

**6.** The new options row at its defaults.

new options row

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) — not applicable as written, these are sliders rather than toggles. They default to the values already in use, `12` for bags and `14` for the bank, so nothing changes for anyone who does not move them.
- [x] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built — none of those are added. The bank reads `bankColumns` inside its existing refresh path.
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations) — one upvalue assignment per bank refresh.
- [x] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames — not applicable, only EllesmereUI's own frames are touched.

### DIFF
--- a/EllesmereUIBags/EllesmereUIBags_Bank.lua
+++ b/EllesmereUIBags/EllesmereUIBags_Bank.lua
@@ -1612,6 +1612,7 @@ end
 --  Refresh
 -------------------------------------------------------------------------------
 function EUI_Bank:RefreshBank()
+    COLUMNS = BP().bankColumns or 14
     if not EUI_Bank:IsVisible() then return end
     NotifyBankTypeForTSM()
 
@@ -2386,6 +2387,7 @@ eventFrame:SetScript("OnEvent", function(_, event)
             end
         end
         -- Set initial size so frame is visible immediately
+        COLUMNS = BP().bankColumns or 14
         local gridW = COLUMNS * (SLOT_SIZE + SPACING)
         EUI_Bank:SetWidth(GetBankSidebarWidth() + gridW + 10 * 2 + SCROLLBAR_HIT_W + 2)
         EUI_Bank:SetHeight(FIXED_H)
--- a/EllesmereUIBags/EllesmereUIBags_DB.lua
+++ b/EllesmereUIBags/EllesmereUIBags_DB.lua
@@ -12,6 +12,7 @@ local BAGS_DEFAULTS = {
         bagScale              = 1,
         bagItemIconZoom       = 0.08,
         bagColumns            = 12,
+        bankColumns           = 14,
         bagAutoSize           = false,
         bagCatTitleSize       = 11,
         bagCountFontSize      = 11,
--- a/EllesmereUIOptions/EUI_Bags_Options.lua
+++ b/EllesmereUIOptions/EUI_Bags_Options.lua
@@ -81,7 +81,7 @@ initFrame:SetScript("OnEvent", function(self)
                       if _G.EUI_Bags then _G.EUI_Bags:SetScale(s) end
                       if _G.EUI_BagsReagent then _G.EUI_BagsReagent:SetScale(s) end
                       if _G.EUI_BagsWindow then _G.EUI_BagsWindow:SetScale(s) end
-                      if _G.EUI_Bank and _G.EUI_Bank:IsVisible() then _G.EUI_Bank:SetScale(s) end
+                      if _G.EUI_BankFrame and _G.EUI_BankFrame:IsVisible() then _G.EUI_BankFrame:SetScale(s) end
                   end },
                 { type="slider", text="Icon Zoom", min=0, max=0.20, step=0.01,
                   tooltip="Crops the border of every item icon in bags and bank. 0 shows the full icon.",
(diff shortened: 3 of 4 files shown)


toanqc:
I just tried this function and it looks great but can we have the options for the row as well?

My Bank looks like this. Would love if we could have a row options at least for Bank
image

LiQiuDGG:
> I just tried this function and it looks great but can we have the options for the row as well?
> 
> My Bank looks like this. Would love if we could have a row options at least for Bank image

I want to make sure I understand what you're asking for.  you want a slider to set the number of rows as well, so that you can set them, and if say you pick 20 rows, any more then 20 is a scroll bar?

toanqc:
Yes

LiQiuDGG:
I'll see what i can do.

toanqc:
look like the bank columns stop working?

LiQiuDGG:
I haven't tested since the new builds have dropped.  It's on my list.

toanqc:
Ah sorry my bad, I copied to the wrong folder

PR #1719, opened 2026-08-24
Unable to track properly sweeping strikes due to API limits

luispemu:
Removing Sweeping Strikes as resource bar as we cannot read or track properly stacks due to 12.1 API limits


EllesmereGaming:
wait when did this happen? last I checked this was still working. are both arms and fury broken or what is it specifically that changed

EllesmereGaming:
fyi i have rewritten all sweeping strikes / whirlwind logic and it's now guaranteed accurate 100% of the time driven by the aura itself. live with next patch (9.0.9)

luispemu:
The issue is only for Arms warrior and the issue is that I cannot guarantee a second target in 8yd if not using IsSpellInRange but warrior only knows one spell to calculate the 8yds that is Intimidating Shout that is a node not usualy selected.

So the tracking of the Sweeping Charges had multiple defects and basically missess charges vs aura.

And I thought also on using the aura but running in combat state 

/dump C_UnitAuras.GetPlayerAuraBySpellID(260708)
image

I even checked that /dump C_Secrets.ShouldSpellAuraBeSecret(260708) is false out of combat but true in combat

image


So before having a buggy tracker I thought would be better to not have it at least by now.

Don't know what you think though? We could maybe ask for it to be Whitelisted as Maelstrom Weapon charges?

The other option is to read the combat-log if possible but I think it would be too much CPU consuming

PR #1723, opened 2026-08-25
Raid Frames: fix absorb coalescer stutter from 9.0.1

dfrisone:
```
Cause: the 9.0.1 absorb paint coalescer had a staleness margin (0.45s)
shorter than its own belt ticker period (0.5s), so any shielded raid
member was resweept on nearly every tick instead of the rare no-event
case it was built for, and the flush drained the whole dirty set in
one uncapped frame -- exactly the "Raid, zero events, 128ms" pattern
in the report.

Fix: bound the flush to a per-frame budget, derive the belt's margin
from its ticker period instead of two independent constants, and
re-read each button's current unit at flush time instead of trusting
a token captured at mark time (closes a roster-reassignment race the
multi-frame drain would otherwise widen).

Test: tested in-game in raid, stutter no longer reproduces.

Reported by @Nenpo.
```


EllesmereGaming:
manually merged since only wanted 1/2 of it

PR #1733, opened 2026-08-25
feat(raidframes): add "in other phase" and "in other party" indicators

manaste:
## What does this PR do?

Adds **“In Other Phase”** and **“In Other Party”** indicators to raid and party frames, analogous to the default Blizzard frames.

They use the existing ready-check slot with the following priority:

**Ready Check > Pending Summon > Incoming Rez > Other Phase > Other Party**

The two new indicators can be toggled independently in the **Status Icon** settings and are **off by default**. The previous **“Ready Check / Summon / Rez”** setting has been renamed to **“Status Icon”** to better reflect its purpose. Tooltips have also been added to both the **Status Icon** and **Status Text** settings to make the distinction between them clear.

The indicators replicate the corresponding Blizzard behavior and provide a tooltip explaining why the player is in another phase or party. The tooltip respects the **“Show Raid Frames Tooltip”** setting.

The **“In Other Phase”** indicator is gated behind a cached distance check, matching Blizzard's behavior.

German localization for the new settings has also been added.

To accommodate the additional possible states in preview mode, the indicator handling used by preview mode has been refactored into a simpler system.

## How was it tested?

Tested on the live 12.1 client, build 9.0.0.

* Checked that players in the same party/raid who are in a different phase (e.g. war mode) and within the relevant distance show the **“In Other Phase”** icon.
* Checked that when one player is in an instance group and the other is not (e.g. one player is in a delve and the other is not), the **“In Other Party”** icon is shown.
* Checked that both indicators display the appropriate tooltip on hover and that the tooltip respects the **“Show Raid Frames Tooltip”** setting.
* Checked that the indicators respect their individual on/off settings and are disabled by default.
* Checked that the indicator priority is respected when multiple states apply.
* Checked the new states in preview mode.
* Checked that the existing Ready Check, Summon, and Rez indicators retain their expected behavior.

## Screenshots

In other phase:
other_phase

In other party with tooltip:
other_group_tooltip

Settings:
settings

## Checklist


manaste:
☑️ Retested with 9.0.7

manaste:
There at least two feature requests for this:

https://discord.com/channels/585577383847788554/1537606324823793714
https://discord.com/channels/585577383847788554/1547799873129025697

PR #1735, opened 2026-08-25
fix(nameplates) (regression) Friendly name size not respect the options settings

toanqc:
## What does this PR do?
v9.0.2 introduces a regression bug where the friendly name size is not respect the options and always reset to the default font size.
This PR fixes it.

## How was it tested?

Tested in-game

## Checklist


EllesmereGaming:
this was actually already fixed a few hours ago by another contributor but thank you!! will be live in a few minutes

PR #1740, opened 2026-08-25
[bugfix] Fix Vault font/icon coloring & Character Screen fonts

lolswirl:
## What does this PR do?

Applies global/module settings to Character Sheet and Great Vault window skins
Also fixes the coloration of the default "white" on the icon borders of great vault items to their rarity

## How was it tested?

Live on 12.1.0.69497

## Screenshots
Settings:
image

### Character Sheet
Before:
image

After:
image

### Great Vault
Before:
image

After:
image

### Different settings
Settings:
image
image
image

item coloration fix:
image

## Checklist


EllesmereGaming:
added the color fix for great vault items but the rest is a bit more complex that I need some more time to work on

PR #1751, opened 2026-08-26
Fix spec data bar sticking to failed talent swap after combat interrupt

dfrisone:
## Summary
When a player clicks to swap talent loadouts, the spec bar writes the new config ID optimistically before the server confirms. If combat interrupts the swap and causes CONFIG_COMMIT_FAILED, the bar now correctly reverts the pointer instead of sticking on the never-applied loadout.

## Changes
- Track pending swap state (spec ID, config ID, previous config ID)
- Revert pointer when CONFIG_COMMIT_FAILED event fires
- Extract swap sequence into BeginLoadoutSwap() helper for clarity

Fixes #[[reported by Tephrite on 2026-08-25]]


EllesmereGaming:
had to fix this a slightly different way

PR #1752, opened 2026-08-26
Fix action bar SetAttribute spam in combat

dfrisone:
Bug: https://discord.com/channels/585577383847788554/1541517868771053658 (reported by @Jera, 9.0.1)

Issue: with bars hidden or on mouseover, pulling a raid boss spams ADDON_ACTION_BLOCKED naming EllesmereUI and ActionButton3:SetAttribute(). Only players with an assisted-combat action on a bar hit it, which is why it read as an assist-highlight issue.

Root cause: EUI kills ActionBarButtonEventsFrame at load and re-registers a slice of it in ApplyBroadcaster, so the frame's whole dispatch loop runs under our taint and Blizzard's own buttons get blocked. ACTIONBAR_SLOT_CHANGED is the only event in either set that reaches Update -> UpdatePressAndHoldAction -> SetAttribute, and an assisted action dirties its slot many times a second, so the pull spams it.

Fix: register ACTIONBAR_SLOT_CHANGED out of combat only, with both edges driven by PLAYER_REGEN_DISABLED/ENABLED and handled undeferred so the pull's first slot change cannot slip through. A slot whose contents change mid-combat is not re-derived on the Blizzard twins until the next out-of-combat change or zone-in. Replaces the earlier UpdatePressAndHoldAction wrapper, which added taint of its own.


EllesmereGaming:
this fix would cause taint issues so can't merge as is, need more information from the bug reported, from what i can see it points more towards assisted highlight. i assume this is this report? https://discord.com/channels/585577383847788554/1541512123518296104

Kulia240:
> this fix would cause taint issues so can't merge as is, need more information from the bug reported, from what i can see it points more towards assisted highlight. i assume this is this report? https://discord.com/channels/585577383847788554/1541512123518296104

https://discord.com/channels/585577383847788554/1541517868771053658 - this is the report.

dfrisone:
Will check today right now!

dfrisone:
You were right on both counts, and they turn out to be two ends of the same bug. I've replaced the fix and updated the PR.

The wrapper did add taint, so that is gone. Digging further, our own registration was the conduit: EUI unregisters ActionBarButtonEventsFrame at load and re-registers a slice of it in ApplyBroadcaster, and because the registration is ours, the mixin's OnEvent dispatch loop runs under our taint. That is why a Blizzard-owned button like ActionButton3 is the one getting blocked. ACTIONBAR_SLOT_CHANGED is the only event in either registration set that reaches Update -> UpdatePressAndHoldAction -> SetAttribute; the others take paint or read branches, and PLAYER_ENTERING_WORLD cannot fire under lockdown.

The assisted highlight is the other half, so your read was right too. An assisted-combat action dirties its slot many times a second, which is the event storm that turns one blocked write into spam, and Blizzard's "arg1 == 0 or arg1 == self.action" gate is why the error names a single button (slot 3 being where that player's assist action sat).

The fix is now just: register ACTIONBAR_SLOT_CHANGED out of combat only, in both broadcaster modes, with both edges driven off PLAYER_REGEN_DISABLED/ENABLED and handled undeferred, since the existing handler's C_Timer.After(0) would let the pull's first slot change through. The cost is that a slot whose contents change mid-combat re-derives on Blizzard's twin buttons at combat end.

Not verified in a raid yet on my end, I will confirm that before it is worth your time.

dfrisone:
On why the first attempt was wrong and why this one holds.

The old one wrapped btn.UpdatePressAndHoldAction on ActionButton1-12 with a closure that skipped the write while InCombatLockdown(). Two things wrong with it. Writing our own function into a field on a secure Blizzard button taints that button, so it spread taint in order to hide a taint symptom, which is your objection. And it only silenced the single call site at the top of Jera's stack while leaving the dispatch loop underneath still running under our taint, so any other protected call Blizzard ever reaches from that OnEvent would block the same way and still be reported as EllesmereUI. It treated the last frame of the stack instead of the first.

The new one works at the point where the taint actually enters. The dispatch is tainted because the registration is ours, so the event that can reach a protected write simply is not registered during the window where that write is illegal. Nothing of ours is written onto a Blizzard frame, no function is replaced, and Blizzard's handler runs exactly as it always did. The block cannot recur from a different path either, because the whole dispatch is off for that event rather than one function being neutered.

What the "ph" mode exists for is unaffected. Press-and-hold still seeds out of combat and on every zone-in through the PLAYER_ENTERING_WORLD branch, which updates every button ungated, and the broadcaster stays dead for everything else so the perf work stands.

The honest limitation: a slot whose contents change mid-combat is not re-derived on the Blizzard twins at the moment combat ends, only on the next out-of-combat change to that slot or the next zone-in. Since actions cannot be placed on a bar in combat anyway, that window is narrow, and it self-heals on the next loading screen.

PR #1755, opened 2026-08-26
feat(qol): add Compact Band layout to Raid Tools

0x963D:
## What does this PR do?

Adds an opt-in **Compact Band** choice to Raid Tools' existing **Show as** setting. **One Window remains the default**, and every existing layout keeps its current behavior.

Compact Band is a single resizable row containing Blizzard's raid-marker and clear icons, **RC** for Ready Check, and **PT** for the configured pull timers. Marker buttons keep the existing target/world mouse mappings. The band registers as a normal Unlock Mode element, so its standard cog provides width, height, anchoring, and size matching.

This PR contains no Chat integration, bundled icon assets, or Quick Fire bindings. RC/PT use text so project-owned icons can be supplied later if desired.

## How was it tested?

Tested in-game on live Retail Midnight 12.1 after `/reload`.

- Checked marker order and the target/world marker mouse actions.
- Checked RC/PT controls and the two-column house tooltip.
- Checked Unlock Mode width, height, and positioning.
- Moved the unanchored band, saved, and confirmed its position survives `/reload`.
- Parsed both changed Lua files as Lua 5.1.
- Ran `git diff --check`, the ASCII scan, locale regeneration, and the UI consistency detector.
- Verified that the branch changes only the Raid Tools runtime and options files.

## Screenshots

**Before (v9.0.6):** Raid Tools has no Compact Band choice. One Window remains the default and is unchanged by this PR.

**After — Compact Band as a standalone Unlock Mode element:**

Compact Band shown as a standalone Unlock Mode element

**After — the same band positioned independently above Chat:**

Compact Band positioned independently above Chat

The PR does not read or modify Chat.

## Checklist

- [x] No writes onto Blizzard-owned frames. Weak-table handling and Blizzard-frame hooks are not applicable; the band uses addon-owned frames only
- [x] Tested in-game, works on live Retail Midnight 12.1 with no load errors

### DIFF
--- a/EllesmereUIOptions/EUI_QoL_RaidTools_Options.lua
+++ b/EllesmereUIOptions/EUI_QoL_RaidTools_Options.lua
@@ -54,10 +54,24 @@ initFrame:SetScript("OnEvent", function(self)
         return (Cfg("mode") or "never") == "never"
     end
 
-    -- The runtime owns the normalize (unknown values read as "one").
+    -- The runtime owns the normalize (unknown values retain One Window).
     local function ShowAsVal()
         if ns.ShowAs then return ns.ShowAs() end
         return Cfg("showAs") or "one"
+    end
+
+    local function LegacyDisabled()
+        return Disabled() or ShowAsVal() == "compact"
+    end
+    local function RoleCheckDisabled()
+        return Disabled() or ShowAsVal() == "markers"
+    end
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
excellent work, ty for splitting up the PRs!

PR #1764, opened 2026-08-26
Fix Mirror Key Presses not matching item-based actions

svart2521:
Bug: https://discord.com/channels/585577383847788554/1541691517218066502

Issue: "Mirror Key Presses" on a custom Cooldown Manager bar wasn't flashing the mirror overlay for trinkets, potions, or healthstones — the actual keybind still fired the item fine, just no visual mirror. Toggling other bar settings had no effect because the bug wasn't a settings issue at all.

Root cause: The press-mirror code (SlotSpellID in EllesmereUICdmHooks.lua) only knew how to resolve action bar slots holding a spell or macro. Any slot holding an item directly (trinket, potion, healthstone) reports actionType == "item", which fell through to nil — silently killing the mirror for every item-tracked icon.

Fix: OnPress now also resolves item-type presses and matches them against how this addon actually keys item-based bar icons — a negative identity (-itemID for potions/healthstones/custom items, -invSlot for equipped-slot tracking like trinkets' -13/-14) — plus the item's on-use spell ID as a fallback for the rarer case where it's tracked as a raw Custom Spell entry.

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
+++ b/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
@@ -10283,6 +10283,14 @@ do
         local actionType, id, subType = GetActionInfo(slot)
         if actionType == "spell" then
             return id
+        elseif actionType == "item" then
+            -- On-use trinkets/potions/healthstones placed directly on the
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
BOOM ANOTHER ONE

PR #1765, opened 2026-08-26
Add all six glow styles plus a live preview to Important Cast Glow

vcherneny:
## What does this PR do?

Important Cast Glow on enemy nameplates offered 2 styles. It now offers the
same 6 as Pandemic Glow (Pixel, Action Button, Auto-Cast Shine, GCD, Modern
WoW, Classic WoW), plus a live preview icon in the row so you can see the glow
while picking it.

`UpdateImportantCastGlow` hardcoded two engines behind an ad-hoc numbering
(1 and 4); it now renders through `Glows.StartGlow` like the pandemic glow
does. Pixel Glow output is unchanged. `NP_TO_SHARED_GLOW` moves next to
`PANDEMIC_GLOW_STYLES` so the aura containers and the cast glow share one copy.

Old saved value `4` meant Auto-Cast Shine but means GCD in the new numbering,
so `np_important_cast_glow_style_reindex_v1` re-points it to `3`. Nothing
changes for existing users.

## How was it tested?

Client 12.1.0.69497.

 Cycled all 6 styles on an enemy important cast; colour swatch
and the Pixel Glow cog settings update plate + preview live; "None" clears it.
Set Auto-Cast Shine on the old build, updated, confirmed it stayed.

Offline: verified the index translation by name against all three style tables;
ran the migration body against real SavedVariables (idempotent, no-ops on
profiles without the key).

 ## Screenshots

glow-ingame-2
 
glow-ingame-4

## Settings

https://github.com/user-attachments/assets/f4e5fb76-b1ad-4383-ac78-04246762b3b3

## Checklist

- [x] New settings default **OFF** — N/A, no new setting. `importantCastGlow`
      (true) and `importantCastGlowStyle` (1) are pre-existing defaults, untouched.
- [x] Zero cost while disabled — returns before building the overlay when off;
      preview only exists while the LoD options panel is open.
- [x] Cheap while enabled — no new events/hooks/timers; uses the existing shared
      Glows driver. The 3 FlipBook styles are C-side, cheaper than what was there.
- [x] No writes onto Blizzard-owned frames — EUI-owned overlay on the EUI cast bar.
- [ ] Tested in-game on live — no version gates or pre-Midnight APIs; tick after
      the in-game pass.

### DIFF
--- a/EllesmereUINameplates/EUI_Nameplates_AuraContainers.lua
+++ b/EllesmereUINameplates/EUI_Nameplates_AuraContainers.lua
@@ -289,9 +289,9 @@ function NPB.GroupGlow(idx)
     return magic == true, "magic"
 end
 
--- PANDEMIC_GLOW_STYLES index -> shared EllesmereUI.Glows.STYLES index (the
--- NP list has no Shape Glow, so the flipbook entries sit one lower there).
-local NP_TO_SHARED_GLOW = { 1, 2, 3, 5, 6, 7 }
+-- PANDEMIC_GLOW_STYLES index -> shared EllesmereUI.Glows.STYLES index. Defined
+-- beside PANDEMIC_GLOW_STYLES in EllesmereUINameplates.lua (loaded before us).
+local NP_TO_SHARED_GLOW = ns.NP_TO_SHARED_GLOW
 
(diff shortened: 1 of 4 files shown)


EllesmereGaming:
good work, thank you!

PR #1766, opened 2026-08-26
Fix Show All on Mouseover triggering from non-Mouseover bars

svart2521:
Bug: https://discord.com/channels/585577383847788554/1542132572916547635

Issue: With "Show All on Mouseover" enabled, hovering any action bar — even one set to Always visible — made all Mouseover-visibility bars appear. The tooltip promises this should only happen when hovering a bar that's itself set to Mouseover.

Root cause: AttachHoverHooks wires the hover-fade handler onto every action bar unconditionally, regardless of that bar's own visibility setting. So FadeIn's "show all" broadcast fired on mouseoverShowAll alone, with no check on whether the bar actually being hovered was a Mouseover bar.

Fix: FadeIn now only broadcasts to the other bars when the entered bar itself has mouseoverEnabled set — matching the setting's documented behavior.


EllesmereGaming:
BOOM UR ON FIRE

PR #1780, opened 2026-08-27
Add gold amount abbreviation option to Databars

Barbiero:
## Summary
- Adds an "Abbreviate Amount" toggle to the Databars Gold block, showing large balances with K/M suffixes (e.g. `284,208g` → `284.2Kg`) instead of the fully grouped number. The gold tooltip breakdown (session earned/spent, per-character list, warbank, total) always keeps full precision regardless of this setting.
- CJK clients (zhCN/zhTW/koKR) group by 万/萬/만 instead of K/M, matching how the client's own number formatting reads for those locales. There's no "B"/亿 tier since the gold cap (99,999,999) never reaches it.

The CJK abbreviation table/config-building approach is copied from the existing pattern in `EllesmereUIDamageMeters/EllesmereUIDamageMeters.lua` (`BuildAbbrevOpts`/`RebuildAbbrevCfg`/`AbbrevNumber`, around line 960 onward), reused here in a simplified, gold-specific form so both modules format abbreviated numbers consistently.

## Test plan
- [x] In-game: toggle "Abbreviate Amount" on the Gold block and confirm the bar shows `284.2Kg`-style text while hovering still shows the exact amount.
- [x] Confirm the gold tooltip (session/characters/warbank/total) still shows full precision with the toggle on.
- [x] Confirm vertical bar layout (side bar) also abbreviates correctly.
- [x] Switch client locale to a CJK locale (or `/console locale koKR`-style test) and confirm 万/萬/만 grouping appears instead of K/M.

### DIFF
--- a/EllesmereUIDataBars/EllesmereUIDataBars.lua
+++ b/EllesmereUIDataBars/EllesmereUIDataBars.lua
@@ -221,7 +221,7 @@ ns.BLOCK_DEFAULTS = {
     -- which the options page pulses about.
     location   = { showIcon = true, showSubZone = true },
     coords     = { showIcon = true, precision = 0, hideInInstance = true },
-    gold       = { showIcons = true, showBagSpace = false, showSmall = false, coinIcons = false, abbreviate = false },
+    gold       = { showIcons = true, showBagSpace = false, showSmall = false, coinIcons = false, abbreviate = false, forceEnglishUnits = false },
     durability = { showIcon = true },
     combat     = { onlyInCombat = false },
     xprep      = { mode = "auto" },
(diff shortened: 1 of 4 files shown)


Barbiero:
question for @EllesmereGaming if you end up accepting this PR: should we centralize locale-specific number formating(CJK formatting) somewhere instead of having duplicate code between gold databars and damage meters?

Should this option be applicable to bags as well? (I don't really use the bag module so I'm not sure if it'd be desirable)

EllesmereGaming:
merged ty! I could see that being useful but for now I think we just leave it as is and see if there's any desire for that in bags as well. somewhat low priority for me currently

PR #1797, opened 2026-08-27
Fix: Range Check for Whirlwind on Twin Fangs

### LINKED ISSUE #1796: Fury Warrior: Improved Whirlwind stack tracker never awards stacks on large-hitbox bosses (Twin Fangs)
Version: 9.0.7
Spec: Fury Warrior, Improved Whirlwind (12950) talented

The Improved Whirlwind stack tracker (Fury Warrior class resource) sits permanently at 0 on certain raid bosses. The affected bosses all share one trait: an oversized model. Twin Fangs in The Venomous Abyss is the clearest case; a handful of last-tier bosses show it too.

```EllesmereUI.lua:4749``` the InReach helper inside the Improved Whirlwind block gates every stack award on:

```return CheckInteractDistance(u, 2) or (wide and CheckInteractDistance(u, 1)) or false```

CheckInteractDistance measures centre-to-centre distance and does not account for the target's combat reach. Melee range is 5 yd + attacker reach + target reach. On a normal-sized mob the two roughly agree, so the probe works. On a boss with a huge model the player stands in melee while the unit's centre is 15–25 yd away — well past the ~9.9 yd that index 2 bottoms out at.

Consequence: EnemyInStrikeRange (:4755) returns false for every Whirlwind cast, so stacks is never set to MAX. The nameplate fallback loop uses the same InReach, so it fails identically, and the deferred PLAYER_STOPPED_MOVING re-probe just re-runs the same failing check. The bar reads 0 for the entire fight.

The Arms Sweeping Strikes tracker in the same file hit this exact wall and was already migrated away from CheckInteractDistance in [#882](https://github.com/EllesmereGaming/EllesmereUI/pull/882) ("melee-range cleave gauge"). Its InReach at :5000 uses a hitbox-scaled C_Spell.IsSpellInRange probe on Mortal Strike, and the comment at :4995 explicitly notes that CheckInteractDistance bottoms out at ~9.9 yd. The Whirlwind block never received the same treatment.

bjth:
## What does this PR do?

Improved Whirlwind for Fury. The built in tracker works for large bosses like Twin Fangs.

## How was it tested?

Local build. LFR Twin Fangs. Random Delve to make sure normal tracking works.

## Checklist

- [ N/A ] New settings default **OFF** (no behavior change without opt-in)
- [ N/A ] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built
- [ N/A] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations)
- [ N/A] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames
- [ Yes] Tested in-game on live; no version gates or pre-Midnight APIs added

Fixes #1796

### DIFF
--- a/EllesmereUI.lua
+++ b/EllesmereUI.lua
@@ -4740,8 +4740,7 @@ do
 
     -- Improved Whirlwind grants stacks only when the swing connects, but
     -- UNIT_SPELLCAST_SUCCEEDED fires on a whiff too: gate the award on an attackable
-    -- living enemy inside the strike radius. Whirlwind is a ~8 yd self-AoE; the index-2
-    -- probe (~11 yd) is slightly generous and resolves on hostile nameplates; Thunder
+    -- living enemy inside the strike radius. Whirlwind is a ~8 yd self-AoE; Thunder
     -- Clap/Blast reach farther with Crackling Thunder. Resolved synchronously at cast
     -- time so a combat-ending kill still counts; with no hostile target it relies on
     -- enemy nameplates showing. InReach is block-scoped (no upvalues) so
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
sweeping strikes and whirlwind bar logic has been fully reworked in upcoming patch (9.0.9) to use the aura as the source of truth resulting in 100% accuracy and better performance! fix is much appreciated though would've happily merged if I hadn't already been reworking this

PR #1817, opened 2026-08-29
feat(cdm): add per-buff stack glow thresholds

0x963D:
## What does this PR do?

Adds an opt-in per-buff `Glow at Stacks` threshold under the existing Charge/Stack Text settings. When a threshold of 2 or higher is configured, the selected Buff Glow starts at that stack count; `0`, `1`, or an unset value keeps the existing behavior.

The threshold path is safe for Midnight secret stack values: a one-unit `StatusBar` range performs the comparison in the UI engine and clips an addon-owned glow wrapper. It reuses the existing native glow implementation so Auto-Cast Shine, Pixel Glow, Shape Glow, and the other styles retain their normal size and settings.

No frame, event, timer, or hook is added while the option is disabled. Enabled icons reuse the existing dirty-gated buff update path.

## How was it tested?

- Live Retail / Midnight 12.1 with Soul Fragments (spell 1245577), Auto-Cast Shine, and a 4-stack threshold.
- Verified the opt-in setting, below/above-threshold behavior, glow-style changes, and reload persistence in game.
- Parsed all three changed files as Lua 5.1 with `luaparse`.
- Ran `git diff --check`, an ASCII-only scan, and `.tools/extract-locale-keys.sh` (765 keys, no generated diff).

## Screenshots

Before (ordinary buff tracking, no threshold glow):

Before: Soul Fragments without a threshold glow

Opt-in setting:

Glow at Stacks input under Charge/Stack Text

After (Auto-Cast Shine active above the configured threshold):

Soul Fragments with Auto-Cast Shine at seven stacks

## Checklist

- [x] Cheap while enabled: reuses the existing dirty-gated buff update path; no new polling, timer, or per-frame allocation path
- [x] No writes onto Blizzard-owned frames (external weak-table state; `HookScript` only)

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
+++ b/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
@@ -38,6 +38,132 @@ ns._spellOrderDirty = true  -- start dirty so first reanchor builds caches
 local hookFrameData = setmetatable({}, { __mode = "k" })
 ns._hookFrameData = hookFrameData
 
+-- Stack counts may be secret in restricted combat. A one-unit StatusBar range
+-- performs the threshold comparison in C; its fill clips the glow at the edge.
+-- Frames are created only for icons with an enabled per-spell threshold.
+local function StackGlowSize(icon)
+    local width, height = icon:GetWidth(), icon:GetHeight()
+    if not width or width < 5 then width = 36 end
+    if not height or height < 5 then height = width end
+    return width, height
+end
+
+local function StartStackGlow(st, width, height)
+    local opts = {
+        owner = st.icon, width = width, height = height,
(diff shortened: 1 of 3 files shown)


EllesmereGaming:
can you type /euidev and then tell me if this works while in combat?

EllesmereGaming:
nvm did my own testing and it's working! only issue is that this will put the file over 200 locals so I can't merge as is, will apply the fixes locally and it'll be live with next patch thank you!!

0x963D:
> nvm did my own testing and it's working! only issue is that this will put the file over 200 locals so I can't merge as is, will apply the fixes locally and it'll be live with next patch thank you!!

Thanks for testing it and for including it in the next patch! 
Sorry about the extra work, I'll make sure to test with /euidev going forward.

PR #1823, opened 2026-08-29
refactor(core): centralize number abbreviation into EllesmereUI_NumberFormat

Barbiero:
## Summary
- Moves the duplicated CJK/K-M-B breakpoint tables and AbbreviateNumbers plumbing out of DamageMeters and DataBars into a single shared EllesmereUI_NumberFormat.lua
- Each CJK locale file (zhCN, zhTW, koKR) registers its own abbreviation algorithm, so it ships inside that locale's load-on-demand file and never loads for English clients
- Options pages now query whether the current locale has its own abbreviation algorithm instead of hardcoding the zhCN/zhTW/koKR list
- The hot AbbreviateNumber() call path caches its resolved config instead of rebuilding it on every call

## Test plan
- Rebased onto latest main and resolved the resulting conflicts: DataBars' gold abbreviation now delegates fully to the shared engine, and its Force English Units option keeps the already-translated "(K/M/B)" wording while switching to the new generic locale-capability check
- This is a straight code-movement refactor with no intended behavior change
- Tested in-game: I've been using this branch for a while even before v9.0.8 released

## Caveat about API

`CreateAbbreviateConfig` and `AbbreviateNumbers` are supported on all modern clients by default, so I opted to drop the usual "does this api exist"(eg. `AbbreviateNumbers and AbbreviateNumbers()`) because having a fallback for "this api does not exist" ended up being just a bunch of dead code. I don't believe these functions will be removed anytime soon, so checking for their existence and having a fallback *just in case* made the code more complex than it needed to be.

### DIFF
--- a/EllesmereUI.toc
+++ b/EllesmereUI.toc
@@ -27,6 +27,10 @@ EllesmereUI_Lite.lua
 # loaded by the engine at ADDON_LOADED for non-English locales only.
 EllesmereUI_Locale.lua
 
+# Central number-abbreviation engine (K/M/B, or a locale's own algorithm --
+# e.g. CJK 万/亿 grouping -- registered from EllesmereUILocales/<code>.lua).
+EllesmereUI_NumberFormat.lua
+
 # One-time data migration (runs before child addons init)
 EllesmereUI_Migration.lua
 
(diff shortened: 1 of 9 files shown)


Barbiero:
continuing from #1780 :)

PR #1837, opened 2026-08-30
Fix tooltip visibility and item comparison flashing

jixinliu666:
## What does this PR do?

Fixes tooltip visibility and item-comparison lifecycle issues when the existing "Show Tooltips" mode is active.

- Prevents delayed comparison cleanup from clearing a newly allowed tooltip.
- Applies the existing EUI tooltip visibility mode to explicitly-owned tooltips, including bag container items and character equipment slots.
- Re-applies suppression when item data asynchronously re-shows GameTooltip.

## How was it tested?

Tested in World of Warcraft Retail:

- "Always": tooltips remain visible in combat, equipment comparison tooltips stay hidden, and the tooltip no longer flashes.
- "Out of Combat": bag and character-panel tooltips remain hidden during combat.

### Bag: Always in combat

Before: bag item tooltip and comparison visible

After: no comparison tooltip and no flashing

### Bag: Out of Combat mode

After: no tooltip on bag item hover during combat

### Character panel: Out of Combat mode

Before: character panel tooltip visible during combat

After: no character panel tooltip during combat

## Checklist

- [x] No new settings or default behavior changes; uses the existing "Show Tooltips" setting.
- [x] Zero additional cost while the existing mode is disabled; no new always-on frame, event, polling, or hook was added.
- [x] Event-driven while enabled; no OnUpdate or wall-clock logic gate was added.
- [x] No new Blizzard-frame state or SetScript usage was introduced; the existing tooltip suppression controller is reused.


jixinliu666:
Superseded by the split PRs #1838 (comparison cleanup) and #1839 (tooltip owner synchronization).

PR #1846, opened 2026-08-30
Fix stutters while the friends list is open

Nnoggie:
With the Friends List module enabled, leaving the friends list open causes repeated stutters. NAP shows spikes of over 100 ms attributed to EllesmereUI. Disabling the module makes them stop.

The list was checking rows every frame and rebuilding the entire friends cache on status updates. This removes the polling, combines repeated updates, and only fetches friend data for the rows being displayed. Search still includes the full friends list.

To reproduce:
1. Open the friends list.
2. Set NAP to **Since Reset**.
3. Leave the list open and observe the stutters and repeated spikes under EllesmereUI.

### DIFF
--- a/EllesmereUIFriends/EllesmereUIFriends.lua
+++ b/EllesmereUIFriends/EllesmereUIFriends.lua
@@ -535,12 +535,23 @@ local function BuildClassNameLookup()
     end
 end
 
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
this was supposed to not even be necessary since the new blizzard friends list was supposed to be implemented >:(

thank you for the fix lol hopefully they decide to actually launch the friends list soon.

PR #1850, opened 2026-08-30
Add vehicle (and possess) visibility condition to action bars

sgtoj:
## What does this PR do?

Adds a Vehicle visibility condition to Action Bars.

The condition covers vehicle UI, override bars, and possess bars. It provides Show and Hide lanes and supports both Match All Conditions and Match Any Conditions.

The condition is available on Action Bar 1 and on Lua-managed action-bar chrome such as the Micro Menu, Bag Bar, and data bars. It is locked on bars whose existing vehicle rules already force the same behavior.

## How was it tested?

- LuaJIT syntax check passed for all changed files.
- Verified Match All and Match Any secure driver output with a standalone harness.
- Tested configuration in-game on live.
- Tested normal dungeon gameplay

 with no reported errors.
- Vehicle, override-bar, and possess-bar runtime testing is still pending.

## Screenshots

Wow_tUCQxK76Mv

CDVuRKncDm

## Checklist

### DIFF
--- a/EllesmereUI.lua
+++ b/EllesmereUI.lua
@@ -12673,6 +12673,7 @@ EllesmereUI.VIS_OPT_KEYS = {
     "visHideDragonriding", "visOnlySkyriding",
     "visHideNoTarget", "visHideWithTarget",
     "visHideNoEnemy", "visHideWithEnemy",
+    "visOnlyVehicle", "visHideVehicle",
 }
 
 -- Cache player class once at load time (never changes).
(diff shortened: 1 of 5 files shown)


sgtoj:
I called the option `Vehicle` because that is more well known. However, maybe it should be called `Possess`?

EllesmereGaming:
this functionality was actually part of a PR from the dev I have working on the visibility feature (Juju) but excellent work it's much appreciated!

PR #1851, opened 2026-08-30
Add saved outfits to Quickdraw

sgtoj:
## What does this PR do?

Adds Midnight saved transmog outfits to Quickdraw.

Quickdraw can now add an individual outfit from an Outfits picker category or create an Action Menu from the Outfits preset. Each entry uses the saved outfit's own name and icon, with a generic transmogrifier icon as a fallback.

Outfit entries store Blizzard's stable outfit ID and resolve their current player-facing index when secure actions are pushed. This prevents a reordered outfit list from applying the wrong outfit.

## How was it tested?

- LuaJIT syntax check passed for all changed files.
- Tested in-game on live:
  - Individual outfit Quickdraw actions apply the selected outfit.
  - The Outfits preset creates entries for saved outfits.
  - The preset uses the first saved outfit's icon.
- Tested during a +11 Mythic+ dungeon with no known errors.

## Screenshots

Wow_rHMBQ7DtSh

## Checklist

### DIFF
--- a/EllesmereUIOptions/EUI_Quickdraw_Options.lua
+++ b/EllesmereUIOptions/EUI_Quickdraw_Options.lua
@@ -1007,6 +1007,15 @@ initFrame:SetScript("OnEvent", function(self)
         return out
     end
 
+    local function OutfitEntries()
+        local out = {}
+        for _, slot in ipairs(ns.OutfitSlots and ns.OutfitSlots() or {}) do
+            local icon, name = ns.SlotDisplay(slot)
+            out[#out + 1] = { icon = icon, name = name, slot = slot }
+        end
+        return out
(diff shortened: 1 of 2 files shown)


sgtoj:
Closing in favour of #1930, which does this better.

It filters unset slots with the documented `isDisabled` field on `TransmogOutfitEntryInfo` where this PR inferred it from the placeholder icon fileID; resolves through `C_TransmogOutfitInfo.GetOutfitInfo(outfitID)` instead of walking `GetOutfitsInfo()`; guards the second-press toggle with `action = "change"`; and also covers `SlotCooldown`, `SlotUsability` and `SlotFromCursor`, none of which this PR touches.

PR #1863, opened 2026-08-31
Bags: Optional Clear button on the Recent Items header

dillan-prince:
## What does this PR do?

Adds a "Clear" button left of the existing "Hide" button on the Recent Items header, which drops the session's tracked recent set. Opt-in via a new "Show Clear Button" toggle in the Recent Items cog popup. It is further gated on the recent set being non-empty, so there is no dead button when there is nothing to clear. Works in both OneBag and categorized views.

## How was it tested?

Tested on live, version 12.1.0.69497

https://github.com/user-attachments/assets/1944e4e9-e092-4f5a-9adc-378c281fe2bd

## Screenshots

Before, setting disabled:

setting_disabled

After, setting enabled:

setting_enabled

## Checklist


EllesmereGaming:
good one! thank you

PR #1883, opened 2026-08-31
Nameplates: configurable name and title styling for friendly NPC name-only plates

mariusch:
Feature request: https://discord.com/channels/585577383847788554/1517605738107633785

## What does this PR do?

Friendly NPCs in Name Only mode draw their name and `<Title>` in one fixed green.
This makes both configurable.

The **Friendly NPC Settings** cog gains three rows next to Name Size:

| Row | Default |
|---|---|
| Title Size | 10 (was hardcoded) |
| Name Color | green |
| Title Color | green @ 70% alpha |

Defaults match today's look, so nothing changes until you touch a swatch.

Worth flagging:

- **The cog is now hidden in full-plate mode.** Every row in it only affected the
  name-only overlay, so it offered controls that did nothing. Matches the Distance cog
  beside it. Easy to drop.
- **Title colour no longer follows name colour.** Identical at defaults, and the two
  swatches sit together.
- Sliders and swatches re-style live overlays instead of rebuilding them, which was
  re-reading every unit's tooltip mid-drag.

Drive-by: the inline full-plate swatch was missing `showFriendlyPlayers` in its gate, so
it stayed live in a mode where friendly plates don't render. Now uses `friendlyPlateOff()`.

Second commit is separate and touches every colour row in the suite. Disabled
`BuildCogPopup` colourpicker rows dimmed the swatch but left the label lit. Every other
row type dims both through a row overlay; colourpicker now dims by alpha instead, since
an overlay would tint the colour you are reading.

No new locale keys.

## How was it tested?

In game on live. Fresh and existing profiles render identically to `main`. Checked live
slider and swatch updates, cog dimming in all three gating states (NPCs off, friendly
players off, full-plate mode), title rows dimming with Show NPC Titles, profile switch,
spec-override capture, and `/reload` persistence.

## Screenshots

**Nameplate with custom name + title colours**

CleanShot 2026-08-31 at 21 37 47@2x

CleanShot 2026-08-31 at 21 38 15@2x

## Checklist

- [x] New settings default **OFF** -- N/A: colours and sizes, defaulted to the current hardcoded values so appearance is unchanged
- [x] Zero cost while disabled: no new events, hooks or frames; reads happen in the existing paint path
- [x] Cheap while enabled: no polling; option changes re-style live overlays in place
- [x] No writes onto Blizzard-owned frames: the overlay is ours


EllesmereGaming:
nice work!! thank you

PR #1884, opened 2026-08-31
Blizzard Skin: configurable color, size, height and offset for the queue accept countdown

mariusch:
## What does this PR do?

The LFG accept countdown paints its number two hardcoded ways depending on whether the
queue popup reskin is on. This makes it configurable, and the same in both.

A swatch and a **Queue Timer Style** cog sit next to the existing **Show Queue Timer**
toggle:

| Control | Default |
|---|---|
| Countdown Text Color | `#ffd400` |
| Text Size | 9 |
| Bar Height | 11 |
| Text Offset Y | 0 |

Defaults match today's EUI-style look, and all four repaint a countdown already on
screen.

Worth flagging:

- **The style applies with the reskin off too**, since the Show Queue Timer tooltip
  already promises it works either way. With the reskin off the number moves from
  `GameFontHighlight` to the EUI extras font and the bar is 11px rather than 9px. Easy
  to gate to the EUI style instead.
- **The bar is raised above the dialog.** Its parent `bgFrame` deliberately sits a level
  below it so the backdrop draws under the dungeon art, which is correct for the backdrop
  but painted over any raised countdown. Only `timerBar` is lifted.

Defaults live in `EllesmereUI.QUEUE_TIMER` next to `RESKIN`, so the skin and the sliders
cannot drift. New keys join `BLIZZ_SKIN_GLOBAL_KEYS` and the section reset.

Drive-by: `Reskin Popups and Menus` never called `RefreshPage()`, so its border cog and
swatches stayed lit after switching it off until you navigated away and back.

No new locale keys; cog labels resolve through `L(row.label)` at runtime.

## How was it tested?

In game on live against real LFG proposals with the reskin both on and off. Colour, size,
height and offset all apply and repaint live. Fresh profiles render identically to `main`.
Also checked Global Search still indexes the page, cog and swatch dimming, and `/reload`
persistence.

## Screenshots

Now (hardcoded):
CleanShot 2026-08-31 at 22 30 08@2x

Afer:
CleanShot 2026-08-31 at 22 18 55@2x

CleanShot 2026-08-31 at 22 07 44@2x

## Checklist

- [x] New settings default **OFF** -- N/A: colours, sizes and offsets, defaulted to the current hardcoded values so appearance is unchanged
- [x] Zero cost while disabled: no new events, hooks or frames; applied inside the existing `ShowQueueTimer` paint path
- [x] Cheap while enabled: no polling; option changes repaint the existing bar in place
- [x] No writes onto Blizzard-owned frames: the bar, its text and its background are ours


EllesmereGaming:
nice one again! excellent implementations ty

PR #1886, opened 2026-08-31
Unit Frames: option to hide trailing zeros on health percent text

mariusch:
## What does this PR do?

With **Show Decimal on Health Text** on, a unit at full health reads `100.0%`. This adds
**Hide Trailing Zeros** to the **Health Text Decimals** cog, which renders that as `100%`
and leaves `99.5%` alone.

Global, health percent only, default off.

| | decimals off | decimals on | + Hide Trailing Zeros |
|---|---|---|---|
| full health | `100%` | `100.0%` | `100%` |
| damaged | `99%` | `99.5%` | `99.5%` |
| boss, Show 2 for Boss | `99%` | `99.55%` | `99.55%` |
| boss at full | `100%` | `100.00%` | `100%` |

`UnitHealthPercent` is `SecretReturns`, so the zero cannot be trimmed with a `gsub`.
`AbbreviateNumbers` drops a zero fraction and keeps a real one, so the percent goes
through it instead of `"%.1f"`.

It truncates, though, so the obvious config (`significandDivisor = 0.1`) reads a tenth
low on about a third of values: `33.3` renders as `33.2`, `7.1` as `7`. The curve does
the scaling instead. A linear `0 -> 0.5`, `1 -> 1000.5` curve hands `UnitHealthPercent`
whole tenths, so the config divides by integers only, and the `+0.5` puts the truncation
where `"%.1f"` would have rounded. Boss frames use `10000` / `100`.

Worth flagging:

- **Nameplates have the same trailing zero and are not covered.** Their decimals are
  per-slot behind their own cog, and trimming there needs a second `UnitHealthPercent`
  call per plate per update. Own change; happy to follow up.
- **`AbbreviateNumbers` is ~4.3x slower than `string.format`**, 0.94us against 0.22us
  measured over 50k calls. That is ~0.7us per repaint, only while the option is on, and
  the call count is unchanged.
- No migration: an absent `showDecimalTrimZeros` reads as `false`.

Second commit shares one body between `perhp` and `perhpnosign`, which were the same
number written twice, and points the options preview at the same curve and config the
frames use.

One new locale key, `Hide Trailing Zeros`.

## How was it tested?

In game on live (12.1). Verified the formatter first: `1000 -> 100`, `995 -> 99.5`,
`333 -> 33.3`, `71 -> 7.1`, `5 -> 0.5`, and the boss config `10000 -> 100`,
`9955 -> 99.55`, `3333 -> 33.33`. Swept it offline against `"%.1f"` and `"%.2f"` across
~16k health/max combinations with zero drift.

Player and target frames in all four toggle combinations, health falling through
fractional values, `Health %` and `Health % (No Sign)`, `Only Show for % Health` both
ways, options preview against the live frame, toggling without `/reload`, and profile
switch. Health percent is secret on an outdoor enemy target, so that is the path being
exercised.

Not tested on a live `boss1` frame; the 2-decimal curve and config were verified through
the shipped globals.

## Screenshots
Before:
CleanShot 2026-08-31 at 23 37 58@2x

After:
CleanShot 2026-08-31 at 23 36 20@2x

Setting:
CleanShot 2026-08-31 at 23 38 44@2x

## Checklist

- [x] Zero cost while disabled: no events, hooks or frames; one nil global read in the existing percent path
- [x] Cheap while enabled: same call count, the trim curve replaces `ScaleTo100` rather than adding one
- [x] No writes onto Blizzard-owned frames -- N/A, only our own text pieces

### DIFF
--- a/EllesmereUIOptions/EUI_UnitFrames_Options.lua
+++ b/EllesmereUIOptions/EUI_UnitFrames_Options.lua
@@ -1318,7 +1318,9 @@ initFrame:SetScript("OnEvent", function(self)
                 return cfg and AbbreviateNumbers(v, cfg) or AbbreviateNumbers(v)
             end
             local function _pvPct(p01)
-                return _G._EUI_TextDecimals and string.format("%.1f", p01 * 100) or tostring(math.floor(p01 * 100))
+                if not _G._EUI_TextDecimals then return tostring(math.floor(p01 * 100)) end
+                if _G._EUI_PctTrim then return tostring(math.floor(p01 * 1000 + 0.5) / 10) end
+                return string.format("%.1f", p01 * 100)
             end
             local function _pvName()
                 if unitKey == "player" then return UnitName("player") or "Player" end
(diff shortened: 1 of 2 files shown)


EllesmereGaming:
you are on FIRE lfggg nice work

PR #1899, opened 2026-09-01
feat(cdm): add stack glow comparisons

0x963D:
## What does this PR do?

Extends the existing opt-in **Glow at Stacks** setting with a comparison selector:

- Below (`<`)
- At Most (`<=`)
- Exactly (`=`)
- At Least (`>=`)
- Above (`>`)

Existing profiles keep the current **At Least** behavior because a missing operator resolves to `gte`; the default is not written to SavedVariables. Stack Count now accepts 1 because upper-bound comparisons make that value useful.

Restricted-combat stack counts remain secret-safe. The existing native StatusBar/mask gate handles lower and upper bounds; equality lazily intersects a second gate. This adds no event, timer, polling path, or disabled-state work.

**UI wording note for review:** the labels currently pair plain-language descriptions with their mathematical symbols for precision. If word-only labels fit EUI better, I can remove the parenthetical symbols before merge without changing behavior or storage.

## How was it tested?

- Installed the exact commit into Retail and verified the three deployed file hashes.
- Tested in a follower dungeon with **/euidev** enabled; stack comparison glows remained functional with restricted aura values.
- Exercised all five operators for thresholds 1-20 and stack values 0-25 (2,600 cases).
- Parsed all three changed files as Lua 5.1 with `luaparse`.
- Ran `git diff --check` and an ASCII-only scan.

### Before: fixed minimum-stack behavior

Existing fixed stack threshold

### After: comparison control

Stack comparison settings

All comparison choices

## Checklist

- [x] New settings default **OFF** (the existing Glow at Stacks opt-in remains off; missing operator preserves its prior behavior)
- [x] No writes onto Blizzard-owned frames (weak-table state; addon-owned glow textures only); `HookScript` only

### DIFF
--- a/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
+++ b/EllesmereUICooldownManager/EllesmereUICdmHooks.lua
@@ -39,18 +39,16 @@ local hookFrameData = setmetatable({}, { __mode = "k" })
 ns._hookFrameData = hookFrameData
 
 -- Glow at Stacks (per-spell): REPLACES the Buff Glow for icons with a stack
--- threshold -- the glow starts only at N or more applications, using the
--- spell's effective Buff Glow style (resolved in RefreshCDMIconAppearance).
+-- comparison, using the spell's effective Buff Glow style (resolved in
+-- RefreshCDMIconAppearance).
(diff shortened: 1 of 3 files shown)


EllesmereGaming:
excellent work ty!

PR #1930, opened 2026-09-03
feat(quickdraw): add Outfits preset

0x963D:
## What does this PR do?

Adds an **Outfits** preset and picker category to Quickdraw. Each entry uses the saved outfit's name and icon, and outfits can also be dragged directly onto a Quickdraw slot.

Slots store the stable `outfitID`, then resolve Blizzard's current player-facing index before updating the secure action. Outfit changes refresh those indexes through `TRANSMOG_OUTFITS_CHANGED`.

The secure button uses `action="change"` so selecting the active outfit again does not toggle it off. Outfit entries also show Blizzard's outfit cooldown and Quickdraw's unusable tint during combat.

Credit to @sgtoj for the original idea and implementation work in #1851. This alternate implementation keeps that player-facing feature while using the direct 12.1 outfit lookup, explicit non-toggle action, cooldown feedback, combat state, and drag-and-drop support.

## How was it tested?

- Retail 12.1.0.69587.
- Created the Outfits preset with six saved outfits and verified names, icons, and Blizzard ordering.
- Changed outfits from the Quickdraw menu and confirmed the active outfit is not removed on repeat selection.
- Verified the cooldown swipe after changing outfits.
- Verified outfit entries become unusable during combat in a dungeon.
- Dragged a saved outfit directly onto an empty Quickdraw slot and activated it.
- Ran `/euidev` after the complete in-game test with no reported errors.
- Parsed both changed files as Lua 5.1, checked locale keys, ASCII-only additions, and `git diff --check`.

### Before

Quickdraw's preset list did not include saved outfits.

Quickdraw presets before adding Outfits

### After

The Outfits preset creates a menu from the character's saved outfits.

Quickdraw Outfits preset with six saved outfits

## Checklist

- [x] New settings default **OFF** (N/A: no setting is added; an outfit menu is created only when the player selects the preset or adds an outfit)
- [x] No writes onto Blizzard-owned frames (N/A: this uses Quickdraw's existing secure buttons and picker)

### DIFF
--- a/EllesmereUIOptions/EUI_Quickdraw_Options.lua
+++ b/EllesmereUIOptions/EUI_Quickdraw_Options.lua
@@ -1007,6 +1007,15 @@ initFrame:SetScript("OnEvent", function(self)
         return out
     end
 
+    local function OutfitEntries()
+        local out = {}
+        for _, slot in ipairs(ns.OutfitSlots and ns.OutfitSlots() or {}) do
+            local icon, name = ns.SlotDisplay(slot)
+            out[#out + 1] = { icon = icon, name = name, slot = slot }
+        end
+        return out
+    end
+
     -- Every palette this one may open. Not a list of things the game owns, so
     -- it is rebuilt on each use rather than cached: adding a palette or filling
     -- one in has to show up here without reopening the picker.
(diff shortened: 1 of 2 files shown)


sgtoj:
#1851 is effectively identical to this PR. Hopefully, one of ours will land.

EllesmereGaming:
merged ty both!

PR #1933, opened 2026-09-03
Feat: Add chat bubble customization

JuJuFX-dev:
## What does this PR do?

Adds a Chat Bubbles page to the Chat module that restyles the chat bubbles the game already draws over players' and NPCs' heads. Background colour and opacity, border size and colour, font size and colour, padding, maximum width and a vertical offset are all configurable, and the bubble text follows the Chat page's own font and outline pickers rather than only the global chat font. The page is off by default and asks for confirmation before it is switched on, because enabling it also switches Blizzard's own bubble settings on.

The text colour can either be a single colour of your own or Blizzard's, picked per channel: a cog beside the Font row's colour swatch carries a Follow Blizzard Default Color toggle, off by default. With it on, each bubble takes the colour the game itself gave that line, so a yell reads as a yell, and the swatch greys out to say so.

The Hide Chat Bubbles in Instances option allows users to disable Blizzard’s default chat bubbles while inside instances. When enabled, it disables the three relevant CVars: ChatBubbles, ChatBubblesParty, and ChatBubblesRaid.
After leaving the instance, the CVars are automatically restored to their previous values.

Rather than drawing its own bubbles on nameplates, EllesmereUI rides on Blizzard's. Their bubbles stay switched on, their chrome is hidden, and a styled frame is anchored to the frame that already carries their position. Every bubble therefore stays exactly where the game put it, including the one over the player's own head, and nameplates are neither involved nor required to be visible. Inside instances the engine's bubble frames are forbidden to addons and can be neither read nor hidden, so the feature stays out of the way there and Blizzard's own bubbles keep working as usual; the options page says so in a banner above the settings.

Channels are picked individually. Say, Yell, NPCs and Emotes share Blizzard's main bubble switch, while Party (which also covers instance chat) and Raid each have a switch of their own. Those switches are borrowed, never kept: the main one is held while at least one of its four channels is ticked, Party and Raid only while their own channel is ticked, and the Party and Raid ticks are seeded from the player's live settings the first time the feature is enabled, so no group bubbles turn up in a chat that had none. Every switch is put back the way it was found when a channel is unticked, when the feature is switched off, when entering an instance, and at logout, which is the only hand-back that still works if the addon is later disabled outright. Guild chat is not offered, because Blizzard draws no bubble for it and there is nothing to restyle.

## How was it tested?

Tested in game on the live client.

## Screenshots
grafik

Ingame Chat

## Checklist

- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations). Bubbles are picked up through `OnShow`/`OnHide` hooks on the engine's own frames and there is no `OnUpdate` anywhere. The single timer is a short bounded fallback that runs only while a chat line is still waiting for its bubble, at most eight ticks, coalesced into one chain no matter how much is being said, and it stops on its own.


Barbiero:
Yessss I'll use this a lot

since chat bubble customization only works outside of instances, I think it'd be cool if this module included a "automatically hide chat bubbles in instances"(essentially toggling the `chatBubbles`/`chatBubblesParty`/`chatBubblesRaid` cvar when entering an instance), since that is a behavior that is pretty useful imo - chat bubbles are a source of annoyance for me when fighting in m+/raids, but I like them in the overworld

JuJuFX-dev:
> Yessss I'll use this a lot
> 
> since chat bubble customization only works outside of instances, I think it'd be cool if this module included a "automatically hide chat bubbles in instances"(essentially toggling the `chatBubbles`/`chatBubblesParty`/`chatBubblesRaid` cvar when entering an instance), since that is a behavior that is pretty useful imo - chat bubbles are a source of annoyance for me when fighting in m+/raids, but I like them in the overworld

Thank you for the suggestion. I added that to the PR. :)

PR #1938, opened 2026-09-03
Fix Grow Direction not moving action bars

dfrisone:
```
Bug: The "Grow Right" setting doesn't move the Stance Bar (reported by @Flocki).

Issue: Picking a Grow Direction other than Centered writes the setting but the
bar never visibly reacts, on any horizontal or vertical bar that hasn't been
manually dragged -- not specific to the Stance Bar.

Root cause: Grow Direction only ever fed a multi-row stacking flag that a
single-row bar never exercises. The actual "grow toward an edge" mechanism in
this codebase is which frame edge stays pinned when the bar resizes, but
nothing ever re-derived that edge when Grow Direction changed outside of a
manual drag -- only the drag-save path used it.

Fix: Grow Direction changes now re-anchor the bar to the new edge immediately,
reusing the frame's current on-screen center so there's no jump; only future
resizes grow from the newly pinned edge. Also handles anchor-chained bars,
legacy corner-anchored positions, combat lockdown (deferred instead of
dropped), and a race where Unlock Mode's Save & Exit could commit a stale
position over the correct one if triggered mid-combat.
```

### DIFF
--- a/EUI_UnlockMode.lua
+++ b/EUI_UnlockMode.lua
@@ -7422,30 +7422,37 @@ local function CreateMover(barKey)
                     end
 
                     -- Restore the bar's visual center so it doesn't jump,
-                    -- then re-anchor based on the new growth direction.
-                    if barFrame and preCX and preCY then
-                        local postCX, postCY = barFrame:GetCenter()
-                        if postCX and postCY then
-                            local dx = preCX - postCX
-                            local dy = preCY - postCY
-                            if math.abs(dx) > 0.5 or math.abs(dy) > 0.5 then
-                                local pt, relTo, relPt, offX, offY = barFrame:GetPoint(1)
-                                if pt then
-                                    barFrame:ClearAllPoints()
-                                    barFrame:SetPoint(pt, relTo, relPt, offX + dx, offY + dy)
+                    -- then re-anchor based on the new growth direction. Skipped in
+                    -- combat: a protected bar's SetPoint is blocked there, so the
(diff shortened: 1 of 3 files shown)


dfrisone:
Closing: the substance of this landed on main, in a better shape than this branch had.

The inline block this PR added to unlock mode is now the helper `EllesmereUI._unlockSetGrowDirection(barKey, val)` (EUI_UnlockMode.lua), same structure and same comments, including the center-capture and the re-anchor:

```lua
-- Capture the bar's visual center before changing grow
...
local eab = EllesmereUI.Lite.GetAddon("EllesmereUIActionBars", true)
if eab and eab.SetGrowDirectionForBar then
    eab:SetGrowDirectionForBar(barKey, val)
end
...
-- Restore the bar's visual center so it doesn't jump,
-- then re-anchor based on the new growth direction.
```

`EAB:SetGrowDirectionForBar` is on main too. Rebasing this branch now drops three of its four commits as redundant and conflicts on the fourth purely because upstream extracted the same code into that helper.

**One gap looks unfixed, recorded here rather than lost.** The options panel's Grow Direction setter still routes elsewhere:

```lua
SSet("growDirection", v, function(k) EAB:ApplyIconRowOverrides(k) end)
```

so changing Grow Direction from the options panel, rather than from unlock mode's side picker, may still write the setting without re-anchoring the bar, which was the original report. That wants confirming in game first, and if it reproduces it is a one-line follow-up rather than this branch.

PR #1964, opened 2026-09-04
Quickdraw: drop the item range check that fires ADDON_ACTION_BLOCKED

### LINKED ISSUE #1963: Quickdraw: opening a palette with an item fires ADDON_ACTION_BLOCKED (C_Item.IsItemInRange)
## What happens

Opening a Quickdraw palette that holds an item logs a blocked action. Captured with `!BugGrabber` on 12.1, Quickdraw 9.1.6:

```
[ADDON_ACTION_BLOCKED] AddOn 'EllesmereUIQuickdraw' tried to call the protected function 'UNKNOWN()'.
[C]: ?
[C]: in function 'IsItemInRange'
EllesmereUIQuickdraw.lua:2210
EllesmereUIQuickdraw.lua:5167
EllesmereUIQuickdraw.lua:5404: in function 'Layout'
EllesmereUIQuickdraw.lua:6644: in function 'Open'
EllesmereUIQuickdraw.lua:8622
```

Line 2210 is the item range check in `SlotUsability`:

```lua
if C_Item.ItemHasRange(slot.id)
   and C_Item.IsItemInRange(slot.id, "target") == false then
    return "OUTOFRANGE"
end
```

## Why

The note above `SlotUsability` states the getters were checked rather than assumed, and it is correct about the predicate it names -- none of them carry `SecretWhenCooldownsRestricted`. Taint is a **different** predicate, and there the spell and item getters differ:

| Getter | Predicate |
|---|---|
| `C_Spell.IsSpellInRange` | `AllowedWhenTainted` |
| `C_Spell.IsSpellUsable` | `AllowedWhenTainted` |
| `C_Item.IsUsableItem` | `AllowedWhenUntainted` |
| `C_Item.ItemHasRange` | `AllowedWhenUntainted` |
| `C_Item.IsItemInRange` | `AllowedWhenUntainted` |

Calling an `AllowedWhenUntainted` function from addon code raises the blocked action.

## Impact

Beyond the error itself: **the feature it guards has never worked.** A blocked call cannot answer `false`, so no item could ever have received the `OUTOFRANGE` tint. The branch only ever cost an error.

Spell entries are unaffected, and item usability still works -- `C_Item.IsUsableItem` two lines below is reached in the same pass and produces no blocked action, despite carrying the same predicate. In practice only the call that takes a **unit token** appears to be blocked, though I cannot say from the outside why.

## Environment

WoW 12.1 retail, EllesmereUI Quickdraw 9.1.6, one occurrence per palette open with an item in it.

Sparxx947:
Fixes #1963.

## What does this PR do?

Removes the item range check in `SlotUsability` that fires `ADDON_ACTION_BLOCKED` every time a palette holding an item is opened, and corrects the predicate note above it.

The full capture, the predicate table and the reasoning are in #1963. The short version: the note above `SlotUsability` checked `SecretWhenCooldownsRestricted`, which is the secret-value predicate. Taint is a separate one, and there the getters differ -- `C_Spell.IsSpellInRange` and `C_Spell.IsSpellUsable` are `AllowedWhenTainted`, while `C_Item.IsItemInRange` is `AllowedWhenUntainted`.

**Nothing is lost.** A blocked call cannot answer `false`, so `OUTOFRANGE` for items could never have been returned -- the branch only ever cost an error. Spell range is untouched, and item usability still tints: `C_Item.IsUsableItem` sits two lines below, is reached in the same pass, and raises no blocked action of its own.

I corrected the note as well, since leaving it would invite the same mistake for the next getter added there.

### One thing I could not resolve

`C_Item.ItemHasRange` and `C_Item.IsUsableItem` carry the same `AllowedWhenUntainted` predicate, yet neither shows up as a blocked action -- `ItemHasRange` must have returned true for the `and` to reach line 2210 at all. So in practice only the call taking a **unit token** is blocked. I did not want to guess at that in a comment, and I deliberately left the other two alone: removing getters that are demonstrably still working would be a bigger change than the bug warrants. If you know the rule there, the note is easy to sharpen.

## How was it tested?

Honestly: **not verified in game.** The diagnosis rests on the captured error with its full stack plus the predicates in the generated API documentation, and the change only removes a call. Syntax checked with `luac5.1 -p`; the diff is ASCII-only, Lua 5.1, no version gates.

What a live test would show is the absence of an error, which the removal guarantees by construction -- the blocked call is simply gone. Happy to exercise a specific palette or item if you want that confirmed first.

## Screenshots

Not applicable. The removed branch could not produce a visible state: a blocked call never returns `false`, so no item ever received the `OUTOFRANGE` tint. Nothing that was on screen before looks different now.

## Checklist

- [x] New settings default **OFF** -- N/A, no setting added
- [x] Zero cost while disabled -- N/A, nothing added; this removes two calls from the paint path
- [x] Cheap while enabled: one getter fewer per item cell per paint
- [x] No writes onto Blizzard-owned frames -- N/A, untouched
- [ ] Tested in-game on live -- **no**, see above. No version gates or pre-Midnight APIs added.

### DIFF
--- a/EllesmereUIQuickdraw/EllesmereUIQuickdraw.lua
+++ b/EllesmereUIQuickdraw/EllesmereUIQuickdraw.lua
@@ -2170,14 +2170,20 @@ local USABILITY_TINT = {
 -- for a kind with nothing to say.
 --
 -- Every call here is secret-SAFE, and that was checked rather than assumed:
--- C_Spell.IsSpellUsable, C_Spell.IsSpellInRange, C_Item.IsUsableItem,
--- C_Item.ItemHasRange and C_Item.IsItemInRange all carry no
--- SecretWhenCooldownsRestricted flag in the generated documentation, unlike
+-- C_Spell.IsSpellUsable, C_Spell.IsSpellInRange and C_Item.IsUsableItem carry
+-- no SecretWhenCooldownsRestricted flag in the generated documentation, unlike
 -- the cooldown and charge getters two functions up. So these results may be
 -- branched on. Do not add a kind here without checking its getter the same
 -- way -- a mount's usability, for one, has to come from the Mount Journal
 -- rather than from its summon spell, which is not in the spellbook and
 -- answers unusable for every mount.
+--
+-- Secret-safe is only half of it, and the other half is a DIFFERENT predicate.
+-- The spell getters are AllowedWhenTainted and may be called from here.
+-- C_Item.IsItemInRange is AllowedWhenUntainted: calling it from addon code
+-- fires ADDON_ACTION_BLOCKED, and a blocked call cannot answer false, so the
+-- range tint it was meant to produce never appeared in the first place. Check
+-- BOTH predicates before adding a getter here.
 --
 -- Out of range OUTRANKS the other two, matching every action bar: a spell you
 -- cannot reach is the thing to say first, and it is the state a step forward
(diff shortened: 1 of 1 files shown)


EllesmereGaming:
fixed with next patch, went with a slightly more surgical approach. ty for report and fix!

PR #1967, opened 2026-09-04
Add(Aura Buff Reminders): flask reminder in combat and inside keystones

dfrisone:
## What does this PR do?

Requested by Kira: the flask reminder previously bailed out of every restricted context and reported "flask present", so it could never appear in combat or during a key.

Outside a key it now performs the check instead of bailing, so a missing flask reminds in combat. A new **Show Below In Combat** toggle (off by default) lets the existing Show Below thresholds apply in combat and in an active key, so a flask running out mid-fight reminds at the configured five minutes rather than only once it has gone.

Inside a keystone nothing about the player's own auras is readable, so the reminder is driven by the engine instead. An `AuraContainer` slot filtered to the flask spell IDs supplies a secret boolean, which is handed to `SetAlphaFromBoolean` with `alphaIfTrue` 0: the icon hides itself while the flask is up without this addon ever learning whether it is.

Two limits are inherent. It is binary in a key, because durations stay unreadable, so the threshold only applies outside one. And the icon holds its slot in the row while restricted whether or not it is visible.

## How was it tested?

Verified in a keystone: flask up and no icon, no flask and the icon shows, drinking a flask with the reminder up and watching it disappear. Verified in the open world that the existing behaviour and the threshold are unchanged.

## Screenshots

Behaviour change only; the icon is the existing flask reminder.

## Checklist

The presence container is built lazily on first use under restriction and only when the flask reminder is enabled, so a profile with it off builds no frames. The appear-sound is suppressed for engine-driven entries, since the addon cannot tell whether one is visible and would otherwise alert with the flask already up.

### DIFF
--- a/EllesmereUIAuraBuffReminders/EllesmereUIAuraBuffReminders.lua
+++ b/EllesmereUIAuraBuffReminders/EllesmereUIAuraBuffReminders.lua
@@ -1479,18 +1479,28 @@ local function PlayerHasWellFed()
     return false
 end
 
--- Unlike the other consumables this one still answers under restriction: the
--- Midnight flask buffs are whitelisted, so the player's own aura reads back
--- plainly in combat and inside a key. The unwhitelisted legacy TWW IDs and the
--- name scan are unreadable there and sit out.
+-- Unlike the other consumables this one still answers under restriction: a
+-- flask buff reads back plainly for the player whenever the client reports that
+-- aura as non-secret, which the static whitelist only approximates --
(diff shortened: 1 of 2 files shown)


dfrisone:
Closing in favour of a narrower PR. The keystone half of this does not work: inside a key nothing about the player's own auras is readable, and the engine-driven alpha route was probed to exhaustion without success. The in-combat work is tested and stands on its own, so it is being re-submitted separately. The branch is kept on the fork if the engine route is ever worth revisiting.

PR #2047, opened 2026-09-10
zhTW: translate 36 new keys for the Chat Bubbles page and Rotation Assist

LoChinAn:
Only `EllesmereUILocales/zhTW.lua` is touched: 37 lines added, 1 replaced. No module code, no `.toc`, no other locale.

The keys come from the two v9.1.8 additions that shipped untranslated: the **Chat Bubbles** page (24 keys - the enable gate and its confirm popup, the channel picker with its per-channel tooltips, the "Only works outside of Instances" banner, and the appearance sliders/swatches) and the Cooldown Manager's **Rotation Assist** style row (9 keys - style dropdown, colour mode, thickness and outset), plus 3 keys shared by both pages and the general page. Scanned incrementally from `0df54bb` to `271ffc3`.

Wording follows the client's own zhTW strings wherever one exists, so the options page reads like the rest of the UI: `Chat Bubbles` uses the client's `對話泡泡` (`CHAT_BUBBLES_TEXT`), and the channel labels use `說` / `大喊` (`SAY` / `YELL`). Tooltips reference options by the exact name the player sees on the same page rather than a literal translation of the English - the party channel tooltip says `小隊` because that is what `L["Party"]` already renders, and the Rotation Assist tooltip points at Blizzard's suggestion by the name the client uses for it, `輔助醒目標示` (`ASSISTED_COMBAT_HIGHLIGHT_LABEL`). `Solid Border` follows the existing `L["Solid"] = "實心"` used for border styles, and `Rotation Assist` follows the neighbouring row `Hide Rotation Helper` = `隱藏輸出循環輔助` so the two rows agree.

The one replaced line is the Chat module's `searchTerms` haystack. Upstream appended `bubbles bubble speech balloon nameplate` to it, and since `L()` is an exact table lookup the old key could never be hit again; the Chinese search words carry over unchanged with the bubble terms added, so nothing that used to be searchable stops being searchable.

Compiles under Lua 5.1, round-trips through the catalog parser, UTF-8 without BOM, CRLF like the sibling locale files.

### Checklist

- [x] N/A - no new settings
- [x] N/A - locale data only, no runtime cost
- [x] N/A - locale data only
- [x] N/A - no Blizzard frames touched
- [ ] Not yet loaded in-game on live; no version gates or APIs involved (locale data only)


LoChinAn:
Superseded by #2138, which carries all 36 keys from this PR unchanged and adds the zhTW strings v9.2/v9.2.1 introduced (Style page, WoW Forever popup, Mythic+ Run Summary, Swing Timer). Closing this one so there is a single locale branch to review -- the two would have conflicted in the same file. Thanks!

PR #2079, opened 2026-09-13
Fix: Re-strip events from disposed Blizzard action buttons on re-Show

svart2521:
Bug: https://discord.com/channels/585577383847788554/1545795450811056130

Issue: Blizzard's periodic re-init of a disposed MultiBarBottomLeft button (hidden via QuietlyHideBlizzButton) re-registers its default events onto an already-tainted button, so the next native cooldown update rejects secret start/duration values with "bad argument #1 to SetCooldown".
Fix: Mirrored the existing bar-level OnShow re-hide hook onto QuietlyHideBlizzButton itself, so a button's own re-init gets silently re-stripped of events too.


EllesmereGaming:
PR 2112 is being merged which includes this fix, ty!

PR #2080, opened 2026-09-13
Fix: NPC name text not repainting on hostility change

svart2521:
Bug: https://discord.com/channels/585577383847788554/1547242657347145728

When a unit frame's name/target text zone is set to class-colored, the text stays at its old color when the unit's hostility flips (friendly -> hostile etc). The health bar recolors correctly at the same moment.

Issue: The "text" channel in EUI_UnitFrames_Engine.lua's CHANNEL_EVENTS table (which drives PaintText, the painter that calls ApplyClassColor on name/target text zones via ns.ResolveUnitNameColor) never listed UNIT_FACTION, unlike the "health" channel. UNIT_FACTION was already treated as an always-deliver identity event (IDENTITY_EVENTS), so the engine was already built to route it to any channel that asked for it -- text just never asked.

Fix: Added UNIT_FACTION to the "text" channel's event list, so name/target text repaints on the same reaction-change edge the health bar already uses.


EllesmereGaming:
needs a little extra for the full fix so doing this one locally

PR #2081, opened 2026-09-14
Nameplates: apply the configured outline to friendly player names (fixes #1653)

### LINKED ISSUE #1653: Friendly player nameplates ignore outline setting (`flags or GetNPOutline()` short-circuits)
### Description

In name-only mode, friendly NPC nameplates show the outline configured in EllesmereUI settings, but friendly player (teammate) nameplates never do - their names always render without an outline, regardless of the outline setting.

### Cause

In `EllesmereUINameplates/EllesmereUINameplatesFriendly.lua`, the function `ApplyFriendlyFontOverride()` reads the stock font flags and falls back to the user setting:

```lua
local _, _, flags = SystemFont_NamePlate:GetFont()
SystemFont_NamePlate:SetFont(font, size, flags or GetNPOutline())
```

Blizzard's stock `SystemFont_NamePlate` has non-nil flags (`SLUG`), so `flags or GetNPOutline()` short-circuits and `GetNPOutline()` is never evaluated. The configured outline is silently dropped.

Friendly NPCs are unaffected because their overlay text (`ShowNPCOverlay`) calls `SetFont(font, size, GetNPOutline())` directly.

### Suggested fix

Use the setting directly:

```lua
SystemFont_NamePlate:SetFont(font, size, GetNPOutline())
SystemFont_NamePlate_Outlined:SetFont(font, size, GetNPOutline())
```

### Environment

- EllesmereUI 8.9.8 (reproduced on every version since at least 8.8.4)
- Client: zhCN retail (12.0)

Laxxzz:
## What does this PR do?

Friendly player names in Name Only mode now render with the outline configured
in EUI's font settings. Previously they always rendered without one, while the
friendly NPC overlays next to them honoured the setting.

`ApplyFriendlyFontOverride` read the flags back off `SystemFont_NamePlate` /
`SystemFont_NamePlate_Outlined` and only fell through to `GetNPOutline()` when
they were nil:

```lua
local _, _, flags = SystemFont_NamePlate:GetFont()
SystemFont_NamePlate:SetFont(font, size, flags or GetNPOutline())
```

Blizzard's stock font objects always carry non-nil flags, so the `or` never
reached the setting and the configured outline was silently dropped. The NPC
overlay path (`ShowNPCOverlay`) calls `GetNPOutline()` directly, which is why
NPC names were fine and player names were not.

Now stamps `GetNPOutline()` on both font objects directly, and folds the flags
into the existing file/size stamp so an outline change re-applies instead of
being skipped as unchanged. Same cost profile as before: one `SetFont` per
object, only when file, size or flags actually differ.

Fixes #1653

## How was it tested?

Live 12.1 client, build 9.1.8, Name Only mode with Show Friendly NPC Nameplates
on as a control group, Outline Mode set to Outline.

- Before: friendly player names rendered with no outline; NPC names beside them
  had one.
- After: player names carry the same outline as the NPC names.
- Outline Mode set to None and reloaded: player names drop the outline as well.

## Screenshots

Before:

image

After:
image

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) - N/A, no new setting; the outline setting already exists and this makes it apply where it was documented to
- [x] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built - unchanged; the function only runs in Name Only mode as before
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations) - one extra string compare in the existing stamp check
- [x] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames - N/A, touches the shared font objects the same way the existing code already did


EllesmereGaming:
went with a more surgical approach, live with next patch ty <3

Laxxzz:
> went with a more surgical approach, live with next patch ty <3

Sounds good buddy 🥰

PR #2088, opened 2026-09-14
Quality of Life: Added a 'Guild repair' reminder for Guild masters.

adbir:
## What does this PR do?

Add a Guild Repair reminder for guild masters when entering an instance with at least 50% guild members, or leaving a raiding instance. The displayed message can be setup the same way as the combat reminder, with an anchor in unlock mode.

## How was it tested?

Tested ingame as guild master

### Leaving instance:
billede

### Entering instance:
billede

### Menu item:
billede

### Tooltip:
billede

### Setup:
billede

### Anchor:
billede

## Checklist


EllesmereGaming:
while I very much appreciate the contribution, this feature is too niche to bring in. part of long term maintenance of EUI includes allowing some of the more niche features like that to be done via external addons. for a feature to be added to EUI it has to have a strong use case by a solid % of the userbase, especially if it adds hundreds of lines of code. Unfortunately as this is only used by GMs, and of that only a small % of GMs, this would be something that would fall under the 'too niche' category

PR #2089, opened 2026-09-14
Fix(Nameplates): apply outline setting to friendly player names (#1653)

herbsi012-glitch:
## Description
In name-only mode, friendly **player** nameplates never show the configured outline, while friendly NPC overlays do.

## Cause
`ApplyFriendlyFontOverride` read the stock font flags and fell back with `flags or GetNPOutline()`. Blizzard's stock `SystemFont_NamePlate` carries non-nil flags (`SLUG`), so the fallback short-circuits and the user setting is silently dropped.

## Fix
Apply `GetNPOutline()` directly to `SystemFont_NamePlate` / `SystemFont_NamePlate_Outlined`, exactly like the friendly NPC overlay already does (`ApplyOverlayStyle`).

## Compliance
- Lua 5.1 only, ASCII only (code, comments, strings)
- Bug fix only: restores intended behavior, no new feature surface, nothing defaults ON
- No new events, hooks, frames or OnUpdate paths added


EllesmereGaming:
already fixed via PR 2081, will be live with next patch thank you!!

PR #2095, opened 2026-09-15
Add modifier-gated buff tooltips

andybergon:
## What does this PR do?

Adds opt-in "Shown on Modifier" for buff tooltips, matching the existing debuff behavior. Shift, Control, and Alt share the existing debuff modifier setting; None requires no key. Buff tooltips remain Hidden by default.

Secure hover blockers now handle overlapping buff groups so holding the modifier reveals the aura beneath every overlapping blocker. This PR keeps the existing settings location.

This is a feature proposal despite the current feature-request pause. The settings reorganization is kept in a separate draft.

## How was it tested?

- Lua 5.1 compilation and focused mocked checks for visibility modes, shared keys, combat deferral, overlay retirement, and overlapping blockers.
- The overlap test fails on the previous implementation and passes with this change.
- The reporter confirmed buff tooltips appear with the modifier in the live client. Combat click-casting and 40-unit performance are not yet verified.
- Known limitations: full-frame bar blockers can intercept other aura hover areas while the modifier is released; bar anchoring can exceed the initial footprint clamp. These remain unresolved.

## Screenshots

Buff tooltip visibility options with Shown on Modifier selected

The screenshot shows the modifier option in the separate settings-layout build. This PR keeps the original Frames tooltip-cog location.

## Checklist

The last item is unchecked because only modifier visibility has live confirmation, not the complete combat/click-casting matrix.


andybergon:
Returning this PR to draft after finding a click-casting bug: with buff tooltips set to Alt, an Alt+left-click binding such as Power Infusion can cast on a different party member than the frame being clicked. Investigating the tooltip overlay and secure click routing; this will stay draft until fixed and verified in-game.

PR #2099, opened 2026-09-15
Fix: Targeted Spells "Where to Show" Bug

svart2521:
Bug: https://discord.com/channels/585577383847788554/1545057451743772732

Issue: Targeted Spell Bars and AuraBuffReminders sections set to only show in specific content types (e.g. Mythic/Non-Mythic Dungeons) still showed up in World Tier Lair boss fights. Both modules classify the current zone into a location bucket via a fixed allowlist of instanceType/difficultyID combos; anything unrecognized falls through to an "always show" fallback meant for genuinely unmappable content like arenas/PvP. Lairs are a brand-new instanced content type (four difficulty tiers: World, Normal, Heroic, Mythic) that didn't exist when these allowlists were written, so they fell into that same always-show fallback in both places, and neither options panel had a Lair entry to filter them with.

Fix: added a single "lair" location bucket to both modules, covering all four Lair difficulty tiers at once, detected via hasWorldTier (Blizzard's own GetInstanceInfo() flag for World-Tier-scaled content, added in patch 12.0.7) rather than guessing at specific difficulty IDs, checked ahead of the existing instanceType/difficultyID branches. Added a matching "Lair" checkbox to both Where to Show options lists.


EllesmereGaming:
needed a rebase but only a few lines of code so just added it locally will be live with new patch ty svart!

PR #2102, opened 2026-09-15
Fix duplicate Raider.IO tooltips on raid frames

KogasaPls:
## What does this PR do?

Raider.IO now matches group units by GUID upstream (RaiderIO/raiderio-addon#394), which resolves tooltip units directly. The manual `ShowProfile` call in EllesmereUI is no longer necessary and caused tooltips to render twice.

## How was it tested?

Verified syntax with `luac -p` and checked diff against upstream Raider.IO tooltip resolution.

## Checklist

- [x] New settings default **OFF** (N/A, no new settings)


EllesmereGaming:
Thanks Kogasa!

PR #2112, opened 2026-09-17
Stop secret cooldown errors from remaining action-bar broadcasts

dfrisone:
## What does this PR do?

Stops the remaining action-bar broadcaster events while cooldowns are secret. The existing full-to-press-and-hold fallback still permits ACTIONBAR_SLOT_CHANGED and PLAYER_ENTERING_WORLD to reach Blizzard's cooldown update path under addon taint, producing repeated secret-value SetCooldown errors. The guard now selects off for both active broadcaster modes.

This ports the runtime behavior of the user-confirmed actionbar-secret-cooldown test ZIP (original commit 0569cb39) onto current main. One file only; no new settings, APIs, timers, or polling.

## How was it tested?

- User confirmed the original ZIP fixed the reported cooldown error in game. Exact client build and test coverage were not recorded.
- Rebased onto `main` at v9.2.1 (41014eb2) on 2026-09-20. v9.2 touched this file, but not this guard, so the patch is unchanged at +5 -10 and still applies cleanly. `luac -p` passes on the full file and `git diff --check` is clean.
- Extracted action-bar source matches original commit 0569cb39 after line-ending normalization. Full file compiles under Lua 5.1; diff-scoped style gate and git diff --check pass.
- **Open risk, not fixed: Evoker empower keys.** Turning off the press-and-hold subset also removes native-button initialization events. Standard-bar empower keys still use native bindings; EUI's attribute reassertion updates its own buttons. Reloading or changing slots/talents in restricted content could leave native pressAndHoldAction stale. This is a source-review risk, not a reproduced client failure, and has not been fixed.
- Still outstanding before merge, and the reason this is up for review rather than ready to land: test Evoker keyboard hold-and-release with Press and Hold Casting off, after restricted-content reload, slot/talent changes, and queued casts; compare mouse casts. Recheck original errors plus extra-action and vehicle cooldowns. Re-enabling tainted cooldown dispatch or switching to the previously problematic queued click path is not a verified solution.
- Distribution ZIP was separately cleaned of repository metadata, with all retained files byte-identical. No ZIP or packaging changes are included in this source PR.

## Screenshots

N/A: no intended visual change. No new in-game captures supplied.

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) - N/A: existing error-path correction; no setting added.
- [x] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built - no new registrations, hooks or frames; changes an existing guard.
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations) - same existing predicate call; removes event dispatch under secrecy.
- [x] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames - existing broadcaster registration changes operate on Blizzard-owned frames; this patch extends when they are unregistered. No new custom frame fields or script replacement, but full taint/regression acceptance remains unverified.
- [x] Tested in-game on live; no version gates or pre-Midnight APIs added - original error fix confirmed by user; exact build/live status and Evoker regression coverage unverified. No new APIs or version gates; Midnight 12.1+ scope retained.


dfrisone:
Reopened. This was closed by mistake on 2026-09-19 while I was clearing out some fork-targeted PRs of mine; nothing was wrong with it and no review had happened.

Rebased onto `main` at v9.2.1 (41014eb2). v9.2 reworked parts of this file but left this guard alone, so the diff is unchanged at +5 -10 in one file. `luac -p` passes on the full file, `git diff --check` is clean.

Flagging the one thing worth your judgement before you spend time on it. The patch takes the broadcaster fully off under secrecy, which deletes the press-and-hold carve-out the current comment block defends:

> Under secrecy "full" drops to the press-and-hold subset when that need exists, else the frame goes bare: the cooldown ticks (~11/s at idle) are the flood, while SLOT_CHANGED and PEW are the only way Blizzard's twin buttons ever learn pressAndHoldAction, and a twin only raises there when its own cooldown is running.

So this trades a documented Evoker empower behaviour for the remaining secret-cooldown errors, and I have not reproduced or ruled out the empower regression in game. It is a source-review concern, not an observed failure. If you would rather keep that subset alive and narrow the fix to the cooldown event only, say so and I will redo it that way.

PR #2113, opened 2026-09-17
Fix: Quest Tracker Clickable When Hidden

svart2521:
Bug: https://discord.com/channels/585577383847788554/1549935368550481992

Issue: When the Quest Tracker is hidden (combat/raid visibility rules, or mouseover mode while idle), it still receives mouse clicks. Reported during a Mythic Twin Fangs pull: clicking to target an add where the (invisible) tracker sits instead opened the quest log and pinged the map, and the click never reached the world. Reproduces reliably against a target dummy in Silvermoon while just "Hide in Combat" is set. A live GetMouseFoci() capture confirmed the actual clickable element is an anonymous pooled button several levels inside CampaignQuestObjectiveTracker, not the tracker's own top-level frame.

Fix: ObjectiveTrackerFrame is EditMode-managed, so Hide()/Show() are combat-protected, and every hide path in this file (raid/arena auto-hide's combat fallback, the general user-visibility toggle, and mouseover idle) used SetAlpha(0) instead -- alpha never affects mouse interaction, and EnableMouse never cascades to child frames either, so a deeply nested quest-line button stayed fully clickable while the whole tracker looked invisible. Added SetTrackerHidden, which reparents the tracker under an already-Hidden container frame -- this drops visibility AND mouse for the entire subtree (children included) without walking any of them, matching this file's own stated design rule. SetParent is not part of the EditMode Hide()/Show() combat protection, so this now runs in combat too; it is pcall-guarded with a fallback to the old top-level-only SetAlpha/EnableMouse behavior in case that assumption ever proves wrong. Every hide/show site in the file (raid/arena auto-hide, the general visibility toggle, and the mouseover hover-reveal) now goes through this one helper.


EllesmereGaming:
had to fix this a bit differently, will be live with next patch!

PR #2116, opened 2026-09-17
Added Blackout Glow.

zanthor:
It's not really a glow, it's the opposite of a glow.

## What does this PR do?

Adds a glow stile to mask a CDM icon a solid color instead of a glow.  I found many of the glows very hard to distinguish and this cleared that up for me.

## How was it tested?

Retail 12.1 - tested in game.

## Screenshots
Before (Shape Glow used for example)
image

After:
image

## Checklist

- [X] New settings default **OFF** (no behavior change without opt-in)
- [X] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built
- [X] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations)
- [X] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames


zanthor:
Made more changes as I discovered other places I wanted to use this glow.  So closing this and opening a new because I'm not sure how to update it :).

PR #2126, opened 2026-09-19
Fix icon-only tabs rendering blank in the tab skinner

Thunderz96:
## What does this PR do?

The tab skinner in `EllesmereUIBlizzardSkin_WindowEngine.lua` clears every texture region on a tab before drawing the EUI look. That is right for text tabs, but icon-only tabs (the ones TabSystem makes with `AddIconTab`) keep their whole face in `tab.Icon` and `tab.IconMask`, so they end up as blank squares with nothing to tell them apart.

This skips those two regions in the clearing loop. Text tabs have neither field, so they are skinned exactly as before.

## How was it tested?

- Found in the WoW Forever beta (1.60.1.69913), where the spellbook's category tabs down the side are icon-only. Before: a column of empty dark squares. After: the class and general icons show, inside the EUI tab styling.
- The rest of the spellbook and the character frame tabs look unchanged.
- `luac -p` passes on the file.
- Not yet tested on live Retail, where I do not know of an icon tab that goes through this skinner. It is a two-line condition, so I expect no change there.

## Screenshots

I can add before/after shots from the beta if that helps, just say.

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) - N/A, no settings
- [x] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built - N/A, nothing added
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations) - two extra comparisons in a loop that runs once per tab
- [x] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames - this removes writes, adds none
- [ ] Tested in-game on live; no version gates or pre-Midnight APIs added - no version gates or old APIs, but tested on the Forever beta only, not live


EllesmereGaming:
superseded by another PR, thank you!

PR #2133, opened 2026-09-19
Fix friends list invites doing nothing

dfrisone:
## What does this PR do?

Inviting someone from the friends list did nothing. No error, no message, no invite. Reported on a client where `C_FriendList.IsLegacyFriendSystemEnabled()` returns **false** and `C_FriendList.GetNumFriends()` returns **0**, while the Battle.net list had 6 friends in it. Every friend on screen was unreachable.

Both invite branches went through `C_PartyInfo.InviteUnit(name)`, which routes through the legacy character-friend system. That API is annotated `RequiresValidInviteTarget = true`, and with the legacy system off a legacy name is not a valid target, so the call is dropped silently.

In 12.1 a Battle.net friend is invited by game account. `BNInviteFriend` only survives in `Blizzard_DeprecatedBattleNet` as a shim:

```lua
BNInviteFriend = function(gameAccountID)
	return C_BattleNet.InviteFriend(gameAccountID);
end
```

and that file returns early unless `loadDeprecationFallbacks` is set. So this calls `C_BattleNet.InviteFriend` directly with the `gameAccountID` the cached account info already carries. The name path stays as a fallback for an account with no game account id, and both remaining legacy calls now check the same flag Blizzard gates `FriendsFrame_GroupInvite` and `FriendsFrame_SendMessage` on, in `Blizzard_FriendsFrame/Mainline/FriendsFrame.lua`.

**This is not a Forever-specific fix.** Retail takes the identical wrong path and only works because the legacy system happens to be enabled there. No version gate, nothing client-specific.

Whispers are unaffected: this module has no whisper implementation of its own, deliberately, per the taint notes at the top of the file.

## How was it tested?

**Not tested in game.** `luac -p` passes, house style linter reports 0 errors on the diff, no new locale keys.

Worth knowing for whoever tests it: the Battle.net branch requires `gameAccountInfo.isOnline`, so it needs a friend who is actually online, not merely in the list.

1. Battle.net friend online, click to invite from the list. Invite goes out.
2. Same on a client with the legacy friend system enabled (retail). Still works, now through the game account rather than the name.
3. A legacy character friend, where that system is enabled. Still invites by name.
4. Battle.net friend offline. Nothing happens, unchanged.
5. Someone whose privacy settings block party invites. Expected to fail, see below.

**Known gap, deliberate:** Blizzard gates whether to even offer the action on `FriendsListUtil.GetBattleNetFriendPartyInviteRestriction(friendIndex) == BattleNetFriendPartyInviteRestrictionType.None`. This PR does not, because the restriction helper wants a friend index that this click path does not carry, and the fix for a dead invite should not drag a UI-gating change along with it. The consequence is that clicking a restricted friend fails quietly rather than the row being greyed out, which is no worse than today. Happy to do it as a follow-up if you want parity.

## Screenshots

N/A, no visual change.

## Checklist

- [x] New settings default **OFF** (no behavior change without opt-in) - N/A, no new settings; this is a bug fix
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations) - one API call per click
- [ ] Tested in-game on live; no version gates or pre-Midnight APIs added - no version gates, no pre-Midnight APIs, but NOT yet tested in game

otter


dfrisone:
Closing for now: the fix does not resolve the reported behaviour in testing. Reopening once the actual invite path is understood.

PR #2143, opened 2026-09-20
Forever: decide snippet support by running one, not by a scrubbed global

dfrisone:
`SecureSnippetsOK` decided Forever's snippet support by looking at a global that is nil on every healthy client, so the gate it feeds could never lift.

## Cause

```lua
v = (type(_G.loadstring_untainted) == "function")
```

`Blizzard_EnvironmentCleanup` is `## LoadFirst: 1` and nils `loadstring_untainted`, alongside `secretunwrap` and the `SecureMixin` family, after `Blizzard_RestrictedAddOnEnvironment` has captured its own file local. Addons are not meant to reach it. So the test reads nil on a working client too, and the comment's promise that support "comes back on its own the day the client is fixed" could not have held. Every `SNIPPETS_OK` feature would have stayed stood down on Forever permanently, even after a client fix.

## Change

Probe the outcome instead: build a `SecureHandlerAttributeTemplate` frame, `Execute` a one-line body, read the attribute back.

Forever still answers false, now for a reason that can change. Retail never probes and never creates the frame.

The compile raises inside Blizzard's own script handler, so the verdict comes from the attribute read rather than the pcall, and a broken client gets one grabber line per session however the error handler is parked. That is the price of a gate that can lift itself.

## Combat

Skipped in combat, where the SecureHandlers API raises outright. It stands down for that session, because the load-time readers latch whatever this returns, so a reload mid-fight leaves the snippet modules off until the next one.

That matches the behaviour being replaced, which answered false on Forever in or out of combat. It does become visible the day the client is fixed: a combat reload would cost RaidFrames and Quickdraw for that session. Persisting the answer per client build would remove it, and I am happy to add that if you would rather not carry the edge.

## Related

This is what [#2142](https://github.com/EllesmereGaming/EllesmereUI/pull/2142) keys off. Independent commits, either order.

Blizzard bug report filed for the underlying client fault: `RestrictedExecution.lua` captures a nil `loadstring_untainted` on Forever, which breaks every secure handler snippet client-wide, for every addon, not just this one.

## Testing

Verified on Forever: probe returns false, suite loads clean, one expected `RestrictedExecution.lua:79` line in BugSack at login. Verified on retail: the same snippet returns 42, and the probe is never reached.

cat


dfrisone:
Parking this as a draft. Not abandoned, and nothing here is wrong.

The reason is timing rather than correctness. Old code and new code both return `false` on Forever and `true` on retail, so nothing observable changes the day this ships. What it fixes is deferred: `SecureSnippetsOK` currently tests a global that `Blizzard_EnvironmentCleanup` nils on every healthy client, so it can never return true, and everything behind `ns.SNIPPETS_OK` would stay stood down even after Blizzard fixes the client. This makes that recover by itself.

The cost is immediate. The probe raises inside Blizzard's handler, and grabbers log it whatever the error handler is parked to, so every Forever user with BugSack gets an EllesmereUI-attributed error at every login from the moment this merges. Benefit arrives only when Blizzard ships a fix.

Persisting the verdict would have removed that, and I checked: it cannot work here. `EllesmereUI.FOREVER_SV_BUG` is true on Forever and a `PLAYER_LOGOUT` handler nils every SavedVariables store the suite declares, `EllesmereUIDB` included, so nothing reaches disk on that client by design. Persistence becomes possible the day that switch flips to false, and the combat-reload edge noted in review goes with it.

Blizzard bug report is filed for the underlying fault: `RestrictedExecution.lua` captures a nil `loadstring_untainted` on Forever, breaking every secure handler snippet client-wide for every addon. Retail returns 42 to the same probe.

Worth picking this back up when Blizzard responds, as part of taking the client fix.

Thunderz96:
This week's beta update, build 1.60.1.70009, fixed the client side of this.

`Blizzard_EnvironmentCleanup.toc` changed from

    ## Dep: Blizzard_RestrictedAddOnEnvironment [AllowLoadGameType classic, standard]

to

    ## OptionalDep: Blizzard_RestrictedAddOnEnvironment

so on Forever the restricted environment now loads, and takes its local copy of `loadstring_untainted`, before the cleanup nils the global. (Diff in Gethe/wow-ui-source, `forever` branch, bd2470a.)

Tested on 70009, out of combat:

```
/run local f=CreateFrame("Frame",nil,nil,"SecureHandlerBaseTemplate") f:Execute("self:SetAttribute('fbok',42)") print("snippet",f:GetAttribute("fbok"))
```

prints `snippet 42`, with nothing in BugSack. `type(loadstring_untainted)` still prints `nil`, exactly as you described.

So as of 70009 the current `SecureSnippetsOK` test looks like the only thing keeping the SNIPPETS_OK features stood down on Forever, which sounds like the day this PR was waiting for. I have only checked that a snippet runs, not every gated module.

PR #2147, opened 2026-09-20
borders: keep textured borders and bar clips on the pixel grid

thoriphes:
> Bug fix with no new setting, so it fits the current "bug fixes only" window in CONTRIBUTING.md
> and satisfies criterion 2 ("only genuine bug fixes may change behavior without opt-in").

## In short

When the UI is scaled so that a unit is not a whole number of pixels, the offsets and insets
that place a border texture and the bar under it come out with decimal pixel sizes, and the
GPU rounds those edges on its own - up on one side, down on the opposite one - so the border
reads a pixel thicker on two sides. The fix does the rounding itself, once, before placing
anything, so every side lands on the same whole pixel and the border's size and distance are
uniform all the way round.

## What does this PR do?

Textured borders (Shadow, Glow, Blizzard, any SharedMedia border) render one pixel thicker on
two sides than on the other two at most UI scales. Two rects were landing on pixel centres;
both are now kept off them. No option, no visible change where the rects already sat on
whole pixels.

### Units and pixels
EUI works in UIParent units. One physical pixel is `PP.mult = PP.perfect / uiScale` units,
with `PP.perfect = 768 / screenHeight`; the inverse, pixels per unit, is
`screenHeight / 768 * uiScale`. Whole pixels per unit happen only when
`uiScale = 768 / screenHeight * N` for an integer N; everything else is fractional.
Reference numbers used below: 2160p, `ppUIScale = 0.6222` -> 1 unit = 2160/768 x 0.6222 =
**1.75 px**.

### How a rect on a half pixel is drawn
A quad edge at pixel coordinate x.0 covers columns cleanly. An edge at x.5 sits exactly on
the centre of column x, and the rasteriser resolves the tie with its fill rule: the boundary
pixel is covered on one side of the quad and not on the other (top-left rule: left/top edges
own their centre pixel, right/bottom edges do not). So a rect whose four edges all sit on
pixel centres covers **one column and one row more on one side than on the other**. That is
the whole bug, twice.

### 1. The border backdrop (`ApplyBorderStyle`)
The BackdropTemplate frame was anchored at raw units:

```lua
bdFrame:SetPoint("TOPLEFT",     borderFrame, "TOPLEFT",     -offsetX + sx,  offsetY + sy)
bdFrame:SetPoint("BOTTOMRIGHT", borderFrame, "BOTTOMRIGHT",  offsetX + sx, -offsetY + sy)
```

For a plain texture `offsetX/Y` is the user's Border Offset (units); for a `scaleOffset`
texture (Glow, Shadow, Blizzard, Lightspark) it is `edgeSize/2 + adj`, with `edgeSize` =
12/16/24/32 units for size 1-4. The owner (`borderFrame`) is on the grid (`PP.Point` /
`PP.Size`); the backdrop is owner +/- offset, so its edges are on the grid only when `offset`
is a whole number of pixels:

| case | offset (units) | x 1.75 px | edges |
|---|---|---|---|
| Raven Rounded, Border Offset 2 | 2 | **3.5 px** | half pixel |
| Shadow, size 1, Offset 0 | 12/2 + 0 = 6 | **10.5 px** | half pixel |
| Shadow, size 2, Offset 0 | 16/2 = 8 | 14 px | on grid |
| Border Offset 4 | 4 | 7 px | on grid |

Measured on the player frame (physical px; frame `L 1304 R 1620 T 868 B 826`, Offset 2):
backdrop `L 1300.50 R 1623.50 T 871.50 B 822.50`. Visible border: 4 px on two sides, 3 px on
the other two (which pair depends on the driver's tie convention; the point is they differ).

Fix: snap the offset and the shift **once each, as magnitudes**, to whole pixels at the
backdrop's own effective scale, then mirror them to the four anchors:

```lua
local sok, ses = pcall(bdFrame.GetEffectiveScale, bdFrame)
if not (sok and ses and ses > 0.01) then ses = UIParent and UIParent:GetEffectiveScale() or 1 end
offsetX, offsetY = PP.SnapForES(offsetX, ses), PP.SnapForES(offsetY, ses)
sx, sy           = PP.SnapForES(sx, ses),      PP.SnapForES(sy, ses)
bdFrame:SetPoint("TOPLEFT",     borderFrame, "TOPLEFT",     -offsetX + sx,  offsetY + sy)
bdFrame:SetPoint("BOTTOMRIGHT", borderFrame, "BOTTOMRIGHT",  offsetX + sx, -offsetY + sy)
```

`PP.SnapForES(x, es)` = `round(x / onePixel) * onePixel` with `onePixel = PP.perfect / es`
(round half up, epsilon-guarded; the helper the PP border system already uses). 3.5 px -> 4 px,
10.5 px -> 11 px, 7 px -> 7 px. Snapping the **magnitude once** matters: snapping each corner's
signed offset separately rounds -3.5 to -3 and +3.5 to +4 (round half up) and just moves the
asymmetry to the other pair of sides - that was the first cut, measured as
`L 1301 R 1624 T 872 B 823`. Final: `L 1300 R 1624 T 872 B 822`, 4 px on every side.

`ses` is the backdrop's effective scale, so the snap is right for `normalizeScale` callers and
for frames with a scale chain below UIParent; the `pcall` mirrors the ratio probe just above
it (forbidden-anchor owners). The secret-safe eight-piece path gets the same snap on its four
corner anchors, which carry the outer edges there.

Why `PP.SnapForES` and not one of the other PP helpers: `PP.Snap` / `PP.SetOutside` work at
UIParent scale only (`PP.mult`), which is wrong for `normalizeScale` callers and for any
owner with a scale chain below UIParent (nameplates), and `PP.SetOutside` goes through
`PP.Scale`, which truncates (3.5 px -> 3 px, thinning the border instead of holding it) and
calls `DisablePixelSnap` on a BackdropTemplate frame. `SnapForES` is the helper the PP border
system already uses for exactly this: N whole pixels at a given effective scale.

### 2. The fill clips
With Show Behind on the border is under the bar, so the bar's own edges decide how much of
it shows - and the bars were on half pixels by construction. Unit frames inset `_barClip` by
`(PP.perfect / es) * 0.5` on every side "so the GPU clips any StatusBar texture rounding past
the frame edge"; resource bars inset their inner StatusBar by `PP.mult * 0.5` for the same
reason. Measured health bar: `L 1304.50 R 1618.50 T 872.50 B 825.50` inside a frame at
`1304 / 1619 / 873 / 826` - every edge on a pixel centre, so the bar covers column 1304 but
not column 1618: visible border 4 px on the left and top, 5 px on the right and bottom.
Border snapped or not, this alone makes Show Behind uneven.

Any positive inset gives the bleed guarantee (the clip is strictly inside the frame either
way); half a pixel is the one value that lands on the centres. Both insets are now
**0.25 px**: edges at x.25 / x.75 sit between a centre and a boundary, so the pixel whose
centre is inside is covered on both sides and the bar covers exactly the frame's pixel rect:

```
left  edge 1304.25 -> column 1304 (centre 1304.5) covered
right edge 1618.75 -> column 1618 (centre 1618.5) covered
```

Bar width in pixels is unchanged (315 -> 315); only where the sub-pixel margin lives moved.

## Acceptance criteria
- **No setting, so nothing to be off** - this is a bug fix (criterion 2's "genuine bug fix"
  case); there is no toggle, no event, no hook, no frame added. Cost per textured-border
  apply: one `pcall` scale probe and four `SnapForES` calls, pure arithmetic, no table
  allocation. In the eight-piece path that is per textured aura border per refresh, same
  order of cost as the anchor calls already there. The clip change is a constant.
- **Zero behaviour change on the grid** - `SnapForES` returns its input when it is already
  whole pixels (any 1 px/unit or 2 px/unit scale); the clip inset moves a quarter pixel
  inside the frame, never visible on its own.
- **Zero taint** - addon-owned frames only (`bdFrame` is the addon's BackdropTemplate child,
  the eight edge textures are addon-created), same widget calls as before, no `SetScript`,
  no field written onto a Blizzard frame.
- Lua 5.1, ASCII, `luac -p` clean on all three files.

## How was it tested?
Live 12.1 client, 2160p, `ppUIScale` 0.6222 (1.75 px/unit). Rects read with a local helper
(`/nui uf <unit>`: frame / clip / health / border / backdrop in physical px). Unit frames with
Raven Rounded (LSM, 128x16 strip) at Offset 2 and built-in Shadow at size 1 Offset 0, Show
Behind on and off; resource bars for the clip change.

- Backdrop before -> after: `L 1300.50 R 1623.50 T 871.50 B 822.50` -> `L 1300 R 1624 T 872 B 822`.
- Health before -> after: `L 1304.50 R 1618.50 T 872.50 B 825.50` -> quarter-pixel edges; the
  border under it equal on all four sides.
- Offset 4 (already whole pixels): numbers unchanged.
- Solid (PP) borders: different code path, untouched.

Not yet covered: CDM icon borders (`normalizeScale`) and nameplate borders (scale chain below
UIParent) - the snap uses the backdrop's effective scale so they should land the same way,
but I have not read their rects. Raid frames have their own clip; not audited.

## Screenshots
Same frame, same settings, only the branch differs.

| before | after |
|---|---|
| before, textured border | after, textured border |
| before, shadow border | after, shadow border |

## Checklist
- [x] New settings default **OFF** (no behavior change without opt-in) - N/A: no setting is
  added. The rendering change applies to everyone because it is a genuine bug fix (borders
  drew unevenly); where the rects were already on whole pixels nothing moves.
- [x] Zero cost while disabled: no events registered, no polling, no hooks doing work, no
  frames built - N/A (nothing to disable); no event, hook, frame or timer added.
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame
  allocations) - one `pcall` and four `SnapForES` calls per textured-border apply, no
  allocation; the clip inset is a constant.
- [x] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript` /
  `hooksecurefunc` only, never `SetScript` on Blizzard frames - only addon-owned frames and
  textures are touched, no new scripts or hooks.
- [x] Tested in-game on live; no version gates or pre-Midnight APIs added - live 12.1 client,
  see "How was it tested?"; no gates, no new API calls (`GetEffectiveScale`, `SetPoint` only).


EllesmereGaming:
need this to be rebased and change PR 2135's line to + clipInset

thoriphes:
Rebased onto v9.2.2 and changed #2135's line to `+ clipInset`. Since the clip inset is a quarter pixel here rather than a half, I also reworded the tail of that comment from "move centered text down half a pixel" to "down by the inset". Nothing else changed.

PR #2149, opened 2026-09-20
unlock: edit an anchor's points and offsets from the cog menu

thoriphes:
> Per CONTRIBUTING.md feature PRs are on hold until a few weeks after 12.1 launch and larger
> features should be pitched on Discord first. The pitch went to @ellesmere by Discord DM; this is the write-up.

## What does this PR do?

In Unlock Mode, the right-click cog menu of an **anchored** element gets an Anchor block under
the Width / Height / X Position / Y Position rows:

```
Anchored to: Action Bar 2
Action Bar 3 point       [ Bottom Left  v ]
Action Bar 2 point       [ Bottom Right v ]
Offset X                 [ 0 ]
Offset Y                 [ 0 ]
```

- Two dropdowns, one per element, each labelled with that element's name and offering the nine
  frame points (Top Left, Top, Top Right, Left, Center, Right, Bottom Left, Bottom, Bottom Right).
- Picking a point **keeps the element where it is**: the offset is re-derived from live bounds
  for the new pair. Typing an offset **moves it** (Enter or focus loss commits, Escape reverts).
  The boxes are in **physical pixels**, like the X/Y Position boxes and the pixel sliders
  (`PP.ToPixels` / `PP.FromPixels`); the record keeps units.
- The X / Y Position boxes are disabled on anchored elements the way a matched Width / Height
  box is (greyed, tooltip "Anchored to X. Edit the anchor offsets below."); they keep reporting
  the position live. Arrow-key nudges and offset commits keep every box in lockstep.
- The dropdown list opens upward when it would run past the bottom of the screen, and closes
  when the mouse leaves it, like the Relative to Screen flyout.

Unanchored elements see no change. Existing anchors behave exactly as before until someone
touches the new rows.

## Why

Anchoring in Unlock Mode is drag-and-snap: pick a target, pick a side, done. That is right for
"put this next to that", but the result is not inspectable or precisely editable: the record
(`target`, `side`, `offsetX`, `offsetY`) never shows up anywhere, and the only way to adjust it
is to nudge with the arrow keys and read the X/Y Position boxes, which report *screen*
coordinates, not the relation.

Two things were out of reach without this:

1. **Exact relations.** "Bar 3's bottom-left corner on Bar 2's bottom-right corner, 10 px apart"
   is the kind of thing a layout is made of, and it was a nudge-and-squint job.
2. **Anything but centred.** The side model centres the child on the target along the axis
   perpendicular to the side. "Sit to the right of this bar with the bottoms aligned" needed a
   hand-computed `offsetY = (childH - targetH) / 2`, redone whenever either bar changed size.

`SetPoint(point, relativeTo, relativePoint, x, y)` is the vocabulary every WoW UI user already
has for exactly this. The menu now speaks it.

### Storage
Two optional fields on the anchor record: `ai.point` (this element) and `ai.relPoint` (target),
any of the nine frame points. `ai.side` is kept in sync from the pair (`_SideFromPoints`) for
the parts of the system that read it - chain extents, mover captions, fallback picks, the
CDM/action-bar growth compensation (whose own "did the centre sit where the side geometry says"
gate keeps it off corner pairs). A record without points reads back as the classic side geometry
(`_PointsFromSide`: child's near edge on the target's edge, centred across), so the dropdowns
show the truth for old anchors too.

### Placement (`ApplyAnchorPosition`)
When both points are present the child centre is computed directly:
`centre = targetPoint + offset - half * size` (`_AnchorPointXY`, `_AnchorPointHalf`, defined
at the top of the file so the pre-load `ReapplyOwnAnchor` stub places a pointed record the same
way). This replaces the side/centre geometry outright and **skips the growth-edge pin and edge
preservation**: the own point *is* the fixed edge, so a resize grows away from it by
construction, which is what the pin exists to emulate for side anchors. Records without points
take the unchanged classic path.

`PropagateAnchorChain`'s axis gate ("a TOP/BOTTOM child is untouched by the parent's width
change") assumes the cross axis is tied to the target's centre; with points it is tied to the
target point, so the gate reads that point instead and a child on a corner cascades.

### Capture
- Point pick: `EllesmereUI.SetAnchorParams(key, { point= | relPoint= })` stores the pair and
  re-derives the offset through `_CaptureAnchorOffsets` (which grew an optional point pair) so
  nothing moves and the raid container's per-tier extra offset stays out of the record, the
  same way a drag ends; drops the growth pin (`refFor`, `refX/Y`, `edgeOffX/Y`) so a later
  side-path apply would recapture movement-free; bumps the anchor-links stamp when the derived
  side changes; re-applies, propagates the chain synchronously (the nudge path's way), syncs the
  mover, marks the session dirty.
- Typed offset: `SetAnchorParams(key, { offsetX= | offsetY= })` writes and re-applies.
- Drag drop: the post-drag capture stores the offset point-to-point when points are present;
  the screen-anchor set/clear recaptures pass the pair to `_CaptureAnchorOffsets` too.
- Nudges: unchanged - they add to `offsetX/Y`, which is the right thing in both models.
- Units: `offsetX/Y` are UIParent units in the record (what `ApplyAnchorPosition` adds to
  the target point); the boxes convert to and from physical pixels so +1 is one pixel, one
  arrow nudge, and the numbers agree with the pixel sliders. At a non-pixel-perfect scale the
  two differ by the scale factor (1.75 px/unit here), which is how "Offset 5" looked equal to
  "Spacing 10" in the first build.
- Discard: the Unlock Mode snapshot taken on enter and restored on "Exit without saving"
  copies anchor records field by field; `point` / `relPoint` ride along with the rest, so a
  pointed anchor reverts to what it was, not to a side record carrying point offsets.

### Menu
Rows follow the existing `MakePosRow` shape (font, colours, input sizes); labels go through
`EllesmereUI.Lf`, the value button's arrow is `MakeDropdownArrow`. The point list is one frame
per mover, kept on the mover and parented to `unlockFrame` like the Relative to Screen flyout
(`BuildCogMenu` reparents every `cogMenu` child to nil on each open), rebuilt on each show with
that flyout's colours, font, orange current row and leave-to-close, clamped to screen and
flipped upward when there is no room below; picking rebuilds the cog menu in place so the new
pair shows. No new locals in the deferred body (at the 200-local cap); helpers hang off
`EllesmereUI` with the underscore the file uses for its private ones.

## Acceptance criteria (CONTRIBUTING.md)
- **Zero cost unless enabled** - no points stored: one nil check each in `ApplyAnchorPosition`,
  the post-drag capture, `_CaptureAnchorOffsets`, the pre-load reapply stub, and one per child
  in `PropagateAnchorChain`'s axis gate; the menu block is built only for anchored elements and
  only when the cog opens.
- **Zero behavior change without opt-in** - `point` / `relPoint` are nil by default; the classic
  placement, pin, edge preservation, drag and nudge paths are untouched for such records. The
  only visible change for existing anchors is the extra menu block (read-only X/Y on them being
  the one deliberate change, see notes).
- **Low cost when enabled** - pure arithmetic on the same live bounds the classic path reads;
  no events, no OnUpdate, no timers as logic (the list's leave-to-close uses the same 0.05 s
  hover check as the Relative to Screen flyout), no allocations in the apply.
- **Zero taint risk** - only addon-owned frames (movers, the cog menu, the list) and the same
  `ClearAllPoints`/`SetPoint` on the element that `ApplyAnchorPosition` already performs, under
  the same combat guard (`SetAnchorParams` refuses in lockdown).
- **Midnight only** - no version gates.
- Lua 5.1, ASCII only, `luac -p` clean.

## How was it tested?
Live 12.1 client, one character, one evening - not days of play. Everything below was seen on
the first three commits; the review-fix commit (`0005fbff`: capture through
`_CaptureAnchorOffsets`, the chain gate on the target point, the pre-load stub, synchronous
cascade, the list on the mover with the flyout's styling and `MakeDropdownArrow`, the disabled
X/Y boxes with tooltip, in-place menu rebuild) is `luac -p` clean but has NOT been run in game
yet. Exercised:
- Action Bar 3 (3x4 grid) anchored to Action Bar 2: Bottom Left -> Bottom Right, 0/0. Probe
  (`/nui anchor Bar3`, a local helper) after apply: bar left 1417.2 = target right 1417.2,
  bottoms 21.1 = 21.1. Save & Exit, `/reload`: same numbers.
- Every other pair on the same bars (corners, edges, centre), offsets positive and negative,
  Enter and focus-loss commits, Escape revert.
- Arrow nudges with the cog open: X/Y and Offset boxes track; offset commit refreshes X/Y.
- Dragging the anchored bar: offsets re-captured point to point, chain still propagates.
- Dropdown at the bottom of the screen opens upward; at the top opens downward.
- Existing anchors (Bar2 -> MainBar, XP -> MainBar, pet/stance bars, unit-frame cast bars)
  untouched: dropdowns show the classic pair, positions unchanged after `/reload`.
- Exit without saving after editing points: every pointed anchor back where it was. (First
  build lost `point`/`relPoint` in the discard snapshot and all pointed elements jumped -
  found in use, fixed in 08e20d7c.)
- Typing 10 + Enter shows 10 and stays 10 across nudges (first cut of the pixel boxes had
  the nudge-sync writing units back into the box, and the Enter -> focus-lost double commit
  then re-read them as pixels: 10 became 3 in one keystroke).
- Offset boxes read back the pixel sliders' numbers: Button Spacing 10 and Offset 10 give
  the same gap between two bars as between two buttons; a typed 10 stores 5.71 units at
  1.75 px/unit and the anchor probe confirms a 10 px gap.
- No BugSack entries.

Not yet covered: CDM bars with custom growth directions (the pin-skip path matters most there),
a pointed child on a target corner when the target resizes (the chain-gate change), a raid
container with a per-tier offset (the `_CaptureAnchorOffsets` change), login with a pointed
CDM bar (the stub change), vehicle / combat lockdown transitions with the menu open, spec
overrides on an anchored element, the fallback-anchor and override-anchor variants. Will update
before this leaves draft.

## Screenshots
Taken on `993b3c86`, before the review-fix commit (the list
now wears the Relative to Screen flyout's styling and the value button the shared dropdown
arrow; the block, the fields and the flow are unchanged).
1. `1-cog-menu-anchor-block` - Pet frame cog menu: the Anchor block under the size/position
   rows (Anchored to: Player, `Pet point` Bottom, `Player point` Top, Offset 1 / 20), X and Y
   Position greyed out.

   cog menu anchor block
2. `2-own-point-list` - the nine-point list open on the element's own point.

   own point list
3. `3-target-point-list` - the same list on the target's point (labelled with the target's name).

   target point list
4. `4-list-flips-upward` - Action Bar 3 in the bottom-right corner: the list opens upward
   because there is no room below; the pair reads Bottom Left -> Bottom Right, Offset 5 / 0.

   list flips upward

## Design notes / things I went back and forth on

0. **The interactive flow is deliberately untouched.** Anchor To, the side picker, drag-and-snap,
   arrow nudges, chains, fallback and override anchors all work exactly as they did; this PR
   only adds a way to *read and edit* the resulting relation once it exists. It would have been
   tempting to make the picker itself point-based (nine-by-nine instead of a side), but that
   turns a one-click gesture into a form, and the side gesture is the right default for the
   common case. Precision is the exception, so it lives in the cog menu, opt-in, after the fact.
   Keeping the two apart also keeps the diff reviewable: nothing in the existing paths changed
   meaning, they just learned to read a record that may carry two extra fields.
1. **Side + Align first.** The first cut exposed the existing `side` plus a new cross-axis
   `align` (Top/Bottom for Left/Right sides, Left/Right for Top/Bottom). Three problems in the
   first hour of use: "Side" reads as *my* side but is the target's edge; the Anchor To picker's
   grow-direction remap ran on a side change and the growth pin then held the growth axis where
   it was, so the alignment silently did not apply; and the pair still could not express
   "my bottom-left on your bottom-right" without an offset someone has to compute. Replaced by
   explicit points on both elements, labelled with the element names.
2. **Points replace the pin, they do not fight it.** With an explicit own point the fixed edge
   is known, so the growth-edge pin (whose whole job is to keep a growth bar's fixed edge still
   while its size changes) has nothing left to protect. Pointed records skip it; side records
   keep it. That is also why picking a point clears the pin fields: a later side-path apply
   recaptures movement-free rather than trusting stale edge offsets.
3. **Keep-position on pick vs. move.** A point change re-derives the offset so the element does
   not jump; only a typed offset moves it. The alternative (jump to the new pair at the old
   offset) makes the dropdowns a way to fling elements across the screen, which is what the
   drag is for.
4. **Read-only X/Y on anchored elements.** Before, editing X/Y Position on an anchored element
   nudged the offsets underneath - two controls for one value, and the Offset boxes did not
   refresh, so the menu could show a number that was not true. Now the anchor owns the position:
   X/Y report, Offset edits. This is the one deliberate behaviour change for existing anchors
   and I would rather have it flagged than slipped in.
5. **Nine points, not corners only.** Edge midpoints (Top, Left, ...) cost nothing and are the
   natural way to say "centred on that edge" (a side anchor with a 0 cross offset).
6. **The half-size factor.** `_AnchorPointHalf` returns the point's offset from the rect centre
   in half-sizes (-1/0/+1); the first build multiplied by the full size and landed elements half
   a bar off. Kept as a note because it is the obvious mistake to make when reviewing that
   arithmetic.
7. **Every writer of the offsets must know the points.** The review pass found three side-only
   writers a pointed record could reach: the point pick itself (raw rects, no per-tier extra
   offset), the screen-anchor clear recapture, and the pre-load reapply stub. All three now go
   through the one point-aware capture / placement; `_CaptureAnchorOffsets` is the single place
   that turns live rects into an offset.
8. **Every field-by-field copy of an anchor record must know the new fields.** There are two
   in the file (the discard snapshot and its restore); both carry `point`/`relPoint` now.
   `CopyTable`-style copies elsewhere (profiles, `SetAnchorInfo`'s fresh record on re-anchor,
   which drops the pair on purpose) need nothing.
9. **Mover overlay vs. bar.** On small bars the hover-expanded mover is much wider than the
   element, so "the box overlaps its neighbour" is not "the anchor is wrong". The probe
   (`/nui anchor <key>`, kept outside this PR) was the way to see the actual rects; a future
   improvement could print them in the cog menu itself.
10. **Boxes in pixels, records in units - and every writer of the box must convert.** The
    Offset boxes display `PP.ToPixels(offset)` and store `PP.FromPixels(typed)`, like the X/Y
    Position boxes. The first pixel build converted only in the box's own read/commit; the
    nudge-sync that keeps the boxes in step with arrow keys still wrote units, and because
    Enter also drops focus (two commits per keystroke) the second commit re-read those units
    as pixels: 10 -> 5.7 units -> shown "6" -> 3.4 units -> "3". One helper for both paths now.


EllesmereGaming:
thank you! excellent work

thoriphes:
Heads-up: v9.2.2 looks like it dropped part of this one.

The merge (`8f24c384`) brought in `ai.point` / `ai.relPoint`, `SetAnchorParams`, `GetAnchorPoints` and the point list. The `Release v9.2.2` commit (`e1d7dd82`) rewrote `EUI_UnlockMode.lua` (292+/413-) and none of those are in the tree any more - `ai.point` and `ai.relPoint` are at 0 occurrences. The cog menu keeps the "Anchored to:" label and the Offset X/Y rows, but the own-point / target-point dropdowns are gone and the X/Y Position boxes are editable again on anchored elements, so a record is back to `{ target, side }` and a corner pair cannot be expressed.

Deliberate? If not, I have the point rows re-applied on top of v9.2.2 and can open it as a PR - the offset boxes keep their v9.2.2 behaviour (Enter applies, Escape reverts, focus loss does neither), so the diff is only the point picker.

EllesmereGaming:
this was deliberate reworked it to use existing infrastructure so when you select anchor it will give you corner options as well, and offsets can be entered manually still based on original anchor point. the point picker part of it was essentially the same system that was already in place with grow direction + anchor, so rather than adding 2 systems that did the same thing i adjusted it to just auto set grow direction and align to corner if user picks a corner entry when anchoring

PR #2157, opened 2026-09-21
fix(unitframes): keep the Forever combo point displays in step

dfrisone:
## What does this PR do?

Follow-up to #2151. That PR decided what to do with Blizzard's classic `ComboFrame` once, inside `InitializeFrames`, while the class resource style is live-togglable through `_toggleClassPower`. Review after the merge found four consequences on Forever, all from that one placement:

- **Modern to None left no combo points at all** until a reload. The pips were destroyed but `ComboFrame` stayed force-hidden, and Blizzard only re-shows it on a target change or a power event. None to Modern drew both at once.
- **Player frame on Blizzard's lost combo points.** The pips need the EllesmereUI player frame, but `ComboFrame` was hidden on the style alone. The migration in #2151 writes `"modern"` into every Forever profile, so this was the upgrade path for anyone not on our player frame.
- **Target frame hidden kept the orphan #2151 exists to remove.** Hiding the target frame suppresses Blizzard's frame too, so "no EllesmereUI target frame" is not "Blizzard's frame is alive"; that case had nowhere to re-anchor and kept drawing at the invisible frame's coordinates.
- **A switched or imported profile could still build `"blizzard"` mid-session**, since the coercion was login-only.

The decision now lives in one function, `RealiseForeverComboFrame`, run after the pip build at login and from `_toggleClassPower`, and keyed on what actually replaces the frame (`frames._customClassPower`) rather than on the style. Specifically:

- Released `ComboFrame` is re-shown through `ComboFrame_Update`, the same call Blizzard makes after a target change, so it appears for the current count and an empty row stays hidden.
- It is parented to the EllesmereUI target frame, not just anchored, so it fades and hides with the frame under the visibility modes instead of drawing at full alpha over a faded one.
- A hidden target frame with EllesmereUI as the source now hides `ComboFrame` rather than stranding it.
- Arg-less `_toggleClassPower()` calls from the colour, position and size setters return early when nothing about the decision changed, instead of re-running the anchor each time.
- One `ForeverClassPowerStyle` helper carries the `"blizzard"` coercion for both the login migration and the live toggle.

## How was it tested?

- `luac -p` clean, `git diff --check` clean, EllesmereUI style check clean against `upstream/main`.
- The changed file is byte-identical to the one deployed on the Forever test client (1.60.1), where the #2151 behaviour (rogue pips, instanced pull, paladin and warlock with no row) was confirmed.

**Not yet run in game, stated plainly:** the cases this PR changes. Cycling Modern to None and back, the player frame set to Blizzard's, the target frame hidden, and `ComboFrame` fading with the target frame have not been observed on the client. Each is reasoned from the code paths above and from Blizzard's own `ComboFrame.lua`, not from a capture.

## Screenshots

Not captured. The changes are to which of two combo point displays is shown and how it follows the target frame; there is no new art.

## Checklist

- [x] New settings default **OFF** (N/A: no setting was added)
- [x] Zero cost while disabled: the whole path is behind `EUI_CLIENT_FOREVER`, one comparison on any other client
- [x] Cheap while enabled: one decision per style change or frame build, with an early return when nothing moved; no new `OnUpdate`, timers, events or per-event allocations
- [x] No writes onto Blizzard-owned frames: `HookScript` rather than `SetScript`, and the decision state lives in addon state. `ComboFrame` is re-parented and re-anchored onto the EllesmereUI target frame when it stays the display, the same handling the Blizzard class power style already applies to the per-class bars it adopts
- [ ] Tested in-game on live; no version gates or pre-Midnight APIs added (Forever-only path; retail behaviour is unchanged and was not re-tested)

Sneaking off


dfrisone:
Closing as superseded by the 9.2.2 release. 9.2.2 stands ComboFrame down through UF_HideBlizzardFrame into the hidden parent, on the finding that ComboFrame_Update re-anchors it to TargetFrame on every change, so re-homing it onto the EllesmereUI target frame (this PR's approach) would be undone by Blizzard each update. The live-toggle concerns here are moot under that design.

One case 9.2.2 still leaves without a combo point display, for the record: a Forever rogue with the player frame on Blizzard's and the target frame on EllesmereUI. The pips need our player frame, and ComboFrame is hidden whenever we own the target frame. Narrow, and flagged rather than fixed here.

PR #2163, opened 2026-09-21
feat(partymode): add a Level Up celebration trigger

Thunderz96:
## What does this PR do?

Adds one more Auto Celebration trigger to Party Mode: **Level Up**. Tick it and Party Mode starts when your character gains a level, then stops after the existing "Auto Celebration Duration", the same as the boss kill and keystone triggers.

I know features are on hold until after 12.1, so no hard feelings if this waits or gets closed. I am sending it because it is small and it suits WoW Forever, where none of the current triggers except Randomly and Bloodlust can happen while levelling. It works the same on retail.

- New setting `partyModeTriggerLevelUp`, **default off**.
- The checkbox goes in the free middle slot of the Bloodlust row, so no row is added and nothing moves.
- `PLAYER_LEVEL_UP` is registered only while the box is ticked, through a small `EllesmereUI_UpdatePartyModeLevelUpListener()` that mirrors the Bloodlust listener. Unticked, the event is not registered at all.
- It counts towards enabling the Auto Celebration Duration slider, and "Reset to defaults" clears it with the other triggers.

## How was it tested?

- WoW Forever beta 1.60.1.69913 on v9.2.2: ticked Level Up, gained a level on a hunter, Party Mode started by itself and stopped after the set duration.
- `luac -p` passes on both files.
- Not tested on live retail. Nothing here is Forever specific; `PLAYER_LEVEL_UP` is current API.

## Checklist

- [x] Zero cost while disabled: no events registered, no polling, no hooks doing work, no frames built - the event is registered only while the trigger is on
- [x] Cheap while enabled: event-driven (no polling, no timer-based logic, no per-frame allocations) - one event per level gained; the timer is the same celebration timer the other triggers use
- [x] No writes onto Blizzard-owned frames (weak-table pattern used); `HookScript`/`hooksecurefunc` only, never `SetScript` on Blizzard frames - no frame access
- [ ] Tested in-game on live; no version gates or pre-Midnight APIs added - no gates or old APIs, but tested on the Forever beta only


EllesmereGaming:
Love this, thank you!

